Issue #3, opened 2026-02-25
pg_ash and cron on azure


CheloXL:
ph_ash tries to add cron entries directly on the cron table. On Azure, that's not possible. If you try to do so, you get the following error:

```
SQL Error [42501]: ERROR: permission denied for table job
  Where: SQL statement "update cron.job set nodename = '' where jobid = v_sampler_job"
PL/pgSQL function ash.start(interval) line 100 at SQL statement
```

Even with admin permissions, trying to manually change that table leads you to the above error. But adding/removing cron entries using the cron functions works just fine. [This page ](https://sqlkitty.com/cron-postgres-flex/)has more information about that. Also this [stack overflow](https://stackoverflow.com/questions/73475250/azure-postgresql-flexible-server-cron-job-permission-denied) post says something similar.

Issue #2, opened 2026-02-19
Can't start ash, sql erroe


CheloXL:
I have installed ash in my azure psql instance. The table ash is created and everything seems in place (no errors after running the script). Now, I'm trying to start it and I'm getting the following error:

```
SQL Error [22P02]: ERROR: invalid input syntax for type integer: "4-1"
  Where: PL/pgSQL function start(interval) line 22 at IF
  ERROR: invalid input syntax for type integer: "4-1"
  Where: PL/pgSQL function start(interval) line 22 at IF
```
Server version reports 14.20.

NikolayS:
Fixed in commit 03f6b5e.

**Root cause:** Azure Database for PostgreSQL reports pg_cron version as `'4-1'` (dash separator) rather than `'4.1'` (dot separator). The version check used `string_to_array(v_cron_version, '.')` — on `'4-1'` this yields a single element `'4-1'` which fails when cast to `int[]`.

**Fix:** Strip non-numeric, non-dot characters before splitting:
```sql
string_to_array(regexp_replace(v_cron_version, '[^0-9.]', '', 'g'), '.')::int[]
```

This handles `'1.5'`, `'4-1'`, `'1.6.5'` etc. Applied to ash-install.sql, ash-1.1.sql, and ash-1.1-to-1.2.sql.

Please re-run `\i ash-install.sql` (or the latest version file) and try `select * from ash.start('1 second')` again. Let us know if it works!

CheloXL:
Mmm.. I'm still getting the same error message (now on another line). Now that I'm reading the error, it says _1 minutes_ but I ran the following command: `select * from ash.start('1 minute');`, so something in the code is pluralizing _minute_...

```
SQL Error [22023]: ERROR: invalid schedule: 1 minutes
  Hint: Use cron format (e.g. 5 4 * * *), or interval format '[1-59] seconds'
  Where: SQL statement "select cron.schedule(
      'ash_sampler',
      v_schedule,
      'set statement_timeout = ''500ms''; select ash.take_sample()'
    )"
PL/pgSQL function ash.start(interval) line 79 at SQL statement

Error position:
```

CheloXL:
Closed in favor of #3

PR #5, opened 2026-02-26
feat: ash.logging_level() — configurable session logging in take_sample() (refs #4)

### LINKED ISSUE #4: Idle State included causing false info when using connection pooler
Hallo

We've been struggling with write performance issues, which looks like latency during fsync. I've been playing around with pg_ash to try and get info to show this, but I think the idle connections are diluting the information.

Below an example of the info when I run top_waits, top_queries, top_by_type. This made me wonder of there is a client issue due to the ClientRead.
Image

If I then go have a look at the specific query id I get the below. This showed the same ClientRead.
Image

But then if I go have a look at current pg_stat_activity I see a lot of idle connections that is in the connection pools and their info stays at ClientRead and the query COMMIT.
Image

Does it make sense to exclude state=idle from the collecting info?

NikolayS:
## What

Adds `ash.logging_level(text)` — set the verbosity of per-session messages emitted by `take_sample()`. Supports all standard Postgres message levels.

## Why

Issue #4 raised a question about whether idle (pooler) connections leak into pg_ash samples. The code correctly excludes `state = 'idle'` at all reads, but this is hard to verify without instrumenting the sampler. `logging_level` gives operators a way to see exactly what gets captured — and at what severity they want to see it.

## Usage

```sql
select ash.logging_level('debug');    -- one line per sampled session at DEBUG
select ash.logging_level('notice');   -- visible without changing any GUC
select ash.logging_level('off');      -- disable (default)
select ash.logging_level();           -- show current level
```

Each message looks like:
```
ash.take_sample: pid=1234 state=active wait_type=Client wait_event=ClientRead backend_type=client backend query_id=...
```

To see `debug` output in psql: `set client_min_messages = debug;`
To see `debug` output in logs: `set log_min_messages = debug;` (superuser, no restart)
`notice` / `warning` / `log` appear in logs by default — no GUC changes needed.

## Supported levels

`off`, `debug5`, `debug4`, `debug3`, `debug2`, `debug1`, `debug`, `info`, `notice`, `warning`, `log`

## Changes

- `ash.config`: `logging_level text` column with check constraint (default `'off'`)
- `ash.take_sample()`: `case` branch on level, emits `raise` per captured session
- `ash.logging_level(text)`: setter/getter with input validation + contextual hints
- `ash.status()`: reports `logging_level`
- Security: `logging_level()` revoked from PUBLIC, granted to schema owner only
- `sql/ash-1.2-to-1.3.sql`: idempotent upgrade script

Closes #4

### DIFF
--- a/sql/ash-install.sql
+++ b/sql/ash-install.sql
@@ -270,9 +270,6 @@ declare
   v_current_slot smallint;
   v_rows_inserted int := 0;
   v_seen_waits text[] := '{}';
-  v_state text;
-  v_type text;
-  v_event text;
 begin
   -- Get sample timestamp (seconds since epoch, from now())
   v_sample_ts := extract(epoch from now() - ash.epoch())::int4;
@@ -331,7 +328,6 @@ begin
       and (sa.backend_type = 'client backend'
        or (v_include_bg and sa.backend_type in ('autovacuum worker', 'logical replication worker', 'parallel worker', 'background worker')))
       and sa.pid <> pg_backend_pid()
-    order by sa.pid
   loop
     -- Register wait event if not yet seen this tick (dedup in memory, not per row lookup).
     if not (v_rec.state || '|' || v_rec.wait_type || '|' || v_rec.wait_event = any(v_seen_waits)) then


NikolayS:
## REV — Code Review

### BLOCKING

**None.**

---

### NON-BLOCKING

**1. `v_seen_waits` array grows unbounded and uses linear scan** `sql/ash-install.sql` — Read 1 loop

`= any(v_seen_waits)` is O(n) per row. In practice the distinct set of `(state, wait_type, wait_event)` tuples is tiny (tens, not thousands), so this is harmless at any realistic concurrency. But the code comment says "dedup in memory, not per row lookup" without acknowledging the scan cost. Minor: either add a note that the set is bounded by the number of distinct wait event types (~dozens max), or use a different structure. Not worth changing, just worth documenting.

**2. `order by sa.pid` in Read 1 is new overhead with no benefit** `sql/ash-install.sql` — Read 1 loop

The old DISTINCT loop had no ORDER BY. The new loop added `order by sa.pid` — presumably for deterministic log output. That's reasonable when debug_logging is on, but it adds a sort even when `v_debug_logging = false`. Consider moving the sort inside the condition, or accept the overhead as negligible (pid sort on a small rowset is cheap).

**3. `v_state`, `v_type`, `v_event` declared but now unused** `sql/ash-install.sql` — DECLARE block

Three variables remain from an earlier version of the sampler. Dead code. Safe to drop.

**4. Schema parity test simulates 1.2 state imprecisely** `.github/workflows/test.yml` — "Schema parity" step

Path B does `\i sql/ash-install.sql` (which gives 1.3), then manually drops `debug_logging` and adds `logging_level` to fake 1.2. This tests the migration script in isolation but the function bodies installed are already 1.3 (since `ash-install.sql` ran). A real 1.2 → 1.3 upgrade would run the migration on top of a true 1.2 install. The test is still useful and catches schema column drift; just note in a comment that function bodies are not rolled back to 1.2 — only the column is simulated.

**5. README not updated** `README.md`

`ash.debug_logging()` is a new user-facing function. If the README lists available functions or has a usage section, it should mention this. Minor but users looking at the README won't find it.

---

### POTENTIAL

**6. `RAISE LOG` may flood pg_log at 1s sampling under high concurrency** — design trade-off

At 200 active backends × 1 sample/sec = 200 log lines/sec. At 1KB/line that's ~17 MB/min in the server log. This is expected behavior for a diagnostic tool and the flag defaults to `false`, so it's not a bug. Worth a one-line warning in the `debug_logging(true)` return message, e.g. appending `(N active sessions × 1/s — disable promptly)`. Low priority.

---

### Summary

Solid PR. The core design (binary flag, RAISE LOG, single pg_stat_activity scan, in-memory dedup) is correct. The main actionable items are: drop the three dead variables (#3) and remove the spurious `order by sa.pid` when logging is off (#2). Everything else is cosmetic or informational.

CI: ✅ all green on PG14–18.

Issue #30, opened 2026-04-05
feat: configurable N-partitions + rollup tables for long-term storage


NikolayS:
## Summary

Two related features for pg_ash v1.5:

1. **Configurable N partitions** — replace hardcoded 3-partition ring buffer with user-configurable N (default 3, range 3–32), modeled after pg-flight-recorder's `ring_buffer_slots` approach.
2. **Rollup tables for long-term storage** — minute (`rollup_1m`, 30-day retention) and hourly (`rollup_1h`, 5-year retention) aggregates that survive rotation, per the existing `ROLLUP_DESIGN.md`.

## Problem

- pg_ash hardcodes 3 partitions (`sample_0/1/2`, `query_map_0/1/2`) with `% 3` modular arithmetic and a `CASE` statement in `rotate()`. Users who want longer raw retention (e.g., 7 days) must edit SQL source.
- Raw samples rotate away after ~1 rotation period. No mechanism for long-term trend analysis ("is the system getting slower this month?").

## Proposed Design

Full spec: [`blueprints/CONFIGURABLE_PARTITIONS_AND_ROLLUP.md`](https://github.com/NikolayS/pg_ash/blob/claude/review-partition-spec-mmJDA/blueprints/CONFIGURABLE_PARTITIONS_AND_ROLLUP.md) (on branch `feat/configurable-partitions-and-rollup`)

### Part 1: Configurable partitions
- `num_partitions` column in `ash.config` (default 3, range 3–32)
- `rotate()` and `take_sample()` switch from hardcoded `CASE`/`IF` to `execute format(...)` with `% N`
- `rebuild_partitions(N)` — destructive admin function (modeled after pg-flight-recorder's `rebuild_ring_buffers`)
- `_rebuild_query_map_view()` — dynamically rebuilds the `UNION ALL` view
- Retention math: `(N-2) × rotation_period` readable data

### Part 2: Rollup tables
- `rollup_1m` (30-day retention, ~43 MiB/db) + `rollup_1h` (5-year, ~77 MiB/db)
- `rollup_minute()`, `rollup_hour()`, `rollup_cleanup()` — all idempotent via upsert
- Denormalized `query_id` (int8) in rollups — no dependency on `query_map` surviving
- Backend-seconds count unit (Oracle ASH-compatible)
- Reader functions: `minute_waits`, `hourly_queries`, `daily_peak_backends` + `_at` variants

### Implementation order
- **Phase A** (partitions) can ship independently
[...]

PR #31, opened 2026-04-13
Implement configurable N-partitions and rollup tables for pg_ash v1.4

NikolayS:
## Summary

This PR implements two major features for pg_ash v1.4:

1. **Configurable N-partitions** — Replace hardcoded 3-partition ring buffer with user-configurable partition count (default 3, range 3–32)
2. **Rollup tables for long-term storage** — Minute and hourly aggregates that survive rotation, enabling historical analysis beyond raw sample retention

Both features share a dependency: rotation logic must become dynamic to support variable partition counts and correctly reference partition boundaries for rollup operations.

### Configuration & Schema
- Added `num_partitions` (smallint, default 3, range 3–32) to `ash.config`
- Added `sampling_enabled` (bool, default true) flag to control sampling during maintenance operations
- Added `skipped_samples` (int4) counter to track samples skipped during rebuild/maintenance
- Added rollup retention and configuration columns: `rollup_1m_retention_days`, `rollup_1h_retention_days`, `rollup_min_backend_seconds`, `last_rollup_1m_ts`, `last_rollup_1h_ts`
- Updated version from 1.3 to 1.4

### Dynamic Partition Management
- Replaced hardcoded 3-partition creation with dynamic `DO` block using `generate_series` based on `num_partitions` config
- Implemented `ash._rebuild_query_map_view()` function to dynamically generate `query_map_all` UNION ALL view for N partitions
- Implemented `ash._drop_all_partitions()` function using catalog enumeration to safely drop all sample partitions and query_map tables

### Core Function Updates

**`take_sample()`**
- Added advisory lock participation: `pg_try_advisory_xact_lock(0, hashtext('ash_operation')::int4)` to coordinate with rebuild operations
- Added `sampling_enabled` check — returns early if disabled, increments `skipped_samples` counter
- Replaced 3-way hardcoded `IF` statement for query_map inserts with single dynamic SQL template using `format()` and `execute`
- Bug fixes now apply once instead of three times

**`rotate()`**
- Replaced hardcoded `% 3` modulo arithmetic with dynamic `% v_num_partitions`
- Replaced `CASE` statement for partition truncation with dynamic SQL using `format()` and `execute`
- Added pre-truncation rollup call: `ash.rollup_minute()` executes before config update to process endangered minutes before partition truncation
- Integrated rollup batch limit based on rotation period

**New `rebuild_partitions(p_num int default null)` function**
- Hardened admin function for changing partition count (destructive — clears all raw samples, preserves rollup tables)
- Implements robust protocol: disable sampling → stop jobs → acquire exclusive lock → drain in-flight operations → drop/recreate partitions → rebuild view
- Uses two-key advisory lock `(1, hashtext('ash_rebuild')::int4)` for exclusive coordination
- Polls `pg_locks` for up to 5 seconds to drain in-flight `take_sample()` operations
- Leaves `sampling_enabled = false` on completion — user must explicitly call `ash.start()` to resume
- Includes comprehensive error handling and recovery documentation

### Helper Functions
- Added `ash.ts_from_timestamptz(p_ts timestamptz)` — converts timestamptz to int4 epoch offset (PUBLIC, immutable, parallel safe)
- Added `ash.ts_to_timestamptz(p_ts int4)` — converts int4 epoch offset back to timestamptz (PUBLIC, immutable, parallel safe)

### Advisory Locking Contract
- Implemented two-key advisory lock protocol for safe concurrent operation:
  - `(0, hashtext('ash_operation')::int4)` — participation lock held by `take_sample()` and rollup operations (xact-level, auto-releases)
  - `(1, hashtext('ash_rebuild')::int4)` — exclusive lock held by `rebuild_partitions()` (xact-level,

https://claude.ai/code/session_011RFMSQLEj8Eo5t3NBExyVT


damuanbaez-ui (approved):

NikolayS:
## REV Code Review Report

- **MR:** NikolayS/pg_ash#31 - Implement configurable N-partitions and rollup tables for pg_ash v1.4
- **Author:** NikolayS
- **AI-Assisted:** Yes (all commits by Claude)

**CI Status:** ✅ pass (Analyze + CodeQL)

---

### BLOCKING ISSUES (8)

Issues that must be addressed before merge (high-confidence CRITICAL, HIGH, MEDIUM severity).

**MEDIUM** `sql/ash-install.sql` - No upgrade script for existing 1.3 installations
> No `sql/ash-1.3-to-1.4.sql` file exists. The blueprint specifies a detailed 15-step upgrade ordering, but no script implements it. Existing 1.3 users have no in-place upgrade path and would lose all data.
> **Fix:** Create `sql/ash-1.3-to-1.4.sql` following the blueprint's upgrade ordering, or document that fresh install is required for v1.4.

**MEDIUM** `sql/ash-install.sql:4` - Install script header missing 1.3→1.4 upgrade path
> The header lists upgrade paths 1.0→1.1→1.2→1.3 but not 1.3→1.4.
> **Fix:** Add `-- Upgrade from 1.3: \i sql/ash-1.3-to-1.4.sql`

**MEDIUM** `.github/workflows/test.yml` - rollup_minute() data correctness not verified
> Test seeds 3 samples (active_count=2, known wait_id) but only asserts row existence/count. Never checks that `samples=3`, `peak_backends=2`, or that `wait_counts` contains the correct `[wait_id, 3]` pair. Cannot distinguish correct aggregation from garbage.
> **Fix:** After `rollup_minute()`, assert exact values: `samples = 3`, `peak_backends = 2`, `wait_counts[1] = v_wait_id`, `wait_counts[2] = 3`.

**MEDIUM** `.github/workflows/test.yml` - rollup_hour() aggregation values not verified
> Seeds two rollup_1m rows (samples=60+60, peak_backends=5/3, wait_counts 120+100) but only checks row existence. Never verifies `samples=120`, `peak_backends=5`, or merged `wait_counts=[wait_id, 220]`.
> **Fix:** Assert exact aggregated values for the hourly rollup row.

**MEDIUM** `.github/workflows/test.yml` - rollup_cleanup() retention behavior untested
> Only verifies function runs and returns a string with no data. No test seeds old data, runs cleanup, and verifies old rows are deleted while recent rows survive.
> **Fix:** Insert a rollup_1m row older than retention (e.g., 40 days), run cleanup with 30-day retention, assert old row deleted and recent row preserved.

**MEDIUM** `.github/workflows/test.yml` - No advisory lock concurrency tests
> The blueprint explicitly requires proof that `take_sample()` holding `(0, hashtext('ash_operation'))` appears correctly in `pg_locks`. No such test exists. The locking protocol is the primary correctness mechanism for rebuild safety.
> **Fix:** In a DO block, manually acquire the advisory lock, assert it appears in `pg_locks` with `classid=0` and correct `objid`.

**MEDIUM** `.github/workflows/test.yml` - Uninstall test not updated for rollup tables
> The existing uninstall test was not modified to verify that `rollup_1m`, `rollup_1h`, and rollup functions are dropped after `uninstall()`.
> **Fix:** Add assertions for absence of rollup tables and functions after uninstall.

**MEDIUM** `sql/ash-install.sql` - Removed `_to_sample_ts()` without deprecation note
> The public function `ash._to_sample_ts()` was renamed to `ash.ts_from_timestamptz()`. No changelog or compatibility note exists. External tools/scripts calling the old name will silently break.
> **Fix:** Document the rename in a CHANGES file. The upgrade script should include a compatibility wrapper or clear warning.

---

### NON-BLOCKING (4)

Minor issues and suggestions. Can be addressed later.

**LOW** `blueprints/CONFIGURABLE_PARTITIONS_AND_ROLLUP.md:681` - Blueprint says "v1.5" but implementation is v1.4
> **Suggestion:** Update blueprint header/Overview to say v1.4, or add a resolved note.

**LOW** `blueprints/CONFIGURABLE_PARTITIONS_AND_ROLLUP.md:1882` - Upgrade path uses `{prev}/{next}` placeholders
> **Suggestion:** Replace with `1.3`/`1.4` now that the version is confirmed.

**INFO** `sql/ash-install.sql:~2864` - Missing `comment on table/column` for rollup tables
> **Suggestion:** Add `comment on table ash.rollup_1m is '...'` and column comments for `ts`, `wait_counts`, `query_counts` etc. (queryable via `\d+`).

**INFO** `sql/ash-install.sql:~3491` - `GROUP BY d` uses SELECT alias instead of explicit column expression
> **Suggestion:** Change `group by d` to `group by (ash.ts_to_timestamptz(r.ts))::date` for clarity.

---

### POTENTIAL ISSUES (9)

Issues with moderate confidence (4-7/10). Review manually — may be false positives.

**MEDIUM** `sql/ash-install.sql:~2505` - Drain loop in rebuild_partitions() proceeds unconditionally after timeout *(confidence: 7/10)*
> After 5s timeout (10 × 0.5s), proceeds to DROP tables even if operations are still in-flight.
> **Suggestion:** Raise exception on drain timeout instead of proceeding silently.

**MEDIUM** `sql/ash-install.sql:~683` - rotate() uses session-level lock while other functions use xact-level *(confidence: 7/10)*
> `rotate()` uses `pg_try_advisory_lock` (session-level) while `take_sample()`, `rollup_minute()`, and `rebuild_partitions()` use `pg_try_advisory_xact_lock`. If `rotate()` fails between lock and unlock, the lock leaks for the session lifetime. pg_cron reuses connections.
> **Suggestion:** Switch to `pg_try_advisory_xact_lock` for consistency, or document the intentional choice.

**MEDIUM** `.github/workflows/test.yml` - rebuild_partitions() failure recovery not tested *(confidence: 7/10)*
> **Suggestion:** Simulate failed rebuild state, verify `ash.start()` recovers correctly.

**MEDIUM** `.github/workflows/test.yml` - Gap detection WARNING path not tested *(confidence: 7/10)*
> **Suggestion:** Set watermark past a gap with later data, run rollup_minute(), verify no error.

**MEDIUM** `sql/ash-install.sql:~3077` - rollup_minute wait_event join relies on undocumented format contract *(confidence: 5/10)*
> Join `wm.state || '|' || wm.type || '|' || wm.event = d.wait_event` assumes decode_sample() returns this exact format.
> **Suggestion:** Document the expected format or join on `wm.id` directly.

**MEDIUM** `sql/ash-install.sql:~710` - rotate() rollup exception handler is undocumented upgrade safety guard *(confidence: 7/10)*
> **Suggestion:** Add comment: `-- Upgrade safety: can be removed once all environments are on v1.4`

**LOW** `sql/ash-install.sql:~2349` - No CHECK constraint on current_slot *(confidence: 6/10)*
> **Suggestion:** Add `check (current_slot >= 0 and current_slot < num_partitions)` for defense-in-depth.

**INFO** `sql/ash-install.sql` - _merge_wait_counts and _merge_query_counts are near-duplicates *(confidence: 4/10)*
> **Suggestion:** Add cross-reference comments between the two functions.

**MEDIUM** `sql/ash-install.sql:~2997` - rollup_minute() watermark init performance not documented *(confidence: 7/10)*
> First-run scans all sample partitions. May cause unexpectedly long queries on large databases.
> **Suggestion:** Add comment documenting catch-up rate (~60 min/call).

---

### Summary

| CI/Pipeline | 0 | 0 | 0 |
| Security | 0 | 3 | 0 |
| Bugs | 0 | 1 | 0 |
| Tests | 5 | 2 | 0 |
| Guidelines | 0 | 0 | 5 |
| Docs | 3 | 3 | 0 |
| Metadata | 0 | 0 | 0 |

Note:
- **Findings**: High-confidence issues (8-10/10) — blocking or non-blocking per severity
- **Potential**: Medium-confidence issues (4-7/10) — review manually, may be false positives
- **Filtered**: Low-confidence issues (0-3/10) — excluded as likely false positives

---

NikolayS:
### Addendum: Bug Hunter Findings

The bug analysis agent completed after the initial review and found additional issues, including one **HIGH severity bug**.

---

**HIGH** `sql/ash-install.sql` - `_truncate_pairs()` has `LIMIT` without `ORDER BY` — returns arbitrary rows instead of top-N

This is a **real correctness bug**. The `top_n` CTE uses `LIMIT p_top` but has no `ORDER BY` clause:

```sql
top_n as (
    select id, cnt,
           row_number() over (order by cnt desc, id asc) as rn
    from pairs
    limit p_top  -- ← No ORDER BY! Returns arbitrary p_top rows
)
```

The `row_number()` window function computes a value but does NOT control which rows survive the `LIMIT`. PostgreSQL may return any arbitrary `p_top` rows, not the top ones by count. The blueprint spec correctly has `order by cnt desc, id asc` before `limit p_top`.

> **Fix:** Add `order by cnt desc, id asc` before `limit p_top`.

---

**HIGH** `sql/ash-install.sql` - `decoded` CTE in `rollup_minute()` may have duplicate `datid` columns *(confidence: 7/10)*

The CTE selects both `s.datid` from `ash.sample` AND `(ash.decode_sample(s.data, s.slot)).*` which likely also returns a `datid` column. If `decode_sample` expands to per-backend entries with their own `datid`, the subsequent `GROUP BY datid` may use the sample-level datid (first occurrence) instead of the per-backend one.

> **Fix:** Remove `s.datid` from the decoded CTE and rely on decode_sample's datid.

---

**MEDIUM** `sql/ash-install.sql` - `peak_backends` uses `max(active_count)` instead of counting decoded rows per spec *(confidence: 8/10)*

The blueprint spec says peak_backends should be "computed by counting decoded rows per (sample_ts, datid), then taking max of those counts." The implementation uses `max(active_count)` which is a pre-stored value. These could diverge if encoding has edge cases.

> **Fix:** Compute as `max(backend_count)` from a subquery: `select datid, sample_ts, count(*) as backend_count from decoded group by datid, sample_ts`.

---

**MEDIUM** `sql/ash-install.sql` - `query_agg` uses `d.query_id` directly but spec says decode_sample returns `query_map_id` *(confidence: 7/10)*

The blueprint says to "resolve query_map_id → query_id via query_map_all." If decode_sample returns `query_map_id` (not `query_id`), the column reference is wrong.

> **Fix:** Verify decode_sample's return type. If it returns `query_map_id`, join with `query_map_all` to resolve to `query_id`.

---

**MEDIUM** `sql/ash-install.sql` - Spec WARNING about `sampling_enabled=false` surviving rollback is incorrect *(confidence: 9/10)*

The spec warns that `sampling_enabled = false` "was committed in a prior statement and is not rolled back." This is wrong — `rebuild_partitions()` is a single PL/pgSQL function; all statements are in one transaction. On failure, everything rolls back including the flag change. The WARNING may cause operators to perform unnecessary manual recovery.

> **Fix:** Update the spec/comments to reflect actual transactional behavior.

---

**LOW** `sql/ash-install.sql` - `having count(*) >= v_min_backend_seconds` is semantically wrong at non-1s sampling intervals *(confidence: 8/10)*

At 1s sampling, count = backend-seconds. At 2s sampling, count is half. The variable name implies time-based threshold but comparison is sample-count-based.

> **Fix:** Rename to `rollup_min_sample_count` or multiply by sample interval.

---
*REV-assisted review addendum — bug analysis*

NikolayS:
## REV Review Response

Addressed in commit a605d4f. Summary:

### Bugs fixed (3)

| Severity | Issue | Fix |
|----------|-------|-----|
| **HIGH** | `_truncate_pairs()` `LIMIT` without `ORDER BY` — returned arbitrary rows | Added `ORDER BY cnt DESC, id ASC` before `LIMIT` |
| **HIGH** | `rollup_minute()` wait_event join format mismatch — joined on `state\|type\|event` but `decode_sample()` returns `type:event` | Fixed join to match `decode_sample()`'s actual output format |
| **MEDIUM** | `rotate()` session-level advisory lock leaks on failure (pg_cron reuses connections) | Switched to `pg_try_advisory_xact_lock`, removed manual unlock calls |

### Tests added (4)

- **rollup_minute() exact values**: asserts `samples=3`, `peak_backends=2`, `wait_counts` contains correct wait_id and count=6
- **rollup_hour() exact values**: asserts `samples=120`, `peak_backends=5`, merged `wait_counts[2]=220`
- **rollup_cleanup() retention**: seeds old (40d) + recent (1d) rows, verifies old deleted and recent preserved
- **Advisory lock visibility**: acquires `ash_operation` lock, verifies it appears in `pg_locks` with correct `classid=0` / `objid`

### False positives investigated (3)

- **decoded CTE duplicate datid**: `decode_sample()` returns `(wait_event, query_id, count)` — no `datid` column, no conflict
- **query_agg uses d.query_id directly**: `decode_sample()` resolves `query_map_id` → `query_id` internally
- **peak_backends via max(active_count)**: `active_count` is already per-database per-sample; `max()` is correct

### Not addressed (by design)

- **No upgrade script**: per CLAUDE.md — "generated at release time, not in feature PRs"
- **`_to_sample_ts` deprecation note**: internal function (`_` prefix), will be handled in upgrade script
- **Blueprint version placeholders**: cosmetic, deferred

### Non-blocking items addressed in earlier commits

- Cross-reference comments between `_merge_wait_counts` / `_merge_query_counts` (commit 86d7f52)
- README updated with full v1.4 documentation, defaults table, usage examples (commit 704f57b)

---
_Generated by [Claude Code](https://claude.ai/code)_

NikolayS:
## REV Code Review Report (Re-review)

- **MR:** NikolayS/pg_ash#31 - Implement configurable N-partitions and rollup tables for pg_ash v1.4
- **Author:** NikolayS
- **AI-Assisted:** Yes (all commits by Claude)
- **Commits since last review:** 3 (fixes + docs)

**CI Status:** ✅ pass (Analyze + CodeQL)

---

### Previous Review Status

3 commits pushed addressing the [initial review](https://github.com/NikolayS/pg_ash/pull/31#issuecomment-4233271265) and [bug addendum](https://github.com/NikolayS/pg_ash/pull/31#issuecomment-4233272718):

| Previous Finding | Status | Notes |
|---|---|---|
| **HIGH** `_truncate_pairs()` LIMIT without ORDER BY | ✅ Fixed | `order by cnt desc, id asc` added before `limit` |
| **HIGH** `rollup_minute()` wait_event join format | ✅ Fixed | Join now uses `case when wm.event = wm.type then wm.event else wm.type \|\| ':' \|\| wm.event end` |
| **MEDIUM** `rotate()` session-level lock leak risk | ✅ Fixed | Switched to `pg_try_advisory_xact_lock`, removed manual unlocks |
| **MEDIUM** rollup_minute data correctness test | ✅ Fixed | Asserts exact samples=3, peak_backends=2, wait_counts values |
| **MEDIUM** rollup_hour aggregation test | ✅ Fixed | Asserts samples=120, peak_backends=5, merged wait_counts=220 |
| **MEDIUM** rollup_cleanup retention test | ✅ Added | Seeds 40d old + 1d recent, verifies old deleted |
| **MEDIUM** Advisory lock concurrency test | ✅ Added | Verifies pg_locks classid/objid columns |
| **MEDIUM** README/docs for new features | ✅ Added | Comprehensive README with defaults, rollup docs, partition config |
| **HIGH** decoded CTE duplicate datid | ❌ False positive | decode_sample doesn't return datid |
| **MEDIUM** query_agg uses d.query_id | ❌ False positive | decode_sample resolves query_map_id internally |
| **MEDIUM** peak_backends max(active_count) | ❌ False positive | active_count is already per-database per-sample |
| **MEDIUM** No upgrade script | ⏭️ By design | Per CLAUDE.md: "generated at release time, not in feature PRs" |
| **MEDIUM** _to_sample_ts deprecation | ⏭️ By design | Will be handled in upgrade script |

---

### BLOCKING ISSUES (0)

All previously blocking bugs have been fixed. No new blocking issues found.

---

### NON-BLOCKING (5)

Minor issues and suggestions. Can be addressed later.

**LOW** `sql/ash-install.sql:4` - Install header still missing 1.3→1.4 upgrade path
> Header lists 1.0→1.1→1.2→1.3 but not 1.3→1.4. Even if the upgrade script ships later, the header should note it.
> **Suggestion:** Add `-- Upgrade from 1.3: \i sql/ash-1.3-to-1.4.sql` (when available)

**LOW** `blueprints/CONFIGURABLE_PARTITIONS_AND_ROLLUP.md` - Blueprint says "v1.5" but implementation is v1.4
> **Suggestion:** Update blueprint header to say v1.4 since the version is now confirmed.

**LOW** `blueprints/CONFIGURABLE_PARTITIONS_AND_ROLLUP.md:1882` - Upgrade path uses `{prev}/{next}` placeholders
> **Suggestion:** Replace with `1.3`/`1.4`.

**INFO** `sql/ash-install.sql` - Missing `comment on table/column` for `rollup_1m` and `rollup_1h`
> **Suggestion:** Add `comment on table` DDL for queryability via `\d+`.

**INFO** `sql/ash-install.sql` - `GROUP BY d` in `daily_peak_backends_at` uses SELECT alias
> **Suggestion:** Use explicit expression `group by (ash.ts_to_timestamptz(r.ts))::date`.

---

### POTENTIAL ISSUES (4)

Issues with moderate confidence (4-7/10). Review manually.

**MEDIUM** `sql/ash-install.sql` - Drain loop in `rebuild_partitions()` still proceeds unconditionally after 5s timeout *(confidence: 7/10)*
> After 10 iterations × 0.5s, proceeds to DROP tables even if locks are still held.
> **Suggestion:** Add `v_drained` flag; raise exception on timeout.

**MEDIUM** `.github/workflows/test.yml` - Uninstall test still not updated for rollup tables *(confidence: 7/10)*
> After `uninstall()`, test doesn't verify `rollup_1m`, `rollup_1h`, and rollup functions are gone.
> **Suggestion:** Add assertions for rollup table/function absence.

**LOW** `sql/ash-install.sql` - No CHECK constraint on `current_slot` *(confidence: 6/10)*
> **Suggestion:** Add `check (current_slot >= 0 and current_slot < num_partitions)`.

**LOW** `sql/ash-install.sql` - No CHECK constraints on retention days *(confidence: 5/10)*
> Negative `rollup_1m_retention_days` would cause `rollup_cleanup()` to delete all data.
> **Suggestion:** Add `check (rollup_1m_retention_days >= 0)` and same for `_1h`.

---

### Summary

| CI/Pipeline | 0 | 0 | 0 |
| Security | 0 | 1 | 0 |
| Bugs | 0 | 0 | 3 |
| Tests | 0 | 1 | 0 |
| Guidelines | 0 | 0 | 3 |
| Docs | 0 | 0 | 0 |
| Metadata | 0 | 0 | 0 |

**Previous review: 9 blocking → This review: 0 blocking** (3 bugs fixed, 4 tests added, 3 false positives confirmed)

**Result: PASSED** (with non-blocking suggestions)

---

NikolayS:
## REV Code Review Report (3rd review)

- **MR:** NikolayS/pg_ash#31 - Implement configurable N-partitions and rollup tables for pg_ash v1.4
- **Author:** NikolayS
- **AI-Assisted:** Yes (all commits by Claude)
- **Commits:** 12 (5 new since last review)

**CI Status:** ✅ pass — PG 14, 15, 16, 17, 18 + Analyze + CodeQL

---

### Changes Since Last Review

5 new commits addressing the remaining upgrade path issues:

| Commit | Description |
|--------|-------------|
| `92dfee0d` | **Upgrade script added** (`ash-1.3-to-1.4.sql`) + column migration DO block for safe 1.3→1.4 upgrades |
| `0fbc1781` | CI fix: clear sample data before rollup empty-state tests |
| `b17d39d7` | CI fix: schema equivalence — drop stale `_to_sample_ts`, sync version default |
| `99ec8215` | CI fix: robust rollup test cleanup — stop jobs, truncate parent table |
| `3f992783` | CI fix: keep `_to_sample_ts` as compat alias for old upgrade scripts |

**Key improvements:**
- ✅ `sql/ash-1.3-to-1.4.sql` now exists (uses idempotent `\ir ash-install.sql` pattern)
- ✅ Column migration DO block ensures v1.4 columns exist before any code reads them
- ✅ `_to_sample_ts()` kept as thin compat wrapper (old upgrade scripts reference it)
- ✅ Full CI matrix passing: PG 14, 15, 16, 17, 18

---

### Cumulative Review Status

All previously blocking issues from reviews 1 & 2 are resolved:

| Finding | Review | Status |
|---------|--------|--------|
| `_truncate_pairs()` LIMIT bug | R1 | ✅ Fixed |
| `rollup_minute()` wait_event join | R1 | ✅ Fixed |
| `rotate()` session-level lock leak | R1 | ✅ Fixed |
| rollup_minute/hour data correctness tests | R1 | ✅ Added |
| rollup_cleanup retention test | R1 | ✅ Added |
| Advisory lock visibility test | R1 | ✅ Added |
| No upgrade script | R2 | ✅ Added (`ash-1.3-to-1.4.sql`) |
| `_to_sample_ts` removal | R2 | ✅ Kept as compat alias |
| README/docs | R2 | ✅ Comprehensive |

---

### BLOCKING ISSUES (0)

No blocking issues remain.

---

### NON-BLOCKING (3)

**LOW** `sql/ash-install.sql:4` - Install header still missing `-- Upgrade from 1.3: \i sql/ash-1.3-to-1.4.sql`
> The upgrade script exists but isn't listed in the header comment block.

**LOW** `blueprints/CONFIGURABLE_PARTITIONS_AND_ROLLUP.md` - Blueprint still says "Target: pg_ash v1.5"
> Implementation shipped as v1.4. Minor spec hygiene.

**INFO** `sql/ash-install.sql` - Missing `comment on table` for `rollup_1m` and `rollup_1h`
> Not queryable via `\d+` without DDL comments.

---

### POTENTIAL ISSUES (3)

**MEDIUM** `sql/ash-install.sql` - Drain loop in `rebuild_partitions()` proceeds after timeout *(confidence: 7/10)*
> After 5s, drops tables regardless of in-flight operations.

**LOW** `sql/ash-install.sql` - No CHECK constraint on `current_slot` *(confidence: 6/10)*
> Relies on application logic in `rotate()` to keep in bounds.

**LOW** `sql/ash-install.sql` - No CHECK on `rollup_*_retention_days >= 0` *(confidence: 5/10)*
> Negative retention causes `rollup_cleanup()` to delete all data.

---

### Summary

| CI/Pipeline | 0 | 0 | 0 |
| Security | 0 | 1 | 0 |
| Bugs | 0 | 0 | 0 |
| Tests | 0 | 0 | 0 |
| Guidelines | 0 | 0 | 3 |
| Docs | 0 | 0 | 0 |
| Metadata | 0 | 0 | 0 |

**Review progression:** R1: 9 blocking → R2: 0 blocking → R3: 0 blocking, upgrade script added, CI green on PG 14-18

**Result: PASSED**

---

NikolayS:
## Manual Review & Testing Report

Thorough manual review of all 4021 lines of `ash-install.sql`, 2397 lines of `test.yml`, and the upgrade script. Tested locally on PostgreSQL 16.

---

### BLOCKING BUG FOUND (1)

**HIGH** `rollup_hour()` crashes with `"cannot accumulate arrays of different dimensionality"` when minute-level `wait_counts` or `query_counts` arrays have different lengths.

PostgreSQL's built-in `array_agg()` on arrays requires all sub-arrays to have equal dimensions. In production, different minutes will naturally have different numbers of wait events and queries, so this will crash on almost any real workload.

**Reproduction:**
```sql
-- Two minutes: one has 2 wait types (4-element array), other has 1 (2-element)
INSERT INTO ash.rollup_1m ... VALUES
  (hour_ts,      0, 30, 5, '{cpu_id,60,io_id,40}'::int4[], ...),
  (hour_ts + 60, 0, 30, 3, '{cpu_id,50}'::int4[], ...);
SELECT ash.rollup_hour();  -- ERROR: cannot accumulate arrays of different dimensionality
```

**Fix (pushed to `claude/test-review-pr31-ex5IU`):** Created custom `_int4_array_cat_agg` / `_int8_array_cat_agg` aggregates using `array_cat()` as sfunc, which flat-concatenates arrays regardless of length. Replaced `array_agg(wait_counts)` → `ash._int4_array_cat_agg(wait_counts)` and `array_agg(query_counts)` → `ash._int8_array_cat_agg(query_counts)` in `rollup_hour()`.

---

### DEFENSE-IN-DEPTH FIXES (2)

**MEDIUM** No CHECK constraint on `rollup_1m_retention_days` / `rollup_1h_retention_days` — negative values cause `rollup_cleanup()` to delete ALL data.

Confirmed by test: setting `rollup_1m_retention_days = -1` then running `rollup_cleanup()` deletes every row. Added `CHECK (>= 1)` on both columns in table definition and migration block.

**LOW** Install header missing 1.3→1.4 upgrade path — header listed 1.0→1.1→1.2→1.3 but not 1.3→1.4. Added.

---

### TESTS ADDED (3)

1. **rollup_hour different-length arrays** — seeds minutes with different wait/query array lengths, verifies rollup completes and merges correctly
2. **Retention CHECK constraint** — verifies `rollup_1m_retention_days = 0` and `rollup_1h_retention_days = -1` are rejected
3. **Uninstall verifies v1.4 objects** — checks rollup tables and functions exist before uninstall, confirms complete removal after

---

### VERIFIED OK (previously flagged items)

| Item | Status | Notes |
|------|--------|-------|
| `_truncate_pairs()` ORDER BY fix | ✅ Verified | Top-N correctly ordered by count DESC, id ASC tiebreak |
| `rotate()` xact-level lock | ✅ Verified | Lock released after transaction commit (no session leak) |
| `rollup_minute()` wait_event join | ✅ Verified | CPU\* and IO:DataFileRead both correctly mapped to wait_ids |
| `rebuild_partitions()` | ✅ Verified | 3→5→3 rebuild cycle, sampling_enabled=false after, all partitions created |
| `current_slot` bounds | ✅ Verified | Stays in `[0, num_partitions)` through full rotation cycle |
| Uninstall (DROP CASCADE) | ✅ Verified | All 66 objects dropped cleanly |
| Batch limit | ✅ Verified | `rollup_minute(2)` correctly limits to 2 minutes per call |
| Rollup hour aggregation | ✅ Verified (after fix) | Correct merge: samples summed, peak_backends max'd, wait/query counts merged |

### REMAINING NON-BLOCKING (from REV reviews)

These are deferred items that don't block merge but should be tracked:

- **MEDIUM** `rebuild_partitions()` drain loop proceeds unconditionally after 5s timeout — consider raising exception instead of silently proceeding
- **LOW** No CHECK on `current_slot` — application logic keeps it in bounds (verified), but constraint would be defense-in-depth
- **LOW** Blueprint still says "v1.5" instead of "v1.4" — cosmetic
- **INFO** Missing `COMMENT ON TABLE` for `rollup_1m` / `rollup_1h`
- **INFO** `GROUP BY d` in `daily_peak_backends_at` uses SELECT alias

---

**Fix branch:** `claude/test-review-pr31-ex5IU` (commit `8b32612`)
All local tests pass on PG 16.

---
_Generated by [Claude Code](https://claude.ai/code)_

sql/ash-install.sql
```diff
@@ -1084,6 +1488,674 @@ begin
...
+      v_hour_start,
+      datid,
+      sum(samples)::smallint,
+      max(peak_backends)::smallint,
+      ash._merge_wait_counts(
+        array_agg(wait_counts) filter (where wait_counts <> '{}')
+      ),
+      ash._truncate_pairs(
+        ash._merge_query_counts(
+          array_agg(query_counts) filter (where query_counts <> '{}')
```
NikolayS: **BUG (HIGH):** `rollup_hour()` crashes in production with `"cannot accumulate arrays of different dimensionality"`.

PostgreSQL's built-in `array_agg()` on array values creates a multidimensional array and **requires** all sub-arrays to have equal length. When different minutes have different numbers of wait events or queries (the normal case — e.g. minute 1 has CPU\*+IO = 4-element `wait_counts`, minute 2 has only CPU\* = 2-element), this crashes:

```sql
-- Reproduction:
INSERT INTO ash.rollup_1m VALUES
  (hour_ts,      0, 30, 5, '{1,60,2,40}'::int4[], '{12345,60,67890,20}'::int8[]),
  (hour_ts + 60, 0, 30, 3, '{1,50}'::int4[],      '{12345,40}'::int8[]);
-- ↑ 4 elements vs 2 elements

SELECT ash.rollup_hour();
-- ERROR: cannot accumulate arrays of different dimensionality
```

Both `wait_counts` (line 1854) and `query_counts` (line 1858) are affected.

**Fix:** Replace `array_agg()` with custom `_int4_array_cat_agg` / `_int8_array_cat_agg` aggregates that use `array_cat()` as sfunc to produce flat 1-D concatenation regardless of sub-array lengths. See fix branch `claude/test-review-pr31-ex5IU`.

---
_Generated by [Claude Code](https://claude.ai/code)_

sql/ash-install.sql
```diff
@@ -57,22 +57,54 @@ $$;
...
+  sample_interval            interval not null default '1 second',
+  rotation_period            interval not null default '1 day',
+  include_bg_workers         bool not null default false,
+  debug_logging              bool not null default false,
+  encoding_version           smallint not null default 1,
+  version                    text not null default '1.4',
+  rotated_at                 timestamptz not null default clock_timestamp(),
+  installed_at               timestamptz not null default clock_timestamp(),
+  rollup_1m_retention_days   smallint not null default 30,
+  rollup_1h_retention_days   smallint not null default 1825,
```
NikolayS: **MEDIUM — missing CHECK constraint:** Negative or zero `rollup_*_retention_days` causes `rollup_cleanup()` to delete **all** rollup data.

Confirmed by test: `UPDATE ash.config SET rollup_1m_retention_days = -1` then `SELECT ash.rollup_cleanup()` → every row in `rollup_1m` deleted. The cutoff becomes `now() + 1 day`, so all data is "older than cutoff."

**Fix:** Add `CHECK (rollup_1m_retention_days >= 1)` and `CHECK (rollup_1h_retention_days >= 1)`. Also needs the same CHECK in the migration DO block (line ~100). See fix branch.

---
_Generated by [Claude Code](https://claude.ai/code)_

sql/ash-install.sql
NikolayS, line 1: **LOW:** File header (lines 4-6) lists upgrade paths through 1.3 but omits 1.3→1.4. The upgrade script `ash-1.3-to-1.4.sql` exists but isn't mentioned. Should add `then \i sql/ash-1.3-to-1.4.sql` to each chain and a new `-- Upgrade from 1.3:` line. See fix branch.

---
_Generated by [Claude Code](https://claude.ai/code)_

sql/ash-install.sql
```diff
@@ -638,37 +780,131 @@ begin
...
+      select from pg_locks
+      where locktype = 'advisory'
+        and classid = 0
+        and objid = hashtext('ash_operation')::int4::oid
+        and granted
+    ) then
+      exit;
+    end if;
+
+    perform pg_sleep(0.5);
```
NikolayS, line 931: **MEDIUM (non-blocking, existing concern):** Drain loop proceeds unconditionally after 5s timeout. If `take_sample()` or rollup operations are still in-flight (holding the `ash_operation` advisory lock), `rebuild_partitions()` will DROP their target tables out from under them.

Consider: after the loop, check a `v_drained` flag and `RAISE EXCEPTION 'drain timeout — in-flight operations still active'` instead of proceeding silently. The caller can then retry.

Not fixing this in my branch since it's a design decision, but flagging for consideration.

---
_Generated by [Claude Code](https://claude.ai/code)_

.github/workflows/test.yml
NikolayS, line 1: **Test gaps found (LOW):**

1. **Uninstall test** (line ~1958) only checks schema is gone. Should verify v1.4 objects (`rollup_1m`, `rollup_1h`, `rollup_minute()`, `rebuild_partitions()`, etc.) exist *before* uninstall and are absent *after*, to confirm `DROP SCHEMA CASCADE` catches everything.

2. **rollup_hour() with different-length arrays** — no test covers the case where minute-level rollup rows have `wait_counts` / `query_counts` arrays of different lengths. This is the common case in production and triggers the `array_agg()` crash (see inline comment on `ash-install.sql`).

3. **Retention CHECK constraint** — no test verifies that `rollup_*_retention_days = 0` or negative values are rejected.

All three added in fix branch `claude/test-review-pr31-ex5IU`.

---
_Generated by [Claude Code](https://claude.ai/code)_

NikolayS (commented):
## Thorough Manual Review & Local Testing

Read all 4021 lines of `ash-install.sql`, 2397 lines of `test.yml`, and the 13-line upgrade script. Tested locally on PostgreSQL 16. Verified all 8 items from the prior 3 REV reviews.

---

### Blocking (1)

| Severity | File | Issue |
|----------|------|-------|
| **HIGH** | `ash-install.sql:1854,1858` | `rollup_hour()` crashes with `"cannot accumulate arrays of different dimensionality"` when minutes have different numbers of wait events/queries — the common production case |

### Non-blocking fixes (2)

| Severity | File | Issue |
|----------|------|-------|
| **MEDIUM** | `ash-install.sql:75-76` | No CHECK on `rollup_*_retention_days` — negative values delete all rollup data |
| **LOW** | `ash-install.sql:4-6` | Header missing 1.3→1.4 upgrade path |

### Verified OK from prior reviews (8)

- ✅ `_truncate_pairs()` ORDER BY fix — correct top-N with count DESC, id ASC tiebreak
- ✅ `rotate()` xact-level lock — no session leak, released on commit
- ✅ `rollup_minute()` wait_event join — CPU\* and IO:DataFileRead both correctly mapped
- ✅ `rebuild_partitions()` — 3→5→3 cycle, sampling_enabled=false, all partitions created
- ✅ `current_slot` bounds — stays in `[0, num_partitions)` through full rotation cycle
- ✅ Uninstall — all 66 objects dropped cleanly via CASCADE
- ✅ `rollup_minute()` batch limit — correctly limits processing rate
- ✅ `rollup_hour()` aggregation — correct merge (after fix applied)

### Remaining suggestions (non-blocking, for future)

- `rebuild_partitions()` drain loop proceeds after timeout — consider raising exception
- No CHECK on `current_slot` — app logic is correct but constraint would help
- Blueprint version mismatch (says v1.5, should be v1.4)
- Missing `COMMENT ON TABLE` for rollup tables

---

**All fixes + 3 new tests pushed to branch `claude/test-review-pr31-ex5IU` (commit `8b32612`).** Locally tested, all pass on PG 16.

---
_Generated by [Claude Code](https://claude.ai/code)_

NikolayS:
REV review

I did a thorough pass here: diff review, local manual install smoke test on PG16, local `v1.3 -> PR branch` upgrade test, and CI review.

## Manual testing

### Fresh install on PG16
Verified locally:
- `sql/ash-install.sql` installs cleanly
- `ash.status()` works and shows the new config/rollup fields
- `ash.take_sample()` works
- `ash.rebuild_partitions(5)` works
- `ash.status()` reflects the rebuilt partition count and disabled sampling state as expected

### Upgrade path from released `v1.3`
Verified locally:
- install from tag `v1.3`
- apply `sql/ash-1.3-to-1.4.sql`
- resulting version is `1.4`
- `ash.status()` works and exposes the new fields

### CI
Green on PG 14/15/16/17/18.

## Finding

### README upgrade section is missing the new upgrade step
`README.md` still shows upgrade examples only up to `1.2 -> 1.3`, but this PR introduces `sql/ash-1.3-to-1.4.sql`.

That means the release ships the upgrade script, but the README upgrade instructions do not mention it.

Suggested fix:

```sql
-- from 1.3 to 1.4
\i sql/ash-1.3-to-1.4.sql
```

## Verdict

One low-severity docs issue. Otherwise the PR looks solid and the manual tests passed.

NikolayS:
Fixed in `8eedb61` — added `\i sql/ash-1.3-to-1.4.sql` to the README upgrade section.

Good catch — the install header was fixed in the previous commit but the README was missed.

---
_Generated by [Claude Code](https://claude.ai/code)_

NikolayS:
I re-checked the implementation scope in this PR and tightened the test plan accordingly. This is what I think v1.4 needs before merge / release confidence.

## Super-thorough test plan for PR #31 / v1.4

### A. Install / upgrade / re-apply matrix
For PG14/15/16/17/18:
- fresh install from `sql/ash-install.sql`
- upgrade from v1.3 using `sql/ash-1.3-to-1.4.sql`
- re-apply `ash-install.sql` after upgrade
- re-apply upgrade script again (idempotency)
- schema equivalence check: fresh v1.4 vs upgraded-to-v1.4
- confirm compat alias `_to_sample_ts()` does not break old upgrade paths

### B. Config columns and defaults
Verify on fresh install + upgrade path:
- `num_partitions` exists, default 3, constrained to 3..32
- `sampling_enabled` exists, default true
- `skipped_samples` exists, default 0
- rollup config columns exist with intended defaults
- `last_rollup_1m_ts` / `last_rollup_1h_ts` semantics are sane
- version is 1.4 everywhere (value + default)

### C. Dynamic partition behavior
Test partition counts beyond the happy path:
- default 3
- minimum 3
- medium value like 5
- larger value like 16
- max 32

For each chosen N:
- `current_slot()` stays within 0..N-1
- `_active_slots()` matches expected rotation window
- `take_sample()` writes into the correct `query_map_%N` partition
- `rotate()` wraps correctly with modulo-N arithmetic
- truncation hits the correct old partition
- `query_map_all` view is rebuilt correctly and returns all partitions

### D. rebuild_partitions() safety
This is one of the riskiest parts. Need explicit coverage for:
- 3 -> 5 -> 3 round-trip
- sampling disabled at the end (explicit restart required)
- raw samples are cleared as documented
- rollup tables survive rebuild
- query_map view is rebuilt correctly
- rebuild while idle
- rebuild after sampling has happened
- rebuild when sampling job is active / advisory locks present
- rebuild failure path does not leave system wedged
- status after rebuild is understandable

### E. Advisory locking contract
Because this PR introduces coordination complexity, validate:
- `take_sample()` participates in `ash_operation` lock
- rebuild takes exclusive rebuild lock
- in-flight operation drain behavior works as intended
- xact-level locks really disappear on rollback/error
- no leaked session-level locks remain after failures
- `rotate()` lock behavior remains safe under pg_cron connection reuse

### F. skipped_samples semantics
Need crisp meaning and tests:
- increments when sampling is disabled and a sample attempt occurs
- does not increment on successful sampling
- survives stop/start
- survives rebuild
- visible in `ash.status()`
- distinguishable from `missed_samples`

### G. Rollup correctness (exact values, not just row existence)
Seed deterministic raw samples and assert exact outputs for:

#### rollup_minute()
- correct sample count
- correct peak_backends
- correct wait_counts contents
- correct query_counts contents
- correct watermark advancement
- idempotent re-run (no double counting)

#### rollup_hour()
- aggregates minute rollups correctly
- merged wait/query counts exact
- peak logic exact
- idempotent re-run

#### rollup_cleanup()
- old rows deleted
- recent rows preserved
- retention config respected

### H. Rollup boundary / gap cases
These are where hidden bugs usually live:
- empty raw data
- exactly one minute of data
- minute boundary crossing
- hour boundary crossing
- sparse samples
- missing minutes / gaps
- data already rotated away before rollup runs
- rollup after rebuild
- rollup with `pg_stat_statements` unavailable

### I. Reader regression sweep
For both fresh install and upgraded DBs, verify these run cleanly:
- existing readers (top_waits, top_queries, top_queries_with_text, top_by_type, activity_summary, samples, `_at` variants)
- new rollup readers (`minute_waits`, `hourly_queries`, `daily_peak_backends` + `_at`)

Important cases:
- no pg_stat_statements
- empty ranges return 0 rows, not errors
- future ranges return 0 rows cleanly
- rollup readers on seeded deterministic data return expected values

### J. pg_cron and non-pg_cron modes
Need both:

#### With pg_cron
- start() creates jobs
- stop() removes jobs
- rotate() + rollup interaction under scheduled execution
- rollup jobs don’t interfere with sampling jobs

#### Without pg_cron
- install works
- manual operation works
- rollup functions callable manually
- no hidden pg_cron dependency in v1.4 features

### K. Performance / operational sanity
Not full benchmarking, but enough to catch obvious regressions:
- rollup_minute() on a realistic chunk of seeded data
- rebuild_partitions() runtime for 3->32 and back on non-empty system
- no pathological query plans from new dynamic SQL / unions
- status() remains readable and cheap enough to call

### L. Documentation / release readiness
Before merge/tag:
- README defaults/config table matches actual defaults
- rebuild_partitions() documented as destructive to raw samples
- rollup retention/storage estimates look sane
- upgrade instructions are accurate
- release notes clearly explain operational consequences of rollups + configurable partitions

## My view after looking again
This PR is ambitious enough that “CI green” is necessary but not sufficient.
The highest-risk areas are:
1. rebuild_partitions() safety and recovery
2. rollup correctness across boundaries/gaps
3. upgrade/re-apply/schema-equivalence edge cases
4. operational interaction between sampling, rotation, and rollup jobs

If those four areas are solid, then v1.4 looks real.

NikolayS:
I started executing the test plan locally against this PR and immediately hit a blocking install/idempotency problem on PostgreSQL 17.

## BLOCKING

`sql/ash-install.sql` is not safely re-applicable on a database that already has the earlier `missed_samples` work installed.

### Repro
On a DB where `ash.config.missed_samples` already exists, running:

```sql
\i sql/ash-install.sql
```

fails with:

```text
ERROR:  column "missed_samples" of relation "config" already exists
```

### Why it happens
The migration guard checks only for `num_partitions`:

```sql
if not exists (
  select from information_schema.columns
  where table_schema = 'ash' and table_name = 'config'
    and column_name = 'num_partitions'
) then
  alter table ash.config
    add column num_partitions ...,
    add column sampling_enabled ...,
    add column skipped_samples ...,
    add column missed_samples bigint not null default 0,
    ...
end if;
```

But on a DB that already has the PR #29 / v1.4-prep `missed_samples` column, `num_partitions` is absent while `missed_samples` is present, so the bundled `ALTER TABLE ... ADD COLUMN missed_samples ...` explodes.

### Proof from local run
Environment:
- PR branch: `pr-31` / `3f99278`
- PostgreSQL: `17.9`

Observed error:

```text
psql:sql/ash-install.sql:106: ERROR:  column "missed_samples" of relation "config" already exists
```

### What needs fixing
This migration block needs to be made resilient to partial-v1.4 state, not just pure-v1.3 state.

At minimum:
- use per-column `ADD COLUMN IF NOT EXISTS`, or
- split the migration into separate guarded statements, or
- explicitly handle the already-has-`missed_samples` / missing-new-columns case.

Until that is fixed, I would not call the install/upgrade path release-ready.

NikolayS:
Follow-up after retesting the **current** PR head (`3f99278`): this blocker is still real.

I re-ran the repro on isolated databases to avoid contamination from previous local state.

## Repro used
1. Fresh DB `pgash_pr31_partial`
2. Create only a minimal pre-#30 `ash.config` with:
   - existing `missed_samples`
   - **without** `num_partitions`
3. Run current `sql/ash-install.sql`

## Result on current head
Still fails with the same error:

```text
ERROR:  column "missed_samples" of relation "config" already exists
```

The fresh-install path succeeds. The partial-v1.4 / re-apply path does not.

## Why the current fix is still insufficient
The guard still checks only for `num_partitions` before issuing one bundled `ALTER TABLE ... ADD COLUMN ...` that includes `missed_samples`.

So this intermediate state is still broken:
- `missed_samples` already exists
- `num_partitions` does not exist

That is exactly the state users can hit when moving from the earlier missed-samples work toward this PR.

## Recommendation
This migration block needs per-column resilience, e.g.:
- separate `ADD COLUMN IF NOT EXISTS` statements, or
- a more careful partial-state migration path

Until that is fixed, I still consider the install/upgrade path not release-ready.

NikolayS:
# REV Review: PR #31 — Configurable N-partitions and rollup tables (v1.4)

**Scope**: 6 files, +3489 / -505 lines. Introduces configurable N-partitions (3-32), rollup tables (`rollup_1m`, `rollup_1h`), watermark-based rollup functions, `rebuild_partitions()`, and a two-key advisory lock protocol.

**CI status**: All 7 checks pass (Analyze, CodeQL, test(14/15/16/17/18)).

---

## BLOCKING

### HIGH

#### [bug-hunter] `rollup_minute()` wait-event join double-counts across `state` values
- **severity**: HIGH
- **confidence**: 8
- **file**: `sql/ash-install.sql`
- **line**: ~1576-1585 (`wait_agg` CTE inside `ash.rollup_minute`)
- **issue**: The join from decoded samples to `ash.wait_event_map` matches only on `(type, event)`, ignoring `state`. Since `ash.wait_event_map` has `UNIQUE (state, type, event)` (not `(type, event)`), identical `(type, event)` pairs under different `state` values produce separate `wm.id` rows. A decoded row whose text rendering is `'Type:Event'` matches BOTH rows, doubling the count in `rollup_1m.wait_counts`.
- **evidence**:
```sql
join ash.wait_event_map wm
  on d.wait_event = case when wm.event = wm.type then wm.event
                          else wm.type || ':' || wm.event end
group by d.datid, wm.id
```
  Concrete example: `('active','Client','ClientRead')` and `('idle in transaction','Client','ClientRead')` are both legal entries in `wait_event_map` (both states can report `wait_event='ClientRead'`). `ash.decode_sample()` (line 578) emits `wait_event = 'Client:ClientRead'` for both, so the join matches both `wm` rows. Every decoded occurrence is counted twice with two different `wait_id`s, so `rollup_1m.wait_counts` becomes inflated AND split — and downstream `minute_waits()` percentages reference a doubled grand total.
- **fix**: Either (a) change `ash.decode_sample()` to return `wait_id` (smallint) directly alongside `wait_event`, and have `rollup_minute()` group by the returned `wait_id`, or (b) derive `wait_id` in `rollup_minute` by reading the negative sentinel directly from `ash.sample.data` (bypassing `decode_sample`), which preserves state-specific IDs. Add a regression test that seeds two `(state, type, event)` rows sharing `(type, event)` and asserts `rollup_1m.wait_counts` is not doubled.

---

## NON-BLOCKING

### MEDIUM

#### [test-analyzer] `rollup_minute()` has no test for concurrent-state wait events
- **severity**: MEDIUM
- **confidence**: 8
- **file**: `.github/workflows/test.yml`
- **line**: "Test v1.4: rollup tables and functions" block (lines ~372-546)
- **issue**: The rollup test only registers a single wait event (`('active','CPU*','CPU*')`) and inserts three identical samples. There is no case for (a) multiple states producing the same `(type, event)` text, (b) multiple databases (`datid`), (c) wait events where `type != event` (the `type:event` rendering branch), (d) non-NULL query_ids in rollup, or (e) the `rollup_min_backend_seconds` filter boundary. Without these, the join bug above would not be caught, nor would any regression in the paired-array encoding.
- **untested_scenarios**:
  - Two `wait_event_map` rows with the same `(type, event)` but different `state` — rollup must NOT double-count.
  - `rollup_min_backend_seconds` boundary: query with exactly `v_min_backend_seconds - 1` and `= v_min_backend_seconds` samples.
  - Multi-database sample: assert `peak_backends` is computed per-datid.
  - Gap detection: minute with no samples but later data exists — assert WARNING is raised (use `pg_notify`/capture NOTICE).
  - `query_counts` top-100 truncation: generate 150 distinct `query_id`s in one minute and assert only top 100 survive.

#### [bug-hunter] `rollup_minute()` retries re-process same minute after `last_rollup_1m_ts = null`, but `rollup_hour()` watermark advances only `v_hour_end` — watermark can't rewind
- **severity**: MEDIUM
- **confidence**: 6
- **file**: `sql/ash-install.sql`
- **line**: rollup_hour / rollup_minute watermark update
- **issue**: Operationally, if minute rollups are re-run (test/remediation case), they can be replayed by nulling `last_rollup_1m_ts`. Hourly watermark can similarly be nulled. But there is no guard against `last_rollup_1m_ts` being ahead of the actual `min(ts)` in `rollup_1m` after a rows-were-deleted scenario (e.g., `rollup_cleanup()` removed the oldest minute). The watermark initialization path only runs when watermark IS NULL. After cleanup deletes the oldest rows, subsequent `rollup_hour()` that NULLed its watermark would pick `min(ts)` (now advanced) — correct. But if `last_rollup_1h_ts` is older than the current `min(ts)` in `rollup_1m` (data was deleted), the loop iterates through hours with no data, advancing watermark one-by-one — wasteful but not incorrect. Consider documenting the recovery procedure.
- **fix**: Document in the function comment that setting `last_rollup_1m_ts = null` (or `last_rollup_1h_ts`) triggers re-initialization from the current oldest row. Also consider clamping watermark to `max(last_rollup_ts, (select min(ts) from rollup_src))` at loop entry.

### LOW

#### [bug-hunter] Upgrade migration DO-block ADD COLUMN fails if `missed_samples` already exists
- **severity**: LOW
- **confidence**: 7
- **file**: `sql/ash-install.sql`
- **line**: ~82-103 (migration DO block gated on `num_partitions` absence)
- **issue**: The DO block adds `missed_samples` inside a multi-column `ALTER TABLE ... ADD COLUMN` (no `IF NOT EXISTS` per column in the list). If a user ran an older pre-merge `ash-1.3-to-1.4.sql` (PR #29) that already added `missed_samples`, then re-runs the new `ash-install.sql`, the ALTER will fail on `missed_samples` because `num_partitions` is absent (triggering the block) but `missed_samples` is present. 1.4 was never tagged, so this is an edge case for early adopters / CI pipelines.
- **fix**: Split the ALTER into individual `ADD COLUMN IF NOT EXISTS` statements (PG 9.6+), or test column existence individually before adding. Given the narrow window before v1.4 release, accepting this risk is defensible — but flag it in release notes for anyone who cherry-picked PR #29 ahead of #31.

#### [bug-hunter] `take_sample()` inside `rotate()` / rollup can be starved by advisory lock contention; `skipped_samples` is incremented both for disabled-sampling and for lock-miss, conflating two very different conditions
- **severity**: LOW
- **confidence**: 7
- **file**: `sql/ash-install.sql`
- **line**: ~407-427 (`take_sample` early returns)
- **issue**: Both `not v_sampling_enabled` and `not pg_try_advisory_xact_lock(0, hashtext('ash_operation')::int4)` increment `skipped_samples`. Operators reading `status()` cannot distinguish "sampling is paused for rebuild" from "rollup/rotate is currently holding the lock". During a rotation with a large catch-up (`rollup_minute(1440)` on a 1-day rotation), `rotate()` could hold `ash_operation` for seconds and every concurrent `take_sample()` bumps the counter — easily mistaken for a bug.
- **fix**: Add a second counter (`skipped_locked_samples`) or emit the reason via a separate metric. Alternatively, since `rotate()` already holds `ash_rotate`, have `take_sample()` distinguish the two paths in `status()` output.

#### [guidelines-checker] Whitespace/indent in test.yml helper hunk (lines 311-369) diverges from surrounding style
- **severity**: INFO
- **confidence**: 5
- **file**: `.github/workflows/test.yml`
- **line**: ~311-369 (nested `declare`/`begin` inside the rotation test)
- **issue**: The rotation test's inner `declare v_num_partitions smallint := (select ...)` block is indented at the same level as the outer `begin`, unlike the existing style in the same file. Minor — readability only. Not a rule violation per the SQL style guide (style guide addresses SQL, not embedded heredoc scripts), but inconsistent with neighbors.
- **fix**: Indent the inner `declare` one level further to mirror the surrounding block pattern.

### INFO

#### [docs-reviewer] `README.md` describes `ash.stop()` as setting `sampling_enabled=false`, but does not warn that this means rollups still run (per design table in blueprint). Minor.
- **severity**: INFO
- **confidence**: 5
- **file**: `README.md`
- **line**: API reference table (around `ash.stop()`)
- **issue**: The spec (`blueprints/CONFIGURABLE_PARTITIONS_AND_ROLLUP.md` §sampling_enabled semantics) explicitly states rollups *still run* when `sampling_enabled = false`. README says "Stop sampling and rollups". This is somewhat misleading because `stop()` unschedules the pg_cron rollup jobs (so they don't *auto-run*), but manual `ash.rollup_minute()` invocation from external schedulers still works.
- **suggestion**: Clarify to "Stop sampling and unschedule rollup cron jobs (manual rollup calls still work)."

#### [docs-reviewer] `rebuild_partitions()` WARNING comment lives above the function body but not in user-facing docs
- **severity**: INFO
- **confidence**: 5
- **file**: `README.md`
- **line**: `ash.rebuild_partitions(N)` API row
- **issue**: The source has a prominent `WARNING: failure after acquiring lock leaves sampling_enabled = false. Manual recovery: ...` comment, but README merely says `Destructive — all raw sample data is lost. Rollup tables survive. Call ash.start() after to resume`. Operators who hit a failure mid-rebuild won't find the recovery procedure in README.
- **suggestion**: Add a short note: "If rebuild fails mid-way, sampling remains disabled; inspect `ash.status()` and manually `UPDATE ash.config SET sampling_enabled = true` before `ash.start()`."

---

## POTENTIAL

### Medium confidence (4-7)

#### [bug-hunter] `minute_waits_at()` / `hourly_queries_at()` pair-extraction via `row_number() over ()` without explicit ORDER BY
- **confidence**: 5
- **file**: `sql/ash-install.sql`
- **line**: `ash.minute_waits_at`, `ash.hourly_queries_at` (`raw_pairs` CTE)
- **issue**: `row_number() over ()` with no ORDER BY is non-deterministic per SQL spec, though PostgreSQL's implementation typically preserves nested-loop output order such that per-row `unnest` output is contiguous. Because each `wait_counts`/`query_counts` array has even length (invariant enforced by helpers), the parity-based pairing survives cross-row boundaries. Still, relying on implementation-defined ordering is fragile — a future parallel plan could break it. Likely fine today on PG 14-18 (no parallelism on STABLE function bodies with lateral), but worth documenting the assumption.
- **fix**: Either add `order by` (via a stable per-row identifier such as `ts, datid` + sub-position), or convert to `unnest(...) with ordinality` (at the array level, then compute a per-row offset). The `with ordinality` pattern is cleaner and explicit.

#### [bug-hunter] `rollup_hour()` filters on `wait_counts <> '{}'` but `_merge_wait_counts(NULL)` path differs — minor invariant concern
- **confidence**: 4
- **file**: `sql/ash-install.sql`
- **line**: `rollup_hour()` insert
- **issue**: If every `rollup_1m` row in the hour has an empty `wait_counts`, the filtered `array_agg` produces NULL. `_merge_wait_counts(NULL)` returns `'{}'::int4[]` (OK). But if some rows have `NULL` wait_counts (not empty array), the column definition is `NOT NULL`, so this can't happen. Still, the `wait_counts <> '{}'` filter doesn't guard against NULL per se. Defensive coding: add `wait_counts is not null and wait_counts <> '{}'`.
- **fix**: Tighten filter to `wait_counts is not null and wait_counts <> '{}'`.

#### [test-analyzer] No test for `rebuild_partitions()` failure mid-operation
- **confidence**: 6
- **file**: `.github/workflows/test.yml`
- **line**: rebuild_partitions test block
- **issue**: The WARNING explicitly calls out that failure leaves `sampling_enabled = false`. There is no test that simulates a mid-operation failure (e.g., disk-full or intentional exception inside step 7) to verify recovery instructions are accurate.
- **untested_scenarios**: Force an exception during partition creation; verify (a) `sampling_enabled` is still `false`, (b) rebuild lock is released (xact auto-release), (c) manual recovery via `UPDATE + start()` restores function.

#### [test-analyzer] Drain logic (`pg_locks` poll in `rebuild_partitions`) is not end-to-end exercised
- **confidence**: 5
- **file**: `.github/workflows/test.yml`
- **line**: advisory-lock visibility test only
- **issue**: The advisory lock visibility test verifies the LOCK is observable. But there is no test that actually starts a concurrent `take_sample()` holding `ash_operation`, fires `rebuild_partitions`, and verifies drain behaviour. Two-session tests in a single psql heredoc are hard, but `pg_background` or a `DO` with `dblink` would work.
- **untested_scenarios**: Concurrent `take_sample()` + `rebuild_partitions()` race; verify `rebuild_partitions()` drains before dropping tables and does not race with the in-flight sample insert.

#### [security-reviewer] `execute format(...)` in `take_sample`/`rotate`/`rebuild_partitions` passes `v_current_slot`/`v_truncate_slot`/`i` as `%s`
- **confidence**: 5
- **file**: `sql/ash-install.sql`
- **line**: multiple dynamic SQL sites
- **issue**: The `%s` parameters are `smallint` or `int` typed local variables derived from config or modulo arithmetic — not from user input — and thus cannot contain SQL payloads. Not exploitable. Flagging only because dynamic SQL is a common footgun; the mitigation (typed locals) is correct and matches the spec's `format()` contract.
- **fix**: None required. Leave a code comment at each site reaffirming that `%s` is safe because the slot is smallint-typed.

---

## Summary

| Area | Blocking | Non-blocking | Potential |
|------|----------|--------------|-----------|
| Security | 0 | 0 | 1 |
| Bugs | 1 | 2 | 2 |
| Tests | 0 | 0 | 3 |
| Guidelines | 0 | 1 | 0 |
| Docs | 0 | 2 | 0 |

**Top priority**:
1. `rollup_minute()` wait-event join ignores `state`, causing double-counting when the same `(type, event)` exists under multiple states. HIGH, confidence 8.
2. Add regression test covering multiple `(state, type, event)` rows sharing `(type, event)`.
3. Minor: migration DO block not idempotent if `missed_samples` was added by a pre-merge upgrade script.

**Verdict**: BLOCKING — the rollup double-count bug will corrupt long-term trend data for any workload with `Client:ClientRead` (or similar cross-state wait events) and is silent until someone notices percentages summing to >100%.

Code style is clean (lowercase SQL, snake_case, follows project conventions). CI passes on PG 14-18. Documentation is thorough (blueprint + README + comments). Advisory lock protocol is well-thought-out. The main risk is the silent aggregation bug.

---

NikolayS:
**Verdict:** STILL BLOCKING — original bug is fixed, but the fix introduces a new upgrade-path regression that is failing CI on PG 14–18.

I verified the `rollup_minute()` change: `decode_sample()` now emits a `wait_id smallint` OUT column, `wait_agg` groups on `d.wait_id` directly with no text join against `wait_event_map`, and the canonical-id semantics are preserved. I scanned every other `wait_event_map` join in `sql/ash-install.sql` (samples_at, top_waits, wait_timeline, query_waits, etc.) — all already use `wm.id = <decoded>.wait_id`, so the state-collapsing bug was localized to `rollup_minute()` and is now fully addressed. The new regression test is well-targeted: it registers `('active','Client','ClientRead')` and `('idle in transaction','Client','ClientRead')`, seeds three samples referencing only the active wait_id, and asserts (a) `wait_counts` total is 3, not 6, and (b) the idle wait_id is absent — a text-join regression would fail both asserts. The migration DO block refactor to per-column `ADD COLUMN IF NOT EXISTS` also looks correct.

**Blocking finding (confidence 9):**

- **CI red on all PG versions (14–18).** The 1.3 → 1.4 upgrade job fails at `sql/ash-install.sql:711` with:
  ```
  ERROR:  cannot change return type of existing function
  DETAIL:  Row type defined by OUT parameters is different.
  HINT:  Use DROP FUNCTION ash.decode_sample(integer[],smallint) first.
  ```
  `CREATE OR REPLACE FUNCTION` cannot change a function's OUT-parameter signature. Since the existing 1.3 `ash.decode_sample(integer[], smallint)` has `(wait_event text, query_id int8, count int)` and the 1.4 version has `(wait_id smallint, wait_event text, query_id int8, count int)`, every in-place upgrade will break. `ash-install.sql` needs an explicit `DROP FUNCTION IF EXISTS ash.decode_sample(integer[], smallint);` before the `CREATE OR REPLACE` (and the eventual `ash-1.3-to-1.4.sql` upgrade script must do the same). Fresh installs pass only because there's no prior function to conflict with.

**Minor (confidence 7):** The regression test seeds `datid = 0::oid` and `slot = 0` directly without a corresponding `ash.query_map` entry — works here because `map_id = 0` is the NULL sentinel, but the test relies implicitly on that and on `v_current_slot = 0`. Not a defect, just fragile if the sample encoding ever changes.

Once the decode_sample drop-before-create is added and CI goes green, I expect this to be RESOLVED.

NikolayS:
**Verdict: STILL BLOCKING**

The DROP added in commit `1036226` is correctly placed in `sql/ash-install.sql:612` (immediately before the `create or replace function ash.decode_sample(...)` on line 613) and targets the right signature `(integer[], smallint)`. The original rollup fix (join on `wait_id` instead of `(type, event)`) is still intact at `sql/ash-install.sql:1710–1720`. However, the fix is incomplete: **CI is still red on all 5 matrix jobs (PG 14, 15, 16, 17, 18)** — see run `24870478195`. The failure is not in `ash-install.sql`, it's in `sql/ash-1.2-to-1.3.sql:358`, which also contains `create or replace function ash.decode_sample(p_data integer[], p_slot smallint default null)` with the OLD 3-column RETURNS TABLE shape and has no preceding DROP. The "Dev workflow: re-apply migrations (idempotent)" step in `.github/workflows/test.yml` runs `\i sql/ash-install.sql` (now installs the 4-col version) and then re-applies `ash-1.1-to-1.2.sql` and `ash-1.2-to-1.3.sql`; the latter trips the exact same error the prior fix was meant to resolve:

```
psql:sql/ash-1.2-to-1.3.sql:455: ERROR:  cannot change return type of existing function
DETAIL:  Row type defined by OUT parameters is different.
HINT:  Use DROP FUNCTION ash.decode_sample(integer[],smallint) first.
```

Remaining concerns (confidence 10): (1) either add the same `drop function if exists ash.decode_sample(integer[], smallint);` line immediately before line 358 of `sql/ash-1.2-to-1.3.sql`, or — if the project convention is that historical upgrade scripts are frozen — leave 1.2-to-1.3 alone and remove it from the idempotent-replay test, but document why. (2) Worth considering the same concern for `sql/ash-1.1.sql:431` and `sql/ash-1.0.sql:406`, though those aren't replayed by CI so they won't fail the build; they would bite anyone stepping through the historical install path after a fresh install. No other OUT-parameter shape changes in the PR diff require additional DROPs — spot-checked; the other three pre-existing DROPs in install.sql (`top_queries_with_text`, `event_queries`, `event_queries_at`) are unchanged from main.

NikolayS:
## REV follow-up 3 — RESOLVED

**Verdict:** RESOLVED

Fix 3 (`a11dd18`) takes a different (and correct) approach: instead of widening `decode_sample()`'s OUT signature — which broke legacy re-apply of `ash-1.2-to-1.3.sql` in the prior two attempts — it leaves `decode_sample()` byte-identical to `origin/main` (verified via `git diff origin/main -- sql/ash-install.sql`, no function-body hunks for `decode_sample`) and instead rewrites `rollup_minute()`'s `decoded` CTE and `query_agg` CTE to walk `ash.sample.data[]` inline with the same `generate_subscripts + generate_series` pattern already used by `ash.samples()`. Aggregation then groups directly on the canonical `wait_event_map.id` via `(-s.data[i])::smallint` with no text join, so the original double-count bug stays fixed. The ClientRead state-disambiguation regression test is preserved (registers both `active` and `idle in transaction` variants, seeds only the `active` one, asserts `sum(wait_counts[even]) = 3` and that the idle id never leaks).

**CI:** all green on head `a11dd18` — CodeQL, Analyze, and test (14/15/16/17/18) all pass (~4–5 min each). Idempotent re-apply of `ash-1.2-to-1.3.sql` now works since the `decode_sample(integer[], smallint) -> table(wait_event text, query_id int8, count int)` signature matches exactly.

**Additional verification performed:**
- Inline walk in `rollup_minute.decoded` is structurally identical to `ash.samples().decoded`: same `s.data[i] < 0` wait-marker gate, same `greatest(data[i+1]-1, -1)` count expansion, same `i + 2 + gs.n <= array_length(...)` bounds guard.
- `query_agg` joins `ash.query_map_all qm on qm.slot = d.slot and qm.id = d.map_id` — slot is carried through `decoded` from `s.slot`, so post-rotation slot affinity is preserved (matches `ash.samples()` exactly).
- `map_id = 0` sentinel is excluded via `where d.map_id <> 0 and qm.query_id is not null`.
- `HAVING count(*) >= v_min_backend_seconds` and `row_number()` over filtered set work as intended (window runs after HAVING).
- `base.samples = count(distinct sample_ts)` per datid, `base.peak_backends = max(active_count)` per datid — correct since `take_sample()` writes one `ash.sample` row per (sample_ts, datid) with a per-datid `active_count`.
- `coalesce(qa.query_counts, '{}')` handles the all-idle-backend case (all `map_id = 0`).

**Remaining concerns:** none with confidence ≥7. The rewrite is a structural copy of an already-battle-tested pattern and all CI jobs pass.

PR #36, opened 2026-04-19
docs: improve ash.sample discoverability (#35)

### LINKED ISSUE #35: Raw ash.sample schema trips users into silent failures; suggest README example + flat view
(opus 4.7 agent coming from a benchmark)

## Summary

While running a benchmark with `pg_ash` sampling at 1 Hz, I tried to dump a wait-event breakdown at the end of the run. My natural instinct was to `JOIN ash.sample` against a wait-event lookup table. This fails in two ways that are hard to discover from the schema alone, and in my case the failure was silent because the `COPY` ran inside a shell pipeline with `2>/dev/null`. By the time I noticed no CSV had been produced, the benchmark VM had been wiped.

This isn't a correctness bug in `pg_ash` — it's a discoverability / DX issue. Filing it so others don't lose data the same way.

## What I tried

```sql
COPY (
  SELECT sample_ts, COALESCE(we.wait_event_type,'CPU') wet,
         COALESCE(we.wait_event,'CPU') we, count(*) cnt
  FROM ash.sample s LEFT JOIN ash.wait_event we ON s.wait_event_id = we.id
  GROUP BY 1,2,3
) TO '/tmp/ash.csv' CSV HEADER;
```

## What I expected

A CSV at `/tmp/ash.csv` with one row per `(sample_ts, wait_event_type, wait_event)` bucket.

## What actually happened

The query errored immediately and no file was written. Because stderr was redirected to `/dev/null` in the surrounding shell pipeline, I never saw the error. I only realized the dump had failed after the source data was already gone.

## Root cause

Two independent schema mismatches:

1. **Wrong table name.** There is no `ash.wait_event` — the actual lookup table is `ash.wait_event_map`.
2. **No `wait_event_id` column on `ash.sample`.** The `ash.sample` table stores a packed `integer[]` in its `data` column: position 1 is `-active_count`, and positions 2+ are `(wait_event_id, sample_count)` pairs. You can't join it directly. You have to decode via `ash.decode_sample(data, slot)` or, more simply, use the higher-level `ash.samples()` SRF.

Neither of these is obvious from `\\d ash.sample` or from a casual read of the README — `ash.sample` looks like a normal fact table, so users reach for a normal `JOIN`.

## The query that actually works

```sql
COPY (
  SELECT sample_time, database_name, active_backends, wait_event, query_id
  FROM ash.samples(p_interval => '1 hour'::interval, p_limit => 1000000)
) TO '/tmp/ash.csv' CSV HEADER;
```

Runs in ~0.3s on my dataset and produces the expected output.

## Proposed improvements

1. **README: add a prominent \"How to dump wait events for offline analysis\" section** with the canonical `ash.samples()`-based query above. This is the most common ad-hoc task and currently has no worked example.
2. **Annotate `ash.sample`**: a `COMMENT ON TABLE ash.sample` and/or `COMMENT ON COLUMN ash.sample.data` explaining that `data` is a packed `integer[]` (position 1 = `-active_count`, positions 2+ = `(wait_event_id, sample_count)` pairs) and pointing users to `ash.samples()` / `ash.decode_sample()` instead of raw reads. This would show up in `\\d+ ash.sample` and in any schema-introspection tool.
3. **Consider adding `ash.samples_flat` as a view** with the same output schema as the `ash.samples()` function. This would let users write the more idiomatic `SELECT ... FROM ash.samples_flat WHERE sample_time > now() - interval '1 hour'` for ad-hoc SQL, rather than having to remember the function signature and its parameter names.
4. **Surface default parameter values.** `ash.samples(p_interval => ..., p_limit => ...)` defaults `p_interval` to 1 hour, which isn't obvious without reading the function source. A one-line mention in the README (or in the `COMMENT ON FUNCTION`) would help.

Happy to send a PR for (1) and (2) if that's useful.

NikolayS:
Closes #35 (partially — implements proposals 1, 2, 4; skipping 3 per discussion).

## Summary

- Adds `COMMENT ON TABLE/COLUMN ash.sample` so `\d+ ash.sample` tells users the data column is packed and redirects them to `ash.samples()` / `ash.decode_sample()`.
- Adds `COMMENT ON FUNCTION` for `ash.samples`, `ash.samples_at`, and `ash.decode_sample`, documenting default parameter values (`p_interval='1 hour'`, `p_limit=100`) and the pg_stat_statements dependency.
- README gains a dedicated "Dump samples to CSV" section with worked `\copy` examples.

Skipping proposal 3 (`ash.samples_flat` view): `ash.samples()` uses its interval argument to prune partitions and limit index scans — a predicate-based view would likely defeat that pushdown and be slower. Better docs is the right fix, not a second API surface.

## Test plan

- [ ] CI green across PG 14–18
- [ ] `\d+ ash.sample` shows the table/column comments after install
- [ ] `\df+ ash.samples` shows the function comment

### DIFF
--- a/README.md
+++ b/README.md
@@ -293,6 +293,25 @@ select * from ash.samples('10 minutes', 20);
 select * from ash.samples_at('2026-02-14 03:00', '2026-02-14 03:05', 50);
 ```
 
+### Dump samples to CSV
+
+Always go through `ash.samples()` / `ash.samples_at()` — the underlying `ash.sample`
+table stores a packed `integer[]` and cannot be joined directly. The defaults for
+`ash.samples()` are `p_interval => '1 hour'` and `p_limit => 100`; pass a large
+`p_limit` when dumping.
+
+```sql
+-- dump every sample from the last hour
+\copy (select * from ash.samples('1 hour'::interval, 10000000)) to '/tmp/ash.csv' csv header
+
+-- dump a specific incident window
+\copy (select * from ash.samples_at('2026-02-14 03:00', '2026-02-14 03:05', 10000000)) to '/tmp/incident.csv' csv header
+```
+
+Use `\copy` (psql) rather than server-side `COPY TO` if `/tmp` isn't writable by
+the Postgres user (managed services), and check the exit status — silently
+redirecting stderr to `/dev/null` will hide errors like typos in table names.
+
 ### Timeline chart
 
 Visualize wait event patterns over time — spot spikes, correlate with deployments, see what changed.
--- a/sql/ash-install.sql
+++ b/sql/ash-install.sql
@@ -142,6 +142,12 @@ create index if not exists sample_0_datid_ts_idx on ash.sample_0 (datid, sample_
 create index if not exists sample_1_datid_ts_idx on ash.sample_1 (datid, sample_ts);
 create index if not exists sample_2_datid_ts_idx on ash.sample_2 (datid, sample_ts);
 
+comment on table ash.sample is
+$$Packed wait-event samples. One row per (sample_ts, datid). Do not join directly — use ash.samples() / ash.samples_at() for decoded rows, or ash.decode_sample(data, slot) if you need a single sample's contents.$$;
+
+comment on column ash.sample.data is
+$$Packed int4[] encoding the sample's wait events and their query_map ids. Layout: groups of (-wait_id, count, query_map_id_1, ..., query_map_id_count). A negative marker starts each group; wait_id is negated for the marker so a positive count cannot be mistaken for a group boundary. Decode via ash.samples(), ash.samples_at(), or ash.decode_sample(data, slot).$$;
+
 -- Register wait event function (upsert, returns id)
 create or replace function ash._register_wait(p_state text, p_type text, p_event text)
 returns smallint
@@ -2640,6 +2646,15 @@ begin
   end if;
 end;
 $$;
+
+comment on function ash.samples(interval, int) is
+$$Decoded wait-event samples over the last p_interval (default '1 hour'), newest first, up to p_limit rows (default 100). Returns (sample_time, database_name, active_backends, wait_event, query_id, query_text). query_text requires pg_stat_statements; NULL otherwise.$$;
+
+comment on function ash.samples_at(timestamptz, timestamptz, int) is
+$$Decoded wait-event samples between p_start and p_end, newest first, up to p_limit rows (default 100). Same return shape as ash.samples().$$;
+
+comment on function ash.decode_sample(integer[], smallint) is
+$$Decodes a single ash.sample.data array into (wait_event, query_id, count) rows. Pass p_slot (ash.sample.slot) to resolve query_ids unambiguously; omitting it searches all query_map partitions and may return a stale id after rotation.$$;
 
 -- Migration: add version column if upgrading from older version
 do $$


NikolayS:
# REV review: PR #36 — docs: improve ash.sample discoverability

Small docs-only PR (+34 / −0, 2 files). Adds `comment on table/column/function` statements to `sql/ash-install.sql` and a "Dump samples to CSV" section to `README.md`.

## CI status

All checks pass: test (PG 14–18), Analyze, CodeQL.

## BLOCKING

None.

## NON-BLOCKING

None.

## POTENTIAL

**LOW — INFO — wording of "negative marker" explanation in `ash.sample.data` comment**
- file: `sql/ash-install.sql:149`
- The sentence "A negative marker starts each group; wait_id is negated for the marker so a positive count cannot be mistaken for a group boundary" is accurate but slightly roundabout. What the code actually relies on (see `ash.decode_sample` at line 527 and `ash.samples` at line 2487, `p_data[v_idx] >= 0`) is: wait_ids are stored as negative so that non-negative `count` and `query_map_id` values are trivially distinguishable from group headers during a linear walk.
- suggestion: consider "Wait_ids are stored negated so the scan can distinguish group headers from the non-negative count and query_map_ids that follow." Non-blocking — the current wording is correct.
- confidence: 5

**LOW — INFO — cast style inconsistency in README example**
- file: `README.md:305`
- The new example uses `ash.samples('1 hour'::interval, 10000000)` while the existing example at line 293 uses an untyped literal `ash.samples('10 minutes', 20)`. Both work; the explicit cast isn't wrong, but mixing styles in adjacent sections is a minor consistency issue.
- suggestion: drop `::interval` to match the existing example, or update the earlier example for uniformity. Non-blocking.
- confidence: 5

**LOW — INFO — "silently redirecting stderr to /dev/null" advice lacks context**
- file: `README.md:311-313`
- The final paragraph warns against `2>/dev/null` hiding typos, but the examples above don't show any stderr redirection, so the warning reads as a non-sequitur to readers who haven't been bitten by this before.
- suggestion: either add a brief "don't do `... 2>/dev/null`" antipattern snippet, or drop the clause — as-is it's context-free caution.
- confidence: 4

## Verification performed

- Cross-checked `comment on function ash.samples(interval, int)` wording against the function definition at `sql/ash-install.sql:2443` — defaults `'1 hour'` and `100` match; return signature `(sample_time, database_name, active_backends, wait_event, query_id, query_text)` matches; pg_stat_statements dependency noted correctly (line 2472 vs. 2507 fallback).
- Cross-checked `ash.samples_at` comment against definition at line 2545 — default `p_limit = 100` matches; return shape is identical to `ash.samples()`.
- Cross-checked `ash.decode_sample` comment against definition at line 490 — `(wait_event, query_id, count)` matches; `p_slot` semantics match the warning comment at line 568 about cross-partition ambiguity after rotation.
- Cross-checked packed-layout claim (`(-wait_id, count, query_map_id_1, …, query_map_id_count)`) against decode walk at line 524-558 and the `ash.samples` expansion at line 2480-2489 — layout is accurate. count=0 is safely handled in both paths (`1..0` FOR loop is empty; `generate_series(0, -1)` is empty).
- README `\copy` example is client-side and uses `/tmp/ash.csv`, which is written by the psql client, not the server — safe on RDS / Cloud SQL / Supabase / AlloyDB / Neon, matching the anti-extension design noted in CLAUDE.md.
- No SQL style-guide violations: lowercase keywords, `snake_case` names, dollar-quoted string bodies for multi-line comments.
- Platform-neutrality rule: "managed services" is used as a neutral category noun, not as criticism — compliant.

## Summary

PASS. Docs-only change, CI green across PG 14–18, all documented claims verified against the actual function bodies and packed-array format. Three LOW/INFO wording suggestions above are optional polish; none block merge.

Issue #37, opened 2026-04-21
Comprehensive review: correctness, security, CI, and docs findings (v1.4)


NikolayS:
## Overview

End-to-end review of pg_ash covering `sql/ash-install.sql`, upgrade scripts, CI workflows, and functional behaviour on PostgreSQL 16 (PG 18 install blocked by PGDG 403 in the review sandbox — CI still covers 14–18).

Scope:
- Static review of `sql/ash-install.sql` (~2,926 lines) and `sql/ash-{1.0,1.0-to-1.1,1.1,1.1-to-1.2,1.2-to-1.3,1.3-to-1.4}.sql`
- Review of `.github/workflows/{test,claude,claude-code-review}.yml` and `.github/dependabot.yml`
- Functional smoke test: fresh install, full upgrade chain 1.0→1.1→1.2→1.3→1.4, schema-equivalence diff between fresh and upgraded, degraded mode without pg_stat_statements, interval validation, rotation cycle, uninstall+reinstall, advisory-lock DoS reproduction
- Security review (injection, privileges, search_path, DoS, supply chain)

Counts: 5 HIGH, 12 MEDIUM, 13 LOW, and 5 INFO findings. Highest-impact items: release-version inconsistency (1.3 vs 1.4), reader functions granted to PUBLIC exposing cross-tenant `query_id`s, `ash.start()` skips interval validation when pg_cron is absent, CI never exercises the documented 1.0→1.1 upgrade wrapper, `wait_event_map` has no overflow cap.

---

### H-REL-1 `ash-install.sql` header says "Version: 1.3" but the code ships 1.4
- `sql/ash-install.sql:2` — header: `-- Version: 1.3 (latest)`
- `sql/ash-install.sql:67` — `version text not null default '1.4'`
- `sql/ash-install.sql:4–6` — upgrade directions list only 1.0/1.1/1.2 paths, omitting `1.3 → 1.4`
- Confirmed at runtime: fresh install reports `version = 1.4` in `ash.status()`

### H-REL-2 README.md documents only up to 1.3
- `README.md:76–87` — "from 1.0 to 1.1…/1.2 to 1.3" but no mention of 1.3 → 1.4
- `README.md:147` — status() example output shows `version | 1.3`
- CLAUDE.md says `v1.3 → v1.4` upgrade scripts are generated at release time, yet 1.4 isn't mentioned in user docs

### H-REL-3 `RELEASE_NOTES.md` has no 1.4 entry
[...]

PR #39, opened 2026-04-22
feat(status): surface epoch-overflow horizon

NikolayS:
## Summary

- Add `epoch_seconds_remaining` metric to `ash.status()`, computed as `2147483647 - extract(epoch from (now() - ash.epoch()))`. Lets operators monitor how close `sample_ts` (int4 seconds since 2026-01-01 UTC) is to wrapping around 2094-01-19.
- Document the overflow horizon in a comment block next to `ash.epoch()` and note that a bigint column migration is the only safe fix before ~2090 (moving the epoch would corrupt every historical sample).
- Add a "Known limitations" bullet in `README.md` linking to issue #37.

## Test plan

- [ ] Fresh install of `sql/ash-install.sql` on PG 14-18 succeeds.
- [ ] `select * from ash.status();` returns a new `epoch_seconds_remaining` row with a positive bigint value (~2.14e9 shortly after 2026-01-01).
- [ ] Spot-check: value decreases by ~1 per wall-clock second and would reach 0 around 2094-01-19.
- [ ] Degraded-mode run (no pg_cron, no pgss) still reports the metric.

Closes the INFO item in #37 (sample_ts wraps at 2094). Schema migration to bigint remains a follow-up before ~2090.


NikolayS:
# REV Review — PR #39: feat(status): surface epoch-overflow horizon

**Scope:** 3 files, +25/-0. Adds `epoch_seconds_remaining` metric to `ash.status()`, expands the `ash.epoch()` comment block, and adds a README "Known limitations" bullet.

**CI:** all 7 checks pass (Analyze, CodeQL, test 14-18).

---

## BLOCKING

_None._

---

## NON-BLOCKING

**1. "Silent overflow" / "wraps" wording is inaccurate** — `sql/ash-install.sql` comment (lines near 46-54) and the README bullet both describe the horizon as a "wrap" / "silent overflow." In practice, `v_sample_ts := extract(epoch from now() - ash.epoch())::int4;` (install.sql line 283, and the identical cast in `ash-1.3-to-1.4.sql`) does not wrap in PostgreSQL — the `::int4` cast raises `ERROR: integer out of range`, aborting `take_sample()`. Post-2094, sampling will hard-fail rather than silently corrupting history, which is arguably safer but different from what the docs promise. Suggest tightening to "sampling will error with `integer out of range`" in both the README bullet and the in-file comment. Confidence: 9. Severity: low (documentation accuracy).

**2. `epoch_seconds_remaining` goes negative after 2094 without explanation** — the subtraction `2147483647::bigint - extract(...)::bigint` yields a negative value after the horizon. That is fine as a "how overdue" signal, but it is worth one line in the comment near the metric (install.sql ~line 1076, upgrade ~line 315) so an operator reading the code isn't surprised. Confidence: 7. Severity: trivial.

**3. Upgrade script touched in a feature PR** — `CLAUDE.md` states "Upgrade scripts (`ash-X.Y-to-X.Z.sql`) are generated at release time, not in feature PRs." This PR edits `sql/ash-1.3-to-1.4.sql`. In practice 1.4 has not been tagged and the file was introduced in PR #29, so editing it here is pragmatic — but it is worth either loosening the rule in `CLAUDE.md` or deferring this specific edit to release time. Confidence: 6. Severity: low (process / guideline).

---

## POTENTIAL

**4. Negative-elapsed edge case pre-epoch** — if the function is ever called with a system clock earlier than `2026-01-01 UTC` (severe clock skew, restore-from-backup test, etc.), `now() - ash.epoch()` is negative and `epoch_seconds_remaining` reports a value > 2,147,483,647. Not a real production concern in 2026+, but a `greatest(0, ...)` or clamp would make the metric monotonically interpretable. Confidence: 4. Severity: trivial.

**5. Epoch reference vs. `extract(epoch from interval)`** — the calculation uses `extract(epoch from (now() - ash.epoch()))`, i.e. `extract` on an interval, which returns total seconds including fractional. The cast to `bigint` truncates toward zero, so the reported value may lag the wall clock by up to 1 second. Acceptable for a horizon metric (~68-year range). Confidence: 8. Severity: none. (Noted only because the PR body claims "decreases by ~1 per wall-clock second" — correct.)

**6. Timezone handling** — `ash.epoch()` is declared `timestamptz` with explicit `+00`, `now()` is `timestamptz`, and subtraction of two timestamptz values yields an absolute interval independent of `TimeZone`. No session-`TimeZone` dependency in the calculation. Verified correct. Confidence: 9. Severity: none (confirmation, not a finding).

---

## Test coverage

No new test was added. The PR body lists a manual test plan only. Given this is a single derived metric with a trivial arithmetic definition and no branching, automated coverage is marginal — a one-liner in the existing fresh-install test asserting `epoch_seconds_remaining` appears in `ash.status()` output and parses as a positive integer would be cheap insurance against future typos. Non-blocking.

---

## Summary

- **Verdict:** approve with minor doc tweaks
- **BLOCKING:** 0
- **NON-BLOCKING:** 3 (wording "wrap/silent" vs. actual `int4 cast error`; comment for post-horizon negative values; upgrade-script edit vs. CLAUDE.md policy)
- **POTENTIAL:** 3 (clock-skew clamp; truncation lag; timezone confirmation)
- **Correctness of the 2094 horizon math:** confirmed. `int4` max is 2,147,483,647; 2,147,483,647 s ≈ 68.096 years; `2026-01-01 UTC + 68y 18d 3h 14m 7s` = `2094-01-19 03:14:07 UTC`, matching the README and `ash.epoch()` comment exactly.
- **Epoch reference:** correct. Uses `ash.epoch()` (the installed constant), not `installed_at`, so the horizon is stable across restarts and re-installs — which is the right choice given `sample_ts` is anchored to `ash.epoch()`.
- **Timezone handling:** correct. `timestamptz - timestamptz` is timezone-invariant.
- **Status output clarity:** metric name is self-explanatory; the README bullet links issue #37 and names the mitigation (`bigint` column migration). Clear enough to ship.

PR #40, opened 2026-04-22
security(ci): pin actions, gate claude.yml, tighten perms, document role req

NikolayS:
## Summary

Supply-chain hardening for #37, slice H-SEC-4 / M-SEC-5 / M-SEC-6 / M-SEC-8.

- **H-SEC-4** — Pinned every `uses:` in the three workflows to a full 40-char commit SHA with a `# vX.Y.Z` comment:
  - `actions/checkout` -> `de0fac2e4500dabe0009e67214ff5f5447ce83dd` (`v6.0.2`)
  - `anthropics/claude-code-action` -> `4e5d8b13ca281a6d163cdb287d8917b216e00d6f` (`v1.0.103`)
- **M-SEC-6** — Added `author_association` gating (OWNER/MEMBER/COLLABORATOR) to `claude.yml` and `claude-code-review.yml` so random GitHub users can't trigger jobs holding `CLAUDE_CODE_OAUTH_TOKEN`. Applied per event type because `author_association` sits on different payload paths for `issue_comment` / `pull_request_review_comment` / `pull_request_review` / `issues` / `pull_request`.
- **M-SEC-8** — Added a top-level `permissions: { contents: read }` block to `test.yml`.
- **M-SEC-5** — Documented the `pg_read_all_stats` requirement: new "Privileges" section in `README.md`, and `ash.start()` now emits a `RAISE NOTICE` warning when the current role is neither a superuser nor a member of `pg_read_all_stats` (so users are told up-front that `query_id` will collapse to sentinel `0` for other-user activity).

## Open question / TODO

`id-token: write` is **retained** on both Claude workflows with an inline `TODO` comment. The public claude-code-action docs don't confirm whether OIDC is consumed for the OAuth-token auth path we use, and dropping it blindly risks breaking the action. Please confirm against the action source / internal docs and drop the permission if it's unused — that's purely a one-line follow-up.

## Test plan

- [ ] `test.yml` still green across PG 14-18 (no write scopes are actually consumed).
- [ ] Trigger `@claude` as a non-collaborator on a throwaway issue and confirm the job is skipped (gate works).
- [ ] Trigger `@claude` as OWNER/MEMBER/COLLABORATOR and confirm the job still runs.
- [ ] Run `ash.start()` as a non-superuser role without `pg_read_all_stats` and verify the warning notice is emitted; re-run after `grant pg_read_all_stats` and verify the warning disappears.
- [ ] Confirm pinned SHAs resolve (`git ls-remote` on each upstream matches the commented tag).

Refs #37 (H-SEC-4, M-SEC-5, M-SEC-6, M-SEC-8).

### DIFF
--- a/.github/workflows/claude-code-review.yml
+++ b/.github/workflows/claude-code-review.yml
@@ -12,28 +12,29 @@ on:
 
 jobs:
   claude-review:
-    # Optional: Filter by PR author
-    # if: |
-    #   github.event.pull_request.user.login == 'external-contributor' ||
-    #   github.event.pull_request.user.login == 'new-developer' ||
-    #   github.event.pull_request.author_association == 'FIRST_TIME_CONTRIBUTOR'
+    # Only review PRs from trusted authors — prevents external contributors
+    # from causing runs that consume CLAUDE_CODE_OAUTH_TOKEN (#37 M-SEC-6).
+    if: |
+      github.event.pull_request.author_association == 'OWNER'
+      || github.event.pull_request.author_association == 'MEMBER'
+      || github.event.pull_request.author_association == 'COLLABORATOR'
 
     runs-on: ubuntu-latest
     permissions:
       contents: read
       pull-requests: read
       issues: read
-      id-token: write
+      id-token: write # TODO: confirm claude-code-action needs OIDC; drop if unused (#37 H-SEC-4)
 
     steps:
       - name: Checkout repository
-        uses: actions/checkout@v6
+        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
         with:
           fetch-depth: 1
 
       - name: Run Claude Code Review
         id: claude-review
-        uses: anthropics/claude-code-action@v1
+        uses: anthropics/claude-code-action@4e5d8b13ca281a6d163cdb287d8917b216e00d6f # v1.0.103
         with:
           claude_code_oauth_token: ${{ secrets.CLAUDE_CODE_OAUTH_TOKEN }}
           plugin_marketplaces: 'https://github.com/anthropics/claude-code.git'
--- a/.github/workflows/claude-code-review.yml
+++ b/.github/workflows/claude-code-review.yml
@@ -24,7 +24,6 @@ jobs:
       contents: read
       pull-requests: read
       issues: read
-      id-token: write # TODO: confirm claude-code-action needs OIDC; drop if unused (#37 H-SEC-4)
 
     steps:
       - name: Checkout repository
--- a/.github/workflows/claude.yml
+++ b/.github/workflows/claude.yml
@@ -12,27 +12,45 @@ on:
 
 jobs:
   claude:
+    # Gate on both @claude mention AND trusted author_association to prevent
+    # arbitrary GitHub users from triggering runs holding CLAUDE_CODE_OAUTH_TOKEN.
+    # author_association lives on different payload paths for each event type.
     if: |
-      (github.event_name == 'issue_comment' && contains(github.event.comment.body, '@claude')) ||
-      (github.event_name == 'pull_request_review_comment' && contains(github.event.comment.body, '@claude')) ||
-      (github.event_name == 'pull_request_review' && contains(github.event.review.body, '@claude')) ||
-      (github.event_name == 'issues' && (contains(github.event.issue.body, '@claude') || contains(github.event.issue.title, '@claude')))
+      (
+        (github.event_name == 'issue_comment' && contains(github.event.comment.body, '@claude')
+          && (github.event.comment.author_association == 'OWNER'
+              || github.event.comment.author_association == 'MEMBER'
+              || github.event.comment.author_association == 'COLLABORATOR'))
+        || (github.event_name == 'pull_request_review_comment' && contains(github.event.comment.body, '@claude')
+          && (github.event.comment.author_association == 'OWNER'
+              || github.event.comment.author_association == 'MEMBER'
+              || github.event.comment.author_association == 'COLLABORATOR'))
+        || (github.event_name == 'pull_request_review' && contains(github.event.review.body, '@claude')
+          && (github.event.review.author_association == 'OWNER'
+              || github.event.review.author_association == 'MEMBER'
+              || github.event.review.author_association == 'COLLABORATOR'))
+        || (github.event_name == 'issues'
+          && (contains(github.event.issue.body, '@claude') || contains(github.event.issue.title, '@claude'))
+          && (github.event.issue.author_association == 'OWNER'
+              || github.event.issue.author_association == 'MEMBER'
+              || github.event.issue.author_association == 'COLLABORATOR'))
+      )
     runs-on: ubuntu-latest
     permissions:
       contents: read
       pull-requests: read
       issues: read
-      id-token: write
+      id-token: write # TODO: confirm claude-code-action needs OIDC; drop if unused (#37 H-SEC-4)
       actions: read # Required for Claude to read CI results on PRs
     steps:
       - name: Checkout repository
-        uses: actions/checkout@v6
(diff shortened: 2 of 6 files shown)


NikolayS:
# REV review — PR #40

**Verdict:** LGTM with minor follow-ups. No BLOCKING findings. The supply-chain hardening is well-executed: all third-party actions are pinned to full 40-char commit SHAs with tag comments, workflow-level permissions default to least-privilege, and the `@claude` invocation path is now gated on `author_association`. CI is green across PG 14–18.

**Per-area counts:**
- Security: 0 BLOCKING, 2 POTENTIAL, 0 NON-BLOCKING
- Bug-hunter: 0 BLOCKING, 1 POTENTIAL
- Guidelines: 0 BLOCKING, 2 INFO
- Docs: 0 BLOCKING, 1 INFO
- Tests: 0 BLOCKING, 1 POTENTIAL

---

## Security review

### [POTENTIAL] `id-token: write` retained on both Claude workflows (confidence 6)
- **file:** `.github/workflows/claude.yml:28`, `.github/workflows/claude-code-review.yml:24`
- **issue:** The PR explicitly calls this out as a TODO. `id-token: write` grants the job a signed OIDC JWT which is primarily consumed by cloud-provider federated identity flows. The `anthropics/claude-code-action@v1.0.103` path used here authenticates via `CLAUDE_CODE_OAUTH_TOKEN` (a secret), not OIDC, so `id-token: write` is almost certainly unnecessary. Leaving it set is not itself an exploit, but it widens the attack surface — a compromised action step (or a future version of the action) could mint an OIDC token scoped to this repo.
- **fix:** Drop `id-token: write` from both workflows and verify the next `@claude` invocation still succeeds. If it regresses, re-add with a justification comment.
- **cwe:** CWE-272 (Least Privilege Violation)

### [POTENTIAL] `author_association` is a payload-time snapshot, not a live membership check (confidence 5)
- **file:** `.github/workflows/claude.yml:16-35`, `.github/workflows/claude-code-review.yml:18-22`
- **issue:** `author_association` on the event payload is the actor's association at the moment the comment/PR was created. It is not a real-time check — if a collaborator is later removed from the repo, any of their older `@claude` comments could still be replayed if the event were redelivered (e.g., via `gh api /repos/.../hooks/.../deliveries/:id/attempts`). For a public repo this is the documented GitHub Actions gate and accepted practice; worth noting as a known limitation rather than a deficit.
- **fix:** Accept as-is. If stricter guarantees are desired later, supplement with a job-level script step that queries `/repos/:owner/:repo/collaborators/:username/permission` using the ephemeral `GITHUB_TOKEN` before any secret-consuming step. Out of scope for this PR.
- **cwe:** CWE-287 (informational)

### Security positives (verified, no finding)
- `actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd` — 40-char SHA, matches `v6.0.2`.
- `anthropics/claude-code-action@4e5d8b13ca281a6d163cdb287d8917b216e00d6f` — 40-char SHA, matches `v1.0.103`.
- `pull_request` (not `pull_request_target`) is used in `claude-code-review.yml` — correct choice for reviewing untrusted PR diffs; runs with read-only `GITHUB_TOKEN` and no access to secrets from forks.
- `test.yml` now declares top-level `permissions: { contents: read }` — good default.
- `claude-code-review.yml` scopes `contents: read`, `pull-requests: read`, `issues: read` — least-privilege matches what the action needs.
- `CLAUDE_CODE_OAUTH_TOKEN` is referenced only via `${{ secrets.* }}` and never echoed into logs.
- `author_association` is correctly accessed on the event-specific payload path for each of `issue_comment` / `pull_request_review_comment` / `pull_request_review` / `issues` — not conflated across events.
- Allowlist is restricted to `OWNER` / `MEMBER` / `COLLABORATOR`; `CONTRIBUTOR` and `FIRST_TIME_CONTRIBUTOR` are correctly excluded (those do not imply write access).

---

## Bug-hunter

### [POTENTIAL] `exception when others then null` silently swallows probe failures (confidence 6)
- **file:** `sql/ash-install.sql:710-725`, `sql/ash-1.3-to-1.4.sql:341-356`
- **issue:** The privilege probe wraps the check in `exception when others then null;`. If the probe ever fails (e.g., a future PG release reorders `pg_has_role` overloads, or `pg_roles` access is denied), the user gets no warning, no log entry, and no indication that the check was skipped — the same state as a role that does have the privilege. For a warning system intended to flag a silent-data-skew condition, silently skipping the warning is precisely the wrong failure mode.
- **fix:** Surface the skip:
  ```sql
  exception when others then
    raise notice 'ash.start(): privilege check skipped (%)', sqlerrm;
  ```

---

## Guidelines checker

### [INFO] Commit type `security(ci):` not in Conventional Commits standard set (confidence 6)
- **issue:** Recent commits use `feat`, `fix`, `docs`, `chore`, `ci` prefixes. `security` is not one of the standard Conventional Commits types. More idiomatic would be `ci(security): ...` or `chore(security): ...`.
- **rule:** `development__git-commit-standards.mdc`
- **suggestion:** Non-blocking. Informational for future commits.

### [INFO] Upgrade script modified in a feature PR (confidence 5)
- **file:** `sql/ash-1.3-to-1.4.sql`
- **issue:** CLAUDE.md states: *"Upgrade scripts (ash-X.Y-to-X.Z.sql) are generated at release time, not in feature PRs."* This PR appends a new `create or replace function ash.start()` block to the existing upgrade script. The script existed pre-PR, so this is an edit to an already-present file rather than a new file generated at release time; still, the policy intent is ambiguous here — worth a quick confirmation with the project owner that in-flight edits to the pending upgrade script are expected.
- **suggestion:** Confirm the current release workflow expects in-flight edits vs. regeneration at tag time.

---

## Docs reviewer

### [INFO] README "Privileges" block does not mention the `RAISE NOTICE` warning behavior (confidence 4)
- **file:** `README.md:73-93`
- **issue:** The new section explains the need for `pg_read_all_stats` but does not tell the reader that `ash.start()` will now emit a `RAISE NOTICE` warning if the requirement is unmet. Mentioning this closes the loop between the README section and the new runtime warning, so operators are not surprised when they see the notice in psql output.
- **suggestion:** Add one sentence: *"`ash.start()` emits a `RAISE NOTICE` warning if the current role is neither a superuser nor a member of `pg_read_all_stats`."*

---

## Test analyzer

### [POTENTIAL] No CI coverage for the new privilege warning path (confidence 5)
- **file:** `.github/workflows/test.yml`, `sql/ash-install.sql:710-725`
- **issue:** The privilege-check branch of `ash.start()` is new logic but is not exercised by CI. The PR's own test plan lists this as a manual verification step. A lightweight, deterministic test would be:
  1. create a non-superuser role without `pg_read_all_stats`;
  2. run `ash.start()` as that role;
  3. assert that the notice message contains `pg_read_all_stats` (capturable via psql `\set VERBOSITY verbose` or a `do $$ ... $$` wrapper).
- **untested_scenarios:** (a) role lacks privilege → warning emitted; (b) role has `pg_read_all_stats` → no warning; (c) superuser → no warning; (d) the `exception when others` fallback path.

---

## CI status

All checks green on the PR tip:
- `test (14)` … `test (18)`: pass
- `CodeQL`: pass
- `Analyze (actions)`: pass (48s)

---

## Summary

- 0 BLOCKING, 4 POTENTIAL, 3 INFO.
- Top issue: decide on `id-token: write` and drop it if the OAuth-token auth path does not consume OIDC — this is the single remaining least-privilege tightening the PR itself flagged.
- Secondary: replace the bare `exception when others then null;` in the privilege probe with a `raise notice` so a silent probe failure is still visible, and add a CI test asserting the warning fires for an unprivileged role.
- The PR correctly uses `pull_request` (not `pull_request_target`), pins full 40-char SHAs with tag comments, gates `@claude` on `author_association` with per-event payload paths, and adds least-privilege `contents: read` at the workflow level. This is a solid supply-chain hardening pass.

PR #41, opened 2026-04-22
fix: silent data loss & start() consistency (H-BUG-1/2, M-BUG-4/6/8/9)

NikolayS:
## Summary

Correctness slice of the issue #37 fix bundle — eliminates silent data loss and behavioural inconsistencies across `ash.take_sample()`, `ash.decode_sample()`, and `ash.start()`. All changes are in `sql/ash-install.sql`.

- **H-BUG-1** `ash.start()` now validates interval shape (seconds / exact-minutes / exact-hours / 23h ceiling) *before* the pg_cron-availability branch, so degraded mode and pg_cron mode reject the same inputs.
- **H-BUG-2** When `ash_sampler` already exists, `ash.start(new_interval)` now calls `cron.alter_job(schedule := ...)` so the cron cadence actually tracks the requested interval (previously only `ash.config.sample_interval` was updated).
- **M-BUG-4** `take_sample()`'s inner `exception when others` no longer silently drops rows — added `ash.config.insert_errors bigint` (idempotent `alter table ... add column if not exists`), incremented on every swallowed error, surfaced as a new row in `ash.status()`.
- **M-BUG-6 / H-SEC-3** `wait_event_map.id` widened `smallint -> int`, and `_register_wait` now enforces a 50 000-row cap mirroring the `query_map` pattern (RAISE NOTICE + sentinel id on cap). All `(-s.data[i])::smallint` reader casts widened to `::int`; `ash.status()`'s `wait_event_map_utilization` denominator updated 32767 → 50000.
- **M-BUG-8** pg_cron version parse wrapped in a `'^\d+\.\d+'` guard plus `begin/exception`; unparseable/empty `extversion` no longer crashes `ash.start()`, it degrades to "assume modern (>= 1.5)" with a NOTICE.
- **M-BUG-9** `decode_sample()` now validates the whole array shape in a first pass and only emits rows in a second pass — malformed input yields a single warning and zero rows instead of 1+ partial rows followed by a warning.

## Test plan

- [x] Fresh install succeeds on PG 16 (`\d ash.wait_event_map` confirms `id integer`).
- [x] `ash.take_sample()` returns 0, `ash.status()` shows `insert_errors = 0` and `wait_event_map_utilization` denominator = 50 000.
- [x] `ash.start('1 second')` in no-pg_cron mode records interval and prints external-scheduler advice.
- [x] `ash.start('90 seconds')` in no-pg_cron mode now rejects with `interval must be exact minutes …` (H-BUG-1 regression check).
- [x] `ash.decode_sample(array[-1,3,100,200])` emits 0 rows + 1 warning (M-BUG-9 regression check).
- [x] `ash.decode_sample(array[-1,2,0,0])` against a seeded wait event still emits the expected 2 rows.
- [ ] CI matrix (PG 14/15/16/17/18): fresh install, upgrade path 1.0→1.3, schema equivalence, degraded mode.
- [ ] Exercise cap behaviour in a dedicated test (requires seeding 50 000 wait events; out of scope for this PR).
- [ ] Reviewer: confirm `cron.alter_job(job_id := ..., schedule := ...)` signature matches the pg_cron versions targeted by the project (1.4+).

Refs #37

### DIFF
--- a/sql/ash-1.3-to-1.4.sql
+++ b/sql/ash-1.3-to-1.4.sql
@@ -356,6 +356,126 @@ exception when query_canceled then
 end;
 $$;
 
+-- Re-create decode_sample with two-pass validation (M-BUG-9): previously a
+-- malformed trailing segment produced partial rows followed by a WARNING;
+-- now validation runs first and emits zero rows on failure.
+create or replace function ash.decode_sample(p_data integer[], p_slot smallint default null)
+returns table (
+  wait_event text,
+  query_id int8,
+  count int
+)
+language plpgsql
+stable
+as $$
+declare
+  v_len int;
+  v_idx int;
+  v_wait_id int;
+  v_count int;
+  v_qid_idx int;
+  v_map_id int4;
+  v_type text;
+  v_event text;
+  v_query_id int8;
+begin
+  -- Basic validation
+  if p_data is null or array_length(p_data, 1) is null then
+    return;
+  end if;
+
+  v_len := array_length(p_data, 1);
+
+  -- Basic structure check: first element must be negative (wait_id marker)
+  if v_len < 3 or p_data[1] >= 0 then
+    raise warning 'ash.decode_sample: invalid data array';
+    return;
+  end if;
+
+  -- ---- Pass 1: validate shape only, emit nothing ----
+  -- Reuses the same walker logic the old code had, but exits with a single
+  -- warning and RETURN (no partial rows) if anything is wrong.
+  v_idx := 1;
+  while v_idx <= v_len loop
+    if p_data[v_idx] >= 0 then
+      raise warning 'ash.decode_sample: expected negative wait_id at position %', v_idx;
+      return;
+    end if;
+    v_idx := v_idx + 1;
+
+    if v_idx > v_len then
+      raise warning 'ash.decode_sample: unexpected end of array at position % (missing count)', v_idx;
+      return;
+    end if;
+    v_count := p_data[v_idx];
+    if v_count <= 0 then
+      raise warning 'ash.decode_sample: non-positive count % at position %', v_count, v_idx;
+      return;
+    end if;
+    v_idx := v_idx + 1;
+
+    if v_idx + v_count - 1 > v_len then
+      raise warning 'ash.decode_sample: not enough query_ids for count % at position %', v_count, v_idx;
+      return;
+    end if;
+    for v_qid_idx in 1..v_count loop
+      if p_data[v_idx] < 0 then
+        raise warning 'ash.decode_sample: expected non-negative query_id at position %', v_idx;
+        return;
+      end if;
+      v_idx := v_idx + 1;
+    end loop;
+  end loop;
+
+  -- ---- Pass 2: emit rows (shape is known good) ----
+  v_idx := 1;
+  while v_idx <= v_len loop
+    v_wait_id := -p_data[v_idx];
+    v_idx := v_idx + 1;
+
+    v_count := p_data[v_idx];
+    v_idx := v_idx + 1;
+
+    -- Look up wait event info
+    select w.type, w.event
+    into v_type, v_event
+    from ash.wait_event_map w
+    where w.id = v_wait_id;
+
+    -- Process each query_id
+    for v_qid_idx in 1..v_count loop
+      v_map_id := p_data[v_idx];
+      v_idx := v_idx + 1;
+
+      -- Handle sentinel (0 = NULL query_id)
+      if v_map_id = 0 then
+        v_query_id := null;
+      elsif p_slot is not null then
+        select m.query_id into v_query_id
+        from ash.query_map_all m
+        where m.slot = p_slot and m.id = v_map_id;
+      else
+        -- No slot context — search all partitions (less efficient).
+        -- WARNING: after rotation, the same id may exist in multiple
+        -- partitions with different query_ids (independent sequences).
+        -- Result is nondeterministic. Always pass p_slot when available.
+        select m.query_id into v_query_id
+        from ash.query_map_all m
+        where m.id = v_map_id
+        limit 1;
+      end if;
+
+      wait_event := case when v_event = v_type then v_event else v_type || ':' || v_event end;
+      query_id := v_query_id;
+      count := 1;
+      return next;
+    end loop;
+  end loop;
+
+  return;
+end;
+$$;
+
 -- Re-create start() with pre-branch validation (H-BUG-1), schedule re-sync
 -- (H-BUG-2), and defensive extversion parsing (M-BUG-8).
 create or replace function ash.start(p_interval interval default '1 second')
--- a/sql/ash-1.3-to-1.4.sql
+++ b/sql/ash-1.3-to-1.4.sql
@@ -5,6 +5,25 @@
 --     silently dropping the sample. Returns -1 on timeout. (#28)
 --   - New missed_samples counter in ash.config — incremented on every
 --     caught timeout, visible in ash.status(). (#27, #28)
+--   - take_sample() inner exception handler now bumps an insert_errors
+--     counter in ash.config instead of silently dropping data with just a
+--     WARNING. (M-BUG-4)
+--   - _register_wait() enforces a 32 000-row cap on wait_event_map to prevent
+--     DoS via unbounded dictionary growth; bumps register_wait_cap_hits in
+--     ash.config when the cap is hit. (M-BUG-6 / H-SEC-3)
+--   - status() surfaces the two new counters.
+--   - start() validates the interval shape before branching on pg_cron
+--     availability so the accept/reject outcome is independent of pg_cron.
+--     (H-BUG-1)
+--   - start() re-syncs cron.job.schedule on re-invocation when the sampler
+--     job already exists. (H-BUG-2)
+--   - start() defends against malformed pg_cron extversion. (M-BUG-8)
+--
+-- NOTE on CLAUDE.md: "Upgrade scripts are generated at release time, not in
+-- feature PRs." This script ships in a feature PR because test.yml runs the
+-- full upgrade chain on every PR and enforces schema equivalence against
+-- ash-install.sql — so the upgrade script must track install.sql changes to
+-- keep CI green. Revisit the policy once there's a dedicated release process.
 
 -- Add missed_samples column to config if missing
 do $$
@@ -18,11 +37,85 @@ begin
   end if;
 end $$;
 
+-- M-BUG-4: track rows silently dropped by take_sample()'s inner exception handler.
+alter table ash.config add column if not exists insert_errors bigint not null default 0;
+
+-- M-BUG-6 / H-SEC-3: track how often _register_wait hits the dictionary cap
+-- and has to skip a new (state,type,event). Non-zero means wait-event
+-- registrations are being silently dropped for that sample (those sessions
+-- won't appear in encoded data for this tick). Surfaced by ash.status().
+alter table ash.config add column if not exists register_wait_cap_hits bigint not null default 0;
+
 -- Update version (both data and column default for schema parity)
 update ash.config set version = '1.4' where singleton;
 alter table ash.config alter column version set default '1.4';
 
--- Re-create take_sample with query_canceled handler
+-- Re-create _register_wait with dictionary size cap (M-BUG-6 / H-SEC-3)
+create or replace function ash._register_wait(p_state text, p_type text, p_event text)
+returns smallint
+language plpgsql
+as $$
+declare
+  v_id    smallint;
+  v_count bigint;
+begin
+  -- Try to get existing
+  select id into v_id
+  from ash.wait_event_map
+  where state = p_state and type = p_type and event = p_event;
+
+  if v_id is not null then
+    return v_id;
+  end if;
+
+  -- Enforce dictionary size cap before inserting. 32 000 stays well below
+  -- smallint's 32 767 ceiling while still leaving room for genuine event
+  -- diversity (real wait-event inventories measure in the hundreds).
+  -- reltuples is a cheap estimate; an exact count would add a scan on
+  -- every fresh wait-event observation on the sampler's hot path.
+  select reltuples::bigint into v_count
+  from pg_class where oid = 'ash.wait_event_map'::regclass;
+
+  if v_count >= 32000 then
+    -- Bump the cap-hit counter so ash.status() can surface the drop.
+    -- Note: counts *registration drops* here — not the number of sampled
+    -- backends observed for this (state,type,event). Many concurrent
+    -- backends blocked on the same dropped event only bump it once per tick.
+    -- Wrap in an inner block: if the UPDATE itself fails (e.g. config row
+    -- missing mid-uninstall), we still want the outer WARNING to fire and
+    -- the function to return NULL without aborting take_sample().
+    begin
+      update ash.config set register_wait_cap_hits = register_wait_cap_hits + 1
+        where singleton;
+    exception when others then
+      null;  -- counter bump is best-effort
+    end;
+    raise warning 'ash._register_wait: wait_event_map at cap (>= 32 000 rows); skipping (state=%, type=%, event=%) — see ash.status()',
+      p_state, p_type, p_event;
+    return null;  -- caller PERFORMs this; snapshot JOIN drops the session
+  end if;
+
+  -- Insert new entry
+  insert into ash.wait_event_map (state, type, event)
+  values (p_state, p_type, p_event)
+  on conflict (state, type, event) do nothing
+  returning id into v_id;
+
+  -- If insert succeeded, return it
+  if v_id is not null then
+    return v_id;
+  end if;
+
+  -- Race condition: another session inserted, fetch it
+  select id into v_id
+  from ash.wait_event_map
+  where state = p_state and type = p_type and event = p_event;
+
+  return v_id;
+end;
+$$;
+
+-- Re-create take_sample with query_canceled handler and insert_errors counter
 create or replace function ash.take_sample()
 returns int
 language plpgsql
@@ -221,6 +314,23 @@ begin
       end if;
 
     exception when others then
+      -- M-BUG-4: previously a CHECK violation on ash.sample.data (or any
+      -- other INSERT-time error) was silently swallowed with just a WARNING,
+      -- dropping a row of observability data without any durable signal.
+      -- Bump insert_errors so ash.status() can surface the count, and keep
+      -- the warning so live log watchers still see it.
+      --
+      -- Nested BEGIN/EXCEPTION around the counter UPDATE: the outer
+      -- `exception when others` is a terminal handler, but the UPDATE
+      -- itself can fail (e.g. config row absent mid-uninstall, lock
+      -- timeout) and propagate out of this block. Before widening this
+      -- handler to do bookkeeping, no propagation path existed — preserve
+      -- that property so a flaky UPDATE never aborts the whole sampler.
+      begin
+        update ash.config set insert_errors = insert_errors + 1 where singleton;
+      exception when others then
+        null;  -- counter bump is best-effort; don't let it abort take_sample()
+      end;
       raise warning 'ash.take_sample: error inserting sample for datid % [%]: %', v_datid_rec.datid, sqlstate, sqlerrm;
     end;
   end loop;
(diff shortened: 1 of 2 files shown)


NikolayS:
# REV Review — PR #41: `fix: silent data loss & start() consistency`

Scope: `sql/ash-install.sql` only (+179/-83). Six bug fixes from issue #37: H-BUG-1, H-BUG-2, M-BUG-4, M-BUG-6/H-SEC-3, M-BUG-8, M-BUG-9. Also exercised against the repo's CI matrix (PG 14/15/16/17/18) and the test.yml upgrade/idempotency scenarios.

## Verdict

**REQUEST CHANGES — 1 BLOCKING regression, CI is red on all 5 PG versions.**

The correctness fixes themselves are well-targeted and most are clean improvements, but widening `ash.wait_event_map.id` / `_register_wait` return type to `int` was not coordinated with the legacy upgrade scripts, and as a direct consequence the "Dev workflow: re-apply migrations (idempotent)" CI step now crashes on every PG version the project supports. There is also one silent-correctness concern inside the new cap branch of `_register_wait`.

## BLOCKING

### B1. `create or replace function ash._register_wait ... returns int` conflicts with legacy upgrade scripts — CI red on PG 14/15/16/17/18 (confidence 10)

- PR widens `ash.wait_event_map.id` from `smallint` to `int` and `ash._register_wait` return type from `smallint` to `int` in `sql/ash-install.sql` (lines 79 / 152-153 on the PR branch).
- `sql/ash-1.2-to-1.3.sql:51-85` still contains `create or replace function ash._register_wait(...) returns smallint ...` (unchanged by this PR, and correct for that version's table shape).
- The test.yml step "Dev workflow: re-apply migrations (idempotent)" (`.github/workflows/test.yml:1406-1450`) runs `\i sql/ash-install.sql` (which installs the **int** variant), then re-applies `ash-1.1-to-1.2.sql`, `ash-1.2-to-1.3.sql`, `ash-1.3-to-1.4.sql` on top. The second script now hits:
  ```
  psql:sql/ash-1.2-to-1.3.sql:85: ERROR:  cannot change return type of existing function
  HINT:  Use DROP FUNCTION ash._register_wait(text,text,text) first.
  ```
- Observed in all five matrix jobs of run 24765160493. `Process completed with exit code 3.` Same error will hit any real user who re-applies the legacy upgrade wrapper on top of a 1.4 install (e.g., the "dev re-apply" flow documented in CLAUDE.md).
- Fix options:
  1. Add an explicit `drop function if exists ash._register_wait(text,text,text);` to the top-of-file DO block in `ash-install.sql` (the one that already sweeps renamed/changed signatures, lines 13-37), **and** do the same in the forthcoming `ash-1.3-to-1.4.sql` so the upgrade wrapper plays the same role.
  2. Or leave `_register_wait` returning `smallint` in `ash-install.sql`, keep the `wait_event_map.id` column as `int` (this works — PG implicitly down-casts the return value while the id can still grow beyond 32767 internally until a wider function is shipped), and plan the signature widening for a dedicated migration in 1.5.

This is a correctness-for-fresh-install PR that **regresses the upgrade/re-apply path**, so it cannot merge as-is.

## POTENTIAL (4-7 confidence)

### P1. `_register_wait` cap sentinel returns id=1 even if id=1 was never inserted (confidence 6)

`sql/ash-install.sql:178-182` on PR branch:
```sql
-- Return a sentinel pointing at id=1 ...
select min(id) into v_id from ash.wait_event_map;
return coalesce(v_id, 1);
```
In practice the cap branch only fires when the table has >= 50 000 rows, so `min(id)` is never null and `coalesce(..., 1)` is dead code — but a re-entering reader later calls `ash.decode_sample` on a sample whose data array contains `-1` (sentinel), looks up `wait_event_map.id = 1`, and gets back **whichever wait event happened to be registered first** (often `active|CPU*|CPU*`). That's silently-wrong data: backends whose real wait event was past the cap will all be decoded as "CPU*" or whatever id=1 points to. The original bug description (#37 M-BUG-6) warned about silent data loss — this fix converts "insert error → swallowed" into "insert skipped → mis-attributed event", which is arguably worse because `insert_errors` does not count it and `status()` gives no user-visible signal beyond `wait_event_map_utilization` being at 100%.

Mitigation suggestions (any of):
- Seed a dedicated `(state='UNKNOWN', type='Overflow', event='Overflow')` row at install/upgrade time, return its id on cap.
- Bump a new counter `wait_event_cap_hits` in `ash.config` and surface it in `status()`.
- Return NULL and let the take_sample caller skip (but then upstream needs to tolerate NULL — larger change).

### P2. `update ash.config set insert_errors = insert_errors + 1` inside an `exception when others` handler is itself unhandled (confidence 5)

`sql/ash-install.sql:488-492` on PR branch. If the UPDATE raises (e.g., config row temporarily missing during uninstall/reinstall, or relation-level lock contention escalating to `canceling statement due to lock_timeout`), the exception propagates out of the inner block. Inside `take_sample`'s outer `exception when query_canceled`, only `query_canceled` is caught — any other exception now aborts the whole sampler. Before this PR the inner `exception when others` was the terminal handler, so no propagation path existed. Low-probability in steady state, but a new failure mode for an observability tool that previously had none.

Suggested fix: wrap the bookkeeping UPDATE in its own `begin / exception when others then null; end;` block, or fold the counter into the outer `query_canceled` path only and rely on the WARNING for the inner case.

### P3. `cron.alter_job` on the "already exists" path runs before the pg_cron version check (confidence 5)

`sql/ash-install.sql:895`. The block above validates the new interval shape and *then* checks `v_cron_version ~ '^\d+\.\d+'`. But `cron.alter_job(job_id, schedule => ...)` was introduced in pg_cron 1.4 (not 1.5 — so it matches the project target). If a site still runs pg_cron 1.3, the pre-existing call at line 2990 already assumes 1.4, so parity is preserved. Worth documenting this minimum in the `ash.start()` error path for version < 1.4 (currently the code only errors on < 1.5, and only for sub-minute intervals).

### P4. Two-pass validation in `decode_sample` walks the array twice on the hot path (confidence 4)

The fix is correct and matches the PR description, but decode_sample is called by every reader (`top_queries`, `query_waits`, `samples`…) on every sample row. For well-formed inputs — the common case — we now walk `p_data` twice end-to-end. For large arrays (hundreds of events × dozens of query ids each) this doubles the per-row CPU in the decoder. A one-pass fix that buffers rows in a `record[]` array and `return query select * from unnest()` at the end would avoid re-walking while still preserving the "zero rows on bad input" contract. Not a merge blocker, but worth measuring before 1.5 ships.

## NON-BLOCKING (8-10 confidence, not merge-stopping)

### N1. H-BUG-1, H-BUG-2, M-BUG-8, M-BUG-9 each fix the referenced defect correctly (confidence 9)

- **H-BUG-1**: interval validation is now structurally before the pg_cron-availability branch; same input produces the same accept/reject outcome either way. Verified against the ash.start('90 seconds') / '24 hours' / '1 month' reproductions in issue #37.
- **H-BUG-2**: `cron.alter_job(job_id := ..., schedule := v_schedule)` is called on the already-exists path; matches the signature already used at line 2990 and the documented pg_cron 1.4+ ABI.
- **M-BUG-8**: guards `extversion` parse with a `^\d+\.\d+` pattern match and a `begin/exception` around the array cast. Confirmed with manually-forced `extversion = 'dev'` — now NOTICEs and proceeds.
- **M-BUG-9**: validate-then-emit structure is correct; the `decode_sample(array[-1,3,1])` reproduction in issue #37 no longer emits the dangling row.

### N2. `alter table ash.config add column if not exists insert_errors` is idempotent and fresh-install-safe (confidence 8)

Matches the pattern `missed_samples` used in `ash-1.3-to-1.4.sql`. No concerns.

### N3. `M-BUG-4` insert_errors counter is exposed in `status()` (confidence 8)

Good — turns a silent data-loss bug into a visible metric. Combined with P2 above, worth hardening the counter write before release.

### N4. Widened `(-s.data[i])::smallint -> ::int` casts across all readers are consistent (confidence 9)

Twelve call sites, all updated. Matches the `wait_event_map.id int` widening.

## Bugs in scope of issue #37 not addressed here

Per PR body this is a correctness slice only, but reviewers of the follow-up PRs should track:
- H-BUG-3 (reader interval > 2× rotation_period returns empty silently).
- M-BUG-5 (`query_map_*` reltuples cap), M-BUG-7 (quadratic `query_waits`), L-BUG-10 (`_validate_data` dead code).
- H-SEC-1 / H-SEC-2 (reader functions PUBLIC, no `search_path` on any function).
- H-REL-1/2/3 (install header still says "Version: 1.3", RELEASE_NOTES has no 1.4 entry).

## Per-area counts

| Area | BLOCKING | POTENTIAL | NON-BLOCKING |
|---|---|---|---|
| bug-hunter (correctness) | 1 | 2 | 4 |
| security-reviewer | 0 | 0 | 0 (out of scope) |
| test-analyzer / CI | — see B1 — | 0 | 0 |
| guidelines-checker | 0 | 0 | 0 |
| docs-reviewer | 0 | 0 | 0 |

## CI

All 5 PG-matrix jobs **FAIL** on run 24765160493 with the same root cause as B1:
- test (14/15/16/17/18) — `psql:sql/ash-1.2-to-1.3.sql:85: ERROR: cannot change return type of existing function`.
- Analyze (actions) pass, CodeQL pass.

## Top actions for author

1. Resolve B1 so CI is green on all 5 PG versions.
2. Decide whether to seed a cap-overflow sentinel wait event (P1) or accept "first-registered event" as the overflow marker (and surface it in `status()`).
3. Wrap the insert_errors UPDATE inside its own `begin ... exception when others then null; end` block (P2).

NikolayS:
# REV Follow-up — PR #41 after `f422996`

**Verdict: STILL BLOCKING** — prior B1 (return-type widening) is genuinely fixed, but the same commit introduces a new schema-equivalence CI regression on all 5 PG versions. Net result: CI is still red for all 5 PG matrix jobs on run 24869319187.

## What was verified (resolved items)

- `_register_wait` return type reverted to `smallint` (sql/ash-install.sql:176); `wait_event_map.id` stays `smallint` (line 94). `create or replace function` on top of `ash-1.2-to-1.3.sql` no longer raises "cannot change return type." Prior B1 is resolved on the signature axis.
- All 20+ reader casts are `::smallint` (confirmed via grep over all `(-s.data[i])::` and `wm.id = ...::` call sites, lines 1334 … 2964).
- Cap lowered 50000 -> 32000 (line 200); safely inside smallint's 32767 ceiling. `status()` denominator updated to 32000 at sql/ash-install.sql:1195.
- Identity starts at 1 for both `wait_event_map` and all 3 `query_map_N`; id=0 is not accidentally emitted, so the packed-layout sentinel (`coalesce(m.id, 0) as map_id` for NULL query_id at line 467) does not collide.
- Cap branch now returns NULL and is consumed via `perform ash._register_wait(...)` at line 395 — caller ignores the return value. The per-datid encoding JOIN at line 469 is an INNER JOIN on `wait_event_map`, so a session whose (state,type,event) was skipped at registration time is dropped from the sample for that tick. This matches the author's claim ("drop-data, not crash-sampler") — confirmed, not a crash path.
- Nested BEGIN/EXCEPTION around the counter UPDATEs at lines 205-210 (`register_wait_cap_hits`) and 522-526 (`insert_errors`) is correct; resolves P2 as a propagation hazard. Both swallow errors silently (no counter for "failed to bump"), which is acceptable for best-effort bookkeeping but leaves zero observability if the config row is persistently unreachable. Low severity.
- `alter table ash.config add column if not exists` for both `insert_errors` (line 79) and `register_wait_cap_hits` (line 85) is idempotent and compatible with the existing `missed_samples` DO-block migration pattern (sql/ash-install.sql:2786-2795).

## NEW BLOCKING (introduced by the follow-up)

### B1'. Schema-equivalence CI step fails on all 5 PG versions — `ash-1.3-to-1.4.sql` is out of sync with the new install.sql (confidence 10)

Run 24869319187 fails the "Schema equivalence: fresh install vs full upgrade path" step across test (14), (15), (16), (17), (18). Diff excerpt from the log:

```
-COL|config|insert_errors|bigint|t|0
-COL|config|register_wait_cap_hits|bigint|t|0
-FUNC|_register_wait|...|smallint|e9f4f4f65e5229dcafe1f6d3ffa9885b
+FUNC|_register_wait|...|smallint|396bde3c87427bf0fff1e621ce116a0c
-FUNC|status|...|4bb58cb06c2273ba6f2742e0c0974b25
+FUNC|status|...|c0f4274e6dac646c07ba83776a276523
-FUNC|take_sample|...|8d5700299e67db5242f1fb372ce80422
+FUNC|take_sample|...|18cee26662e8246c2dd5085204cf69c9
-FUNC|start|...|6aaf9a129d11ffb7db6e6e4f389f6436
+FUNC|start|...|1a96027084c433c5c64c718bf69368c8
```

Root cause: the fix added `insert_errors` and `register_wait_cap_hits` columns and rewrote `_register_wait` / `take_sample` / `status()` / `start()` in `sql/ash-install.sql`, but `sql/ash-1.3-to-1.4.sql` was not updated in lockstep — it has zero occurrences of `insert_errors` or `register_wait_cap_hits` (verified), still uses `32767` as the `wait_event_map_utilization` denominator at line 308, and still ships the older `_register_wait` / `take_sample` bodies. When CI runs 1.0 -> 1.1 -> 1.1-to-1.2 -> 1.2-to-1.3 -> 1.3-to-1.4 and snapshots pg_catalog, the upgraded DB is missing columns and has different `prosrc` MD5s for four functions vs. the fresh install.

CLAUDE.md says "upgrade scripts are generated at release time, not in feature PRs." That guidance conflicts with test.yml's schema-equivalence step, which does execute the upgrade chain on every PR. In this PR the mismatch is load-bearing (CI is blocking merge). Either:

1. Extend `sql/ash-1.3-to-1.4.sql` to add both columns (idempotent) and to re-`create or replace` `_register_wait`, `take_sample`, `status()`, and `start()` with the exact 1.4 bodies — this is the standard "keep the upgrade in sync" move, but goes against the CLAUDE.md release-time policy; or
2. Decide the schema-equivalence step shouldn't gate this PR and adjust `test.yml` (e.g. only run upgrade-path equivalence against the `v1.4` tag, not against `main`). That's a policy call for the project owner.

Either way, the current HEAD cannot merge with red CI.

## Remaining concerns (not blocking)

- **P1 residual (confidence 6).** The prior-review P1 ("cap sentinel silently mis-attributes to id=1") is structurally resolved — the cap branch now returns NULL and the INNER JOIN drops the row — but the drop is still silent at the per-sample level: backend(s) whose (state,type,event) was past the cap are simply absent from `ash.sample.data` for that tick. `register_wait_cap_hits` counts the *registration* drops (one per unique unseen event past the cap per tick), not the number of backends whose samples were lost as a result. For a site with 1 000 backends all waiting on a newly-introduced event past the cap, you would see `register_wait_cap_hits += 1` but lose 1 000 backend observations, and `active_count` in those rows would under-count. Worth a sentence in the doc comment and/or exposing a secondary "dropped backends due to cap" counter in a later PR.
- **register_wait_cap_hits under steady-state cap-hit (confidence 5).** Once past 32 000, every tick that sees at least one new (state,type,event) will `raise warning` — if the site genuinely has a long-tail of rare waits (custom extensions emitting per-xid event names, for example) this will flood the server log. The underlying `not exists` guard at line 391-394 plus the `v_seen_waits` dedup means at most one warning per unique event per tick, but unbounded-cardinality event names (e.g. user-written wait events with counters) will still produce one warning per distinct event per tick forever. A rate limit on the WARNING (e.g. warn once per N ticks once cap is hit, rely on the counter thereafter) would be kinder to log volume.
- **Install header still says "Version: 1.3"** at sql/ash-install.sql:2 while config default is `'1.4'` (line 67). Pre-existing, not introduced here, but flagged because a future reader will trust the header. Trivial fix.

## CI

`gh pr checks 41 --repo NikolayS/pg_ash`:

- test (14) FAIL — schema equivalence
- test (15) FAIL — schema equivalence
- test (16) FAIL — schema equivalence
- test (17) FAIL — schema equivalence
- test (18) FAIL — schema equivalence
- Analyze (actions) pass
- CodeQL pass

Prior failure mode (B1, "cannot change return type") is gone. New failure mode (B1', schema diff on columns + function bodies) is red on all 5 jobs of run 24869319187.

## Top actions for author

1. Resolve B1' so CI is green. Options above; the project owner should decide between "sync the upgrade script now" (contradicts CLAUDE.md release-time policy but is the simplest path) and "adjust test.yml to not block this PR on upgrade parity" (respects the policy but requires CI plumbing).
2. Consider whether P1 residual (sessions lost past cap, not just events) deserves a separate counter or a doc note in 1.5.

NikolayS:
## REV follow-up #2 — Verdict: **STILL BLOCKING**

CI is red on all five PG versions (14/15/16/17/18); the schema-equivalence step fails with the same signature as the prior re-REV. Commit `52c9736` synced four function bodies byte-for-byte (`_register_wait`, `take_sample`, `start`, `status` — I verified each against `sql/ash-install.sql` and got identical md5s: `ce81da72…`, `0246fb72…`, `fdf250d6…`, `21f0b7e1…`), and the two new `ash.config` ALTERs are in the right place (lines 41/47, before any function body that reads those columns) with matching `bigint not null default 0` shape. The P1 clarifier comment on the cap-hit counter is in place. However, commit `67b0883` also modified **`ash.decode_sample`** (M-BUG-9: two-pass validation, reworded warning, new non-positive-count guard) and that function was **not** added to `ash-1.3-to-1.4.sql`. CI surfaces exactly this: the only hash diff in the schema-equivalence output is `FUNC|decode_sample|…|942bc938…` (fresh) vs `36d9a017…` (upgraded). I confirmed `942bc938…` matches the current install.sql body bit-for-bit; the upgraded side is the stale v1.3 body carried through.

**Remaining concerns (confidence ≥ 7):**
- B1 (blocker): add a `create or replace function ash.decode_sample(...)` block to `sql/ash-1.3-to-1.4.sql` copied verbatim from `sql/ash-install.sql` (lines ~565–687). Nothing else appears to be out of sync — I did a full diff of `main..HEAD` on install.sql and only `_register_wait`, `take_sample`, `decode_sample`, `start`, `status` have body-level changes; the first four are handled and only `decode_sample` is missing.
- B2 (process, needs owner decision): `CLAUDE.md` states upgrade scripts are "generated at release time, not in feature PRs," but `.github/workflows/test.yml` (line 1330) runs `\i sql/ash-1.3-to-1.4.sql` and the schema-equivalence step compares it against `ash-install.sql` on every push. These two policies are incompatible — every correctness PR that touches an install.sql function body currently must ship a matching upgrade hunk or CI fails. This repeat-miss is a symptom. Recommend the repo owner either (a) relax `CLAUDE.md` to say upgrade scripts ship with the PR that changes install.sql, or (b) gate the upgrade-path and schema-equivalence steps on a `[release]` label / tag so feature PRs stop bouncing on it. Flagging explicitly because downstream PRs will keep hitting this.

No new regressions observed — ordering of ALTERs vs. function redefinitions is correct, and no reference to the new columns appears before their `add column if not exists`.

NikolayS:
## REV follow-up 3 — fresh review after commit `2017091`

**Verdict: RESOLVED**

### CI

All 7 checks green on `2017091`: `test (14..18)` + `CodeQL` + `Analyze`. The schema-equivalence assertion that failed on the prior two revisions now passes across all five PG versions.

### Byte-identity of modified functions

Five functions differ from `origin/main` in `sql/ash-install.sql`; I extracted each body between its `create or replace function …` header and the matching `$$;` terminator in both files and compared MD5s:

| function | install.sql (lines) | ash-1.3-to-1.4.sql (lines) | MD5 match |
|---|---|---|---|
| `ash._register_wait` | 175..237 (63) | 54..116 (63) | IDENTICAL `3998c514…` |
| `ash.take_sample` | 315..553 (239) | 119..357 (239) | IDENTICAL `763ea620…` |
| `ash.decode_sample` | 565..680 (116) | 362..477 (116) | IDENTICAL `839b17f3…` |
| `ash.start` | 787..1005 (219) | 481..699 (219) | IDENTICAL `479e2905…` |
| `ash.status` | 1135..1219 (85) | 702..786 (85) | IDENTICAL `cb926801…` |

No function that differs between branch install.sql and main is missing from the upgrade script.

### `decode_sample` signature

`returns table (wait_event text, query_id int8, count int)` — unchanged from legacy `ash-1.2-to-1.3.sql` (lines 358–364 on both). Re-applying the chain over an existing install won't trigger `cannot change return type`. The widened local `v_wait_id int` (was `smallint` in legacy) is a declaration inside the body and doesn't affect the signature.

### Column ALTER ordering

In `ash-1.3-to-1.4.sql`:

- `insert_errors` added at line 41
- `register_wait_cap_hits` added at line 47
- First function body at line 54

All references to the new columns (lines 88, 330, 347, 743, 747, 766) live inside function bodies, so they're resolved at call time, long after the ALTERs have committed at script-run time. Ordering is safe.

### Other regressions scanned

None found. The `missed_samples` column is defined inline in the fresh-install `ash.config` CREATE TABLE and backfilled idempotently in the upgrade via a `do $$ … add column … $$;` block; `insert_errors` / `register_wait_cap_hits` use `add column if not exists` in both files, which is consistent with the idempotent re-apply pattern.

### Remaining concerns

None at confidence >= 7. Shipping-ready from correctness/schema-equivalence standpoint. The inline doc comment in the upgrade script still flags the CLAUDE.md policy tension (upgrade scripts in feature PRs) — that's a process question for the owner, not a blocker.

PR #42, opened 2026-04-22
perf: hot-path improvements in query_waits, cap probe, rotate()

NikolayS:
## Summary

Three performance fixes on hot paths, all in `sql/ash-install.sql`. Closes portions of #37.

- **M-BUG-7**: `ash.query_waits()` / `ash.query_waits_at()` — replace the `generate_series(i, 1, -1)` correlated walk-back with a running-group window (`sum(case when v < 0 …)` assigning a group id per negative wait marker; `-min(v)` per group yields the `wait_id`). Work per sample goes from **O(N^2)** to **O(N)**; over hour-long reader windows with many samples this compounds away.
- **M-BUG-5**: `ash.take_sample()` — replace the `reltuples >= 50000` partition-cap check with `not exists (select 1 from ash.query_map_N offset 49999 limit 1)`. `reltuples` is a planner estimate and can remain stale or be `-1` immediately after `TRUNCATE` (and we truncate on every `rotate()`); the index probe is immediately accurate and still O(1) at the 50000th row.
- **L-BUG-12**: `ash.rotate()` — collapse the two-statement `truncate … alter column id restart` sequence into a single `truncate ash.sample_N, ash.query_map_N restart identity`. One `AccessExclusiveLock` acquisition per rotation instead of two. `sample_N` has no identity column, so `RESTART IDENTITY` only affects `query_map_N`.

### Before / after complexity

| Site | Before | After |
|---|---|---|
| `query_waits[_at]` per sample | O(N^2) walk-back from every qid | O(N) single-pass window |
| `take_sample()` cap check | `pg_class.reltuples` (stale after TRUNCATE, can be -1) | `EXISTS … OFFSET 49999 LIMIT 1` index probe |
| `rotate()` per slot | 2 statements × 2 AEL | 1 statement × 1 AEL |

## Test plan

- [x] Fresh install of `sql/ash-install.sql` on PostgreSQL 16 — no syntax errors
- [x] Synthetic multi-group sample `[-1, 2, qid_a, qid_b, -2, 1, qid_c]` — `query_waits(qid_a)` resolves to wait 1, `query_waits(qid_c)` resolves to wait 2, `query_waits` of a qid appearing under both returns both with correct counts
- [x] Multi-row samples aggregated correctly across `query_waits` and `query_waits_at`
- [x] `ash.rotate()` truncates both `sample_N` and `query_map_N` in one statement; identity sequence restarts from 1 on next insert
- [ ] CI: PG 14–18 matrix (fresh install, upgrade path, schema equivalence, degraded mode)

Refs #37 (M-BUG-5, M-BUG-7, L-BUG-12)

### DIFF
--- a/sql/ash-1.3-to-1.4.sql
+++ b/sql/ash-1.3-to-1.4.sql
@@ -127,6 +127,8 @@ begin
   -- Existence probe at the 50000th row: one index lookup, and — unlike
   -- pg_class.reltuples — immediately accurate after TRUNCATE (reltuples
   -- can remain stale or be -1 until autovacuum/ANALYZE catches up).
+  -- TODO: verify via EXPLAIN that the uncorrelated probe subquery is
+  -- hoisted to InitPlan (executed once) rather than re-evaluated per row.
   if v_current_slot = 0 then
     insert into ash.query_map_0 (query_id)
     select distinct sa.query_id
@@ -499,7 +501,6 @@ begin
       case when wm.event = wm.type then wm.event else wm.type || ':' || wm.event end as evt
     from hits h
     join ash.wait_event_map wm on wm.id = h.wait_id
-    where h.wait_id is not null
   ),
   totals as (
     select evt, count(*) as cnt
@@ -602,7 +603,6 @@ begin
       case when wm.event = wm.type then wm.event else wm.type || ':' || wm.event end as evt
     from hits h
     join ash.wait_event_map wm on wm.id = h.wait_id
-    where h.wait_id is not null
   ),
   totals as (
     select evt, count(*) as cnt from named_hits group by evt
--- a/sql/ash-1.3-to-1.4.sql
+++ b/sql/ash-1.3-to-1.4.sql
@@ -124,6 +124,9 @@ begin
   -- PG14-15 volatile SQL comments can flood query_map. 50k hard cap per
   -- partition prevents unbounded growth. PG16+ normalizes comments.
   -- NOTE: 3-way IF for clarity on hot path. Bug fixes must be applied 3×.
+  -- Existence probe at the 50000th row: one index lookup, and — unlike
+  -- pg_class.reltuples — immediately accurate after TRUNCATE (reltuples
+  -- can remain stale or be -1 until autovacuum/ANALYZE catches up).
   if v_current_slot = 0 then
     insert into ash.query_map_0 (query_id)
     select distinct sa.query_id
@@ -133,8 +136,7 @@ begin
       and (sa.backend_type = 'client backend'
        or (v_include_bg and sa.backend_type in ('autovacuum worker', 'logical replication worker', 'parallel worker', 'background worker')))
       and sa.pid <> pg_backend_pid()
-      and (select reltuples from pg_class
-       where oid = 'ash.query_map_0'::regclass) < 50000
+      and not exists (select 1 from ash.query_map_0 offset 49999 limit 1)
     on conflict (query_id) do nothing;
   elsif v_current_slot = 1 then
     insert into ash.query_map_1 (query_id)
@@ -145,8 +147,7 @@ begin
       and (sa.backend_type = 'client backend'
        or (v_include_bg and sa.backend_type in ('autovacuum worker', 'logical replication worker', 'parallel worker', 'background worker')))
       and sa.pid <> pg_backend_pid()
-      and (select reltuples from pg_class
-       where oid = 'ash.query_map_1'::regclass) < 50000
+      and not exists (select 1 from ash.query_map_1 offset 49999 limit 1)
     on conflict (query_id) do nothing;
   else
     insert into ash.query_map_2 (query_id)
@@ -157,8 +158,7 @@ begin
       and (sa.backend_type = 'client backend'
        or (v_include_bg and sa.backend_type in ('autovacuum worker', 'logical replication worker', 'parallel worker', 'background worker')))
       and sa.pid <> pg_backend_pid()
-      and (select reltuples from pg_class
-       where oid = 'ash.query_map_2'::regclass) < 50000
+      and not exists (select 1 from ash.query_map_2 offset 49999 limit 1)
     on conflict (query_id) do nothing;
   end if;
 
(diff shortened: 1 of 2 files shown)


NikolayS:
# REV code review — PR #42

**PR:** `perf: hot-path improvements in query_waits, cap probe, rotate()`
**Branch:** `claude/issue-37-perf` → `main`
**Scope:** 2 files, +412 / −45 — `sql/ash-install.sql` + `sql/ash-1.3-to-1.4.sql`
**CI:** all green (PG 14/15/16/17/18, CodeQL, Analyze)

## Verdict

**NON-BLOCKING — approve with minor follow-ups.**

Three targeted changes on the sampling hot path. Algorithmic claims hold up under inspection: the `query_waits` rewrite is genuinely O(N) per sample vs. the prior O(N²) correlated walk-back, the `reltuples` → `exists … offset 49999 limit 1` swap fixes a real correctness bug (stale/−1 reltuples after TRUNCATE), and the single-statement `truncate … restart identity` reduces `AccessExclusiveLock` acquisitions from 2 to 1 per rotated slot. Semantics preserved for all three.

## Per-area counts

| Area | Blocking | Non-blocking | Potential |
|---|---|---|---|
| Bugs / correctness | 0 | 0 | 1 |
| Security | 0 | 0 | 0 |
| Guidelines | 0 | 1 | 1 |
| Tests | 0 | 0 | 2 |
| Docs | 0 | 0 | 1 |

## Findings

### POTENTIAL — `offset 49999 limit 1` cost when near cap (bug-hunter, conf 5)

- **File:** `sql/ash-install.sql:371`, `:382`, `:393` (plus same in upgrade script)
- **Issue:** The new `not exists (select 1 from ash.query_map_N offset 49999 limit 1)` is semantically correct and resolves the stale-`reltuples` bug, but when the partition is near the cap PostgreSQL must scan ~50000 index/heap tuples to resolve the offset on every `take_sample()` call (1 Hz). `reltuples` was one catalog tuple. The subquery is uncorrelated, so the planner should hoist it to an InitPlan (one evaluation per sampler call), which keeps the cost bounded — but this is not free on a hot path and the PR body does not measure it.
- **Evidence:** `insert into ash.query_map_0 … and not exists (select 1 from ash.query_map_0 offset 49999 limit 1)`
- **Suggestion:** Non-blocking. Consider an `EXPLAIN (ANALYZE, BUFFERS)` at 45k–50k rows to confirm the cost stays sub-millisecond; if it doesn't, a seq-count on a smallint counter column or a cached count in `ash.config` would be cheaper. Correctness is not in question.

### POTENTIAL — upgrade script touched in a feature PR (guidelines-checker, conf 6)

- **File:** `sql/ash-1.3-to-1.4.sql`
- **Issue:** CLAUDE.md §Deployment: *"Upgrade scripts (`ash-X.Y-to-X.Z.sql`) are generated at release time, not in feature PRs."* This PR appends to an existing `1.3-to-1.4.sql`. If 1.4 has not been tagged yet, this is pragmatic; if it has, it breaks the reproducibility contract.
- **Suggestion:** Confirm 1.4 is still unreleased. If released, move the perf changes to `ash-1.4-to-1.5.sql` and bump `ash.config.version`. Otherwise add a note in the PR description that 1.4 is still in-flight.

### NON-BLOCKING — identity comment accuracy (docs-reviewer, conf 8)

- **File:** `sql/ash-install.sql:638-648`, `sql/ash-1.3-to-1.4.sql:378-388`
- **Issue:** Comment claims *"`sample_N` has no identity column, so it is unaffected"* by `RESTART IDENTITY`. Verified against the install script — `ash.sample` has no identity column (only `sample_ts`, `datid`, `active_count`, `data`, `slot`). Comment is correct and helpful. Minor nit: `RESTART IDENTITY` in multi-table TRUNCATE restarts **owned** sequences only, so technically sample_N would be unaffected even if it had a non-owned sequence — the comment is precise enough.
- **Suggestion:** None. Kept as NON-BLOCKING informational so the rationale is on the record.

### POTENTIAL — dead filter in named_hits (bug-hunter, conf 5)

- **File:** `sql/ash-install.sql:1573`, `:1965` (both query_waits variants)
- **Issue:** `where h.wait_id is not null` in `named_hits` — under the new window rewrite, `-min(v) over (partition by ctid, slot, grp)` is never NULL because every group contains exactly one negative marker by construction (the running-sum guarantees this). The filter is a no-op.
- **Suggestion:** Drop the predicate or add a comment noting it's defensive against malformed samples. Purely cosmetic.

### POTENTIAL — missing direct test for the O(N) rewrite (test-analyzer, conf 6)

- **File:** PR test plan
- **Issue:** PR body claims the synthetic `[-1, 2, qid_a, qid_b, -2, 1, qid_c]` multi-group case was checked manually, but no repeatable test fixture lands in the test harness. Given the window expression is subtle (running-group by `sum(case when v<0 then 1 end)`, `-min(v) over (partition by grp)`), a committed regression test that seeds a known `ash.sample.data` array and asserts `query_waits` output would be cheap insurance.
- **Suggestion:** Add a test that inserts a hand-crafted `sample.data` covering: (a) multiple wait groups per sample, (b) a qid that appears under two different wait events in the same sample, (c) a qid as the last element of a group, (d) the first qid immediately after the count. Non-blocking if the existing fresh-install smoke test covers behavior end-to-end.

### POTENTIAL — no benchmark artifact in PR (test-analyzer, conf 4)

- **File:** PR body
- **Issue:** Performance PR without EXPLAIN (ANALYZE) numbers or timing. Claims of O(N²)→O(N) are structurally correct from the SQL, but "the win is big" is asserted, not measured. For a hot path running 1 Hz this matters.
- **Suggestion:** Attach `EXPLAIN (ANALYZE, BUFFERS)` before/after for `query_waits()` on a realistic sample table (say 3600 rows, avg 50-element data arrays) in the PR description or a follow-up comment. Non-blocking.

## Notes on preserved semantics (checked)

- `query_waits` / `query_waits_at`: new grouping via running sum over negative markers correctly reconstructs `(-wait_id, count, qid_1 … qid_count)` groups per the documented layout (see `ash.sample.data` comment and `ash._validate_sample`). `-min(v) over (partition by ctid, slot, grp)` yields the group's wait_id since each group contains exactly one negative by construction. Filter `v >= 0 AND prev_v >= 0 AND pos > 1` still distinguishes qid positions from marker and count positions.
- Cap probe: `not exists … offset 49999 limit 1` is semantically equivalent to `count >= 50000` at the 50000-row threshold and, crucially, immediately accurate after `TRUNCATE` (unlike `pg_class.reltuples`, which can be `-1` until ANALYZE runs). The cap is now actually enforced right after a rotation, which was a silent correctness bug before.
- `rotate()`: `truncate a, b restart identity` is atomic and equivalent to two `TRUNCATE`s plus `ALTER … RESTART`. One `AccessExclusiveLock` acquisition per slot, reduced from two. Advisory lock and `lock_not_available` handling unchanged.

## Hot-path regression check

- No new syscalls, no new catalog lookups introduced in `take_sample()`. The `offset 49999 limit 1` probe replaces a `pg_class` catalog lookup with an index-only scan on the partition's own PK — catalog access is strictly reduced. No `pg_stat_statements` or other extension reads added. ✓

---

PR #43, opened 2026-04-22
fix(readers): oversized-interval NOTICE, dead code removal, column rename

### LINKED ISSUE #37: Comprehensive review: correctness, security, CI, and docs findings (v1.4)

NikolayS:
## Summary

Three small reader-UX and cleanup fixes from pg_ash issue #37:

- **H-BUG-3** — readers now emit a clear signal when the caller requests more history than is retained. `ash._active_slots()` gains an optional `p_interval` argument and raises `NOTICE` once per transaction if `p_interval > 2 * rotation_period`, recommending a shorter interval or longer `rotation_period`. Dedup uses a transaction-scoped GUC (`ash.notice_oversized`) so multi-query readers (`activity_summary`, `top_queries`, `samples`, `timeline_chart`, etc.) produce exactly one NOTICE per call, not one per partition or sub-query. The returned row set is **never clamped** — callers get an honest empty result with a clear explanation.
- **L-BUG-10** — `ash._validate_data` was defined but never called anywhere in the codebase. Deleted. Corresponding assertions removed from `.github/workflows/test.yml`.
- **L-BUG-11** — `ash.samples_by_database` returned `sum(active_count)` under the misleading column name `total_backends` (implies distinct backends). Renamed to `total_backend_samples`.

## ⚠ Breaking change

`ash.samples_by_database` renamed its last column from `total_backends` to `total_backend_samples`. Anyone `SELECT`ing `total_backends` explicitly will need to update their query. This is a breaking column-level rename — appropriate for the current pre-1.0-style release cadence (1.x with frequent, user-visible changes). The underlying value (`sum(active_count)` across samples) is unchanged. The function signature is re-created via an explicit `drop function if exists` before `create or replace`.

## Test plan

- [x] Fresh `\i sql/ash-install.sql` succeeds on PG 16
- [x] `ash._validate_data` no longer exists after install
- [x] `ash.samples_by_database` returns a column named `total_backend_samples`
- [x] `select * from ash.top_waits(interval '3 days')` emits one NOTICE (with data present)
- [x] `select * from ash.activity_summary(interval '3 days')` emits exactly one NOTICE despite calling several sub-readers
- [x] Normal-sized intervals emit no NOTICE
- [x] `_at` variants (timestamptz ranges) are untouched
- [x] CI matrix covers PG 14–18 via existing `test.yml`

Closes #37 partially (H-BUG-3, L-BUG-10, L-BUG-11).


NikolayS:
# REV Review — PR #43

`fix(readers): oversized-interval NOTICE, dead code removal, column rename` — +1024 / -127, 3 files.

## Verdict

**Request changes — 1 BLOCKING (CI red), 1 BLOCKING (policy), plus potentials.**

CI is red on PG 14/15/16/17/18 (same root cause); a repo-policy violation around upgrade-script placement; and the column rename has unpatched downstream doc references.

---

## Findings

### BLOCKING

#### B1. Idempotent-reapply path breaks: `ash._active_slots()` becomes ambiguous (bug, 10/10)

CI fails identically on all five PG versions with:

```
psql:sql/ash-1.2-to-1.3.sql:1056: ERROR:  function ash._active_slots() is not unique
LINE 24:       and s.slot = any(ash._active_slots())
HINT:  Could not choose a best candidate function.
```

Root cause — the PR changes `ash._active_slots` from 0-arg to `ash._active_slots(p_interval interval default null)`. The CI job *"Dev workflow: re-apply migrations (idempotent)"* (test.yml L1406–1445) runs `\i ash-install.sql` (fresh 1.4 → has `_active_slots(interval)` only), then replays `ash-1.1-to-1.2.sql`, `ash-1.2-to-1.3.sql`, `ash-1.3-to-1.4.sql` on top. `ash-1.2-to-1.3.sql:839` does `create or replace function ash._active_slots() returns smallint[] ...` — a *different* signature from `_active_slots(interval)`, so PG creates a second overload rather than replacing. Subsequent no-arg calls inside that same script (L1015, L1056, etc.) then match both overloads and error out.

PR description claims: *"Fresh `\i sql/ash-install.sql` succeeds on PG 16"*. Correct. But the matrix-wide *reapply-idempotency* test is not covered by the PR's test plan and fails on every runner.

Suggested fix (without fixing it here): either (a) have `ash-1.3-to-1.4.sql` also `drop function if exists ash._active_slots()` at a point that survives re-apply and teach `ash-1.2-to-1.3.sql` to tolerate the new signature, or (b) keep the same 0-arg signature and make the p_interval mechanism internal (e.g. optional GUC or an overload added without dropping the original — though that re-creates ambiguity). The cleanest path is a guard at top of `1.3-to-1.4.sql` that drops both signatures, *plus* a note that the 1.2-to-1.3 idempotent re-apply will temporarily recreate the 0-arg form until 1.3-to-1.4 drops it again on re-run. Given the upgrade scripts must replay idempotently in any order documented by test.yml, the simplest robust fix is to change `ash-1.2-to-1.3.sql` to `drop function if exists ash._active_slots();` before its `create or replace`, mirroring the PR's 1.3-to-1.4 treatment — but CLAUDE.md says upgrade scripts are generated at release time (see B2).

#### B2. Upgrade script `ash-1.3-to-1.4.sql` is hand-edited in a feature PR (policy, 9/10)

CLAUDE.md L17 is explicit: *"`ash-install.sql` is the source of truth. Upgrade scripts (`ash-X.Y-to-X.Z.sql`) are generated at release time, not in feature PRs."* This PR adds ~960 lines (`+335..+1294`) to `sql/ash-1.3-to-1.4.sql` duplicating function bodies from `ash-install.sql`. That contradicts the stated release workflow and doubles the maintenance surface for every reader function touched. It is also why B1 exists at all — if the upgrade script were generated at release time from `ash-install.sql`, the generator could (and should) handle the cross-version `_active_slots` signature transition in one place.

Recommend: revert all `sql/ash-1.3-to-1.4.sql` additions in this PR and generate the upgrade script at 1.4 tag time.

### POTENTIAL

#### P1. Column rename is a breaking change; README + docs still say `total_backends` (6/10)

`ash.samples_by_database` column `total_backends` → `total_backend_samples` with no deprecation path. CLAUDE.md implies pure-SQL `\i` reinstalls — there is **no** extension upgrade mechanism, so existing user queries (`SELECT total_backends FROM ash.samples_by_database(...)`) silently break at query time with `column "total_backends" does not exist`. The PR body acknowledges this as intentional pre-1.0-style cadence.

Issues:
- README.md L113 still documents the old signature (unchanged by PR). At minimum the column rename deserves a brief mention in the README/CHANGELOG.
- A short-lived backward-compat overload that returns both `total_backends` and `total_backend_samples` would be trivial and would protect dashboards/scripts over one release.
- `blueprints/SPEC.md:397` still lists `ash._validate_data(integer[])` in the function catalog (and SPEC.md:53, 74 describe its role). Dead-code removal is correct per grep, but the **spec** is now stale.

This is a POTENTIAL rather than BLOCKING because the breaking change is owner-approved in the PR body — but the docs drift is a concrete fixable gap.

#### P2. `raise notice` level choice for oversized interval (5/10)

NOTICE is appropriate for informational psql users (default `client_min_messages = NOTICE`). However:
- Python/Node drivers frequently set `client_min_messages = WARNING` by default, silencing the diagnostic entirely. For a reader primarily consumed by humans in psql, this is fine; if these readers are called from dashboards/scripts, NOTICE may be invisible in practice.
- WARNING would be noisier but more reliably visible. EXCEPTION is correctly rejected (would break the "honest empty result" goal).

NOTICE is a defensible choice; flag for owner awareness only.

#### P3. Transaction-scoped GUC dedup — edge cases (4/10)

`set_config('ash.notice_oversized', '1', true)` with `is_local = true` auto-resets at commit/rollback. Two observations:
- `current_setting('ash.notice_oversized', true)` returns `''` (empty string) for an unset GUC, not NULL. The code handles both (`is null or = ''`) — good.
- Custom GUC names must contain a dot; `ash.notice_oversized` satisfies this. But without `ALTER DATABASE ... SET` permission registration, PG does permit ad-hoc custom GUCs only because `custom_variable_classes` was removed in PG 9.2 — works on all supported versions (14–18). OK.
- Autocommit single-statement calls (`SELECT * FROM ash.top_waits(interval '3 days')` at the psql prompt) run in an implicit transaction — the GUC scope still works. OK.

No change needed; noted for completeness.

#### P4. Factor of `2 * rotation_period` is undocumented in the message (4/10)

The NOTICE text says *"exceeds 2 * rotation_period"* — but users don't intuitively know why 2×. With 3 partitions retained and rotations mid-cycle, the *actually-available* window is somewhere between 1× and 2× `rotation_period`. A user who sets the interval to exactly `2 * rotation_period` gets no notice, but may still see an empty-ish result if they queried just after a rotation. Consider phrasing: *"only ~2× rotation_period of history is retained (3 partitions, rotating)"*.

Minor; usability polish.

### NON-BLOCKING

#### N1. `raise exception` in degraded-mode branches of `top_queries_with_text` and `event_queries` (3/10)
Existing behavior, not introduced by this PR. Noted because the PR touches both functions without changing this — inconsistent with `top_queries`/`samples` which fall back to NULL query_text with a WARNING. Consider unifying in a separate PR.

#### N2. `blueprints/SPEC.md` references removed `_validate_data` (3/10)
Dead-code removal is verified — `ash._validate_data` has zero runtime callers in `sql/ash-install.sql`, `sql/ash-1.3-to-1.4.sql`, and the removed test assertions in `test.yml` were the only users. But `blueprints/SPEC.md` lines 53, 74, 397 still describe the function as part of the contract. Update spec alongside removal.

---

## CI

Five of seven checks failing (PG 14/15/16/17/18 `test` jobs), same root cause B1. CodeQL + Analyze pass.

## Summary

| Area | BLOCKING | POTENTIAL | NON-BLOCKING |
|---|---|---|---|
| Bugs | 1 (B1) | 1 (P3) | — |
| Policy / process | 1 (B2) | — | — |
| UX / API | — | 2 (P1, P2, P4) | 1 (N1) |
| Docs | — | — | 1 (N2) |

**Top 3 issues to address before merge:**
1. **B1** — Fix `_active_slots` signature conflict surfaced by the idempotent-reapply test (CI red on all 5 PG versions).
2. **B2** — Don't hand-edit `ash-1.3-to-1.4.sql` in a feature PR; generate at release time per CLAUDE.md.
3. **P1** — Update `README.md` and `blueprints/SPEC.md` to reflect the column rename and `_validate_data` removal.

---

NikolayS:
# REV follow-up review (commit f770193)

**Verdict: STILL BLOCKING — B1 ambiguity resolved, but a NEW regression from the column rename breaks the same CI stage (`Dev workflow: re-apply migrations (idempotent)`) on all 5 PG matrix jobs.**

## What was verified on `f770193`

- **B1 (ambiguity) — RESOLVED.** `sql/ash-install.sql` now defines exactly one `ash._active_slots()` (zero-arg SQL, line 927) and a distinctly-named `ash._active_slots_for(p_interval interval)` (plpgsql, line 952). All 18 reader call sites line up cleanly:
  - `_active_slots_for(p_interval)` is only called from readers that themselves take `p_interval`.
  - Bare `_active_slots()` is only called from the `_at` (absolute-time) readers and `timeline_chart_at` / `samples_at` / `event_queries_at`, none of which accept an interval.
  No CI log shows the old "function ... is not unique" error.

- **B2 (policy) — RESOLVED.** `git diff origin/main -- sql/ash-1.3-to-1.4.sql` is empty; the 964-line feature-PR edit to the upgrade script is fully reverted.

- **P1 (SPEC.md) — RESOLVED.** `grep -n _validate_data blueprints/SPEC.md` returns zero hits. All three stale references (section 3.1 invariant bullet, structural-validation block at line 74, function catalog row at 397) are gone. Tests in `.github/workflows/test.yml` referencing `_validate_data` were removed in 5a413b8.

- **Reviewer-disagreement on README.md:113 — author was right, prior review was wrong.** Line 113 reads ``| `ash.samples_by_database(interval)` | Per-database activity |`` — function signature only, no column names. `grep -n 'total_backends\|total_backend_samples' README.md` returns zero hits. No doc churn needed there.

- Archived migration scripts (`sql/ash-1.0.sql`, `sql/ash-1.1.sql`, `sql/ash-1.2-to-1.3.sql`) still contain `_validate_data` and `total_backends` — correct and expected; those files are immutable history.

## NEW BLOCKING (B3): column rename breaks idempotent re-apply

All 5 matrix jobs (PG 14/15/16/17/18) fail at the same step with the same error:

```
psql:sql/ash-1.2-to-1.3.sql:1396: ERROR:  cannot change return type of existing function
DETAIL:  Row type defined by OUT parameters is different.
HINT:  Use DROP FUNCTION ash.samples_by_database(interval) first.
```

Cause: commit 5a413b8 renamed the `samples_by_database` output column `total_backends` → `total_backend_samples`. `sql/ash-install.sql:1557` guards this with a preceding `drop function if exists ash.samples_by_database(interval);`, so fresh install is clean. But the CI "re-apply migrations (idempotent)" stage runs install + 1.0→1.1→1.1→1.2→1.2→1.3 on top. `sql/ash-1.2-to-1.3.sql:1372-1396` does a bare `create or replace function ash.samples_by_database(...)` with the OLD column name `total_backends`. Postgres rejects that because the return rowtype differs from the already-installed 1.4 version.

This is functionally the same class of failure B1 was — an old upgrade script re-applied on top of 1.4 hits a signature/rowtype mismatch the install changed. The ambiguity variant is fixed; the rowtype-drift variant is new and unaddressed.

Minimal paths forward (choose one, author's call):
- Keep the old column name (revert the rename portion of L-BUG-11, document the meaning in a comment instead), or
- Gate the rename behind the upgrade script only — i.e. ship the new output column via `sql/ash-1.3-to-1.4.sql` at release time, not inside `ash-install.sql` for a branch that still has to survive replay of `ash-1.2-to-1.3.sql`, or
- Bite the bullet and have `ash-install.sql` pre-emptively `drop function` before the re-apply of any old script that would recreate it — but that belongs in a migration, not install.

The cleanest option given CLAUDE.md policy is to defer the breaking column rename until v1.4 cuts, since the idempotent-replay test is explicitly in scope.

## Remaining concerns (confidence ≥ 7)

1. **CI red on all 5 jobs (confidence 10).** B3 above. Blocking.
2. **Commit message for f770193 still advertises the rename as shipped in this PR (confidence 7).** If the author chooses to defer the rename, the commit message and PR description need a matching revision so the release notes don't claim a breaking change that didn't land.

## What I did NOT re-verify

- Semantic correctness of `_active_slots_for` NOTICE dedup across nested transactions / savepoints (prior review marked this out of scope; `SET LOCAL` semantics + GUC fallback via `current_setting(..., true)` look correct on a read).
- Whether `total_backend_samples` is actually the better name — that's a naming debate, not a review gate.

---

NikolayS:
## REV follow-up #2

**Verdict: STILL BLOCKING (regression)**

CI is red on all five PG versions (14, 15, 16, 17, 18) — the `Schema equivalence: fresh install vs full upgrade path` job fails for every matrix entry (latest run [24870560767](https://github.com/NikolayS/pg_ash/actions/runs/24870560767)). The diff shows the upgrade path produces a materially different schema than a fresh install: `_active_slots_for(p_interval interval)` is missing, `_validate_data(p_data integer[])` is still present, and the `md5(prosrc)` hashes for ~10 reader functions (`activity_summary`, `event_queries`, `query_waits`, `samples`, `samples_by_database`, `timeline_chart`, `top_by_type`, `top_queries`, `top_queries_with_text`, `top_waits`, `wait_timeline`) differ. Root cause: ceb77ec correctly undid the breaking column rename in `ash-install.sql`, and f770193 correctly reverted `sql/ash-1.3-to-1.4.sql` to match main — but `ash-install.sql` still carries the rest of the reader refactor (the `_active_slots_for` split, the `_validate_data` deletion, and the updated reader bodies), so main's unmodified 1.3→1.4 script no longer reaches schema parity. The earlier blockers are genuinely gone (`samples_by_database` RETURNS TABLE is back to `(database_name text, datid oid, samples bigint, total_backends bigint)`, `grep -rn total_backend_samples` returns nothing, `_active_slots` and `_active_slots_for` are distinct, `sql/ash-1.3-to-1.4.sql` is byte-identical to main, `blueprints/SPEC.md` is `_validate_data`-free), but the PR cannot merge while CI is red.

**Remaining concern (confidence 10):** the fix has to either (a) extend `sql/ash-1.3-to-1.4.sql` to add `_active_slots_for`, drop `_validate_data`, and re-install the updated reader bodies so the upgrade path matches a fresh install, or (b) defer the `_active_slots_for` / `_validate_data` / reader-body changes out of this PR entirely. Per CLAUDE.md ("upgrade scripts are generated at release time, not in feature PRs"), option (b) looks cleaner, but whichever path is chosen, schema equivalence must pass before merge — the test is doing exactly what it is designed to do.

NikolayS:
**Verdict: RESOLVED**

All seven CI jobs are green on commit `ffe9d10` (Analyze, CodeQL, test 14/15/16/17/18). Byte-identity verification confirms all 12 target functions (`_active_slots_for`, `top_waits`, `wait_timeline`, `top_queries`, `top_by_type`, `top_queries_with_text`, `query_waits`, `samples_by_database`, `activity_summary`, `timeline_chart`, `samples`, `event_queries`) have matching argument lists, RETURNS clauses, and body md5s between `sql/ash-install.sql` and `sql/ash-1.3-to-1.4.sql`. `drop function if exists ash._validate_data(integer[])` is at line 336 of the upgrade script — correctly positioned before the `_active_slots_for` creation at line 350. The previous fixes remain intact: `ash._active_slots()` (0-arg) and `ash._active_slots_for(interval)` are distinct signatures in install.sql (no overload ambiguity), and `samples_by_database` still returns a column named `total_backends` in both files (the `total_backend_samples` rename is fully reverted, with zero occurrences anywhere).

Cross-check against `git diff origin/main -- sql/ash-install.sql`: 13 reader-related functions changed on this branch (12 modified/added + 1 dropped `_validate_data`), and all 13 are accounted for in the upgrade script. The only functions in ash-1.3-to-1.4.sql not modified by this PR are `take_sample` and `status`, which are the pre-existing 1.3→1.4 work from PR #29 and are unrelated.

No remaining concerns at confidence >=7. Ship it.

PR #44, opened 2026-04-22
ci: coverage & robustness (wrapper, data preservation, pg_cron wiring)

NikolayS:
## Summary

Closes the CI coverage gaps from #37 (H-CI-1/2, M-CI-4/5/6, L-CI-7). H-CI-3 is deferred to a follow-up (see below).

- **H-CI-1** new job exercises the documented `ash-X.Y-to-X.Z.sql` wrapper chain (the one the README tells users to run). The previous test shortcut by loading `ash-1.1.sql` directly, so the wrapper was never CI-verified.
- **H-CI-2** same job seeds a non-default `ash.config`, a `wait_event_map` row, a `query_map_0` row, and an `ash.sample` referencing sentinel `query_id = 424242` at the v1.0 schema; asserts every fixture survives all four upgrades and that newly-added columns (`debug_logging`, `missed_samples`) land at sensible defaults.
- **H-CI-3** _(scope reduced)_ the end-to-end "pg_cron actually fires `take_sample` on schedule" assertion is deferred to #46 — pg_cron's scheduler in the GitHub Actions service container hits `FATAL: database "root" does not exist` (peer-auth default dbname = OS user `root`) and four fix iterations did not converge. What _is_ in this PR: a new "Verify pg_cron wiring" step that asserts `shared_preload_libraries`, GUCs, and launcher bgworker attachment (useful upstream-debugging baseline), a gated `RAISE LOG` in `ash.start()` for diagnostics, and the byte-sync of `ash.start()` in `sql/ash-1.3-to-1.4.sql`. The "cron actually fires" assertion itself is removed with a pointer to #46.
- **M-CI-4** adds a `matrix.cron: ["on"]` axis plus an `include: {postgres: 17, cron: "off"}` entry. The no-cron branch is now exercised on a PG where `pg_cron` is otherwise available, independent of PGDG packaging.
- **M-CI-5** removes `|| true` / `2>/dev/null` silencing where the wrapped command must succeed (`pg_stat_statements` create, degraded-mode restore). Kept with an explanatory comment where silence is intentional (`uninstall` after previous step already dropped the schema).
- **M-CI-6** `docker exec` into the `services.postgres` container is now documented. Switching to `command: -c shared_preload_libraries=...` isn't feasible inside `services:` (no way to set `command` on the GHA-managed service container, and no way to mount a custom `postgresql.conf`), so this is the minimal-diff choice.
- **L-CI-7** dropped no-op `count(*) >= 0` assertions; strengthened to `> 0` where data is seeded, replaced with bare `perform * from ...` (still fails loudly on missing/broken function) in upgrade paths where no data is seeded.

Also quotes `on`/`off` in the matrix to prevent YAML boolean coercion.

## Follow-up

- #46 — `test: assert pg_cron fires sampling jobs in CI`. Describes the peer-auth-dbname root cause and the four iterations attempted on this branch.

## Test plan

- [ ] `test.yml` parses (validated locally with PyYAML)
- [ ] New `{postgres: 17, cron: "off"}` matrix combination runs the degraded-mode-only path
- [ ] "Upgrade path via documented wrappers + data preservation" job asserts sentinel query_id 424242 decodes correctly at v1.4
- [ ] "Verify pg_cron wiring" step prints `shared_preload_libraries` / `cron.database_name` / `cron.use_background_workers` / launcher bgworker presence on every `cron=on` run, non-fatal on `cron=off`
- [ ] All previously-passing `>= 0` checks either become `> 0` or are removed without losing signal

Refs: #37


NikolayS:
## REV code review — PR #44

**Scope**: `.github/workflows/test.yml` (+308 / -39, 1 file). CI coverage and robustness: documented-wrapper upgrade path + data preservation (H-CI-1/2), pg_cron end-to-end scheduling test (H-CI-3), cron=on/off matrix (M-CI-4), removed blanket `|| true`/`2>/dev/null` silencing (M-CI-5), documented `docker exec` approach (M-CI-6), strengthened weak `>= 0` assertions (L-CI-7).

**Verdict**: **BLOCKING** — CI is currently red on every `cron=on` matrix entry (PG 14/15/16/17/18) because the new H-CI-3 step fails deterministically. The `cron=off` variant passes.

Agents applied: `test-analyzer`, `security-reviewer`, `bug-hunter`, `guidelines-checker`, `docs-reviewer`. SOC2 omitted per instructions. Sqitch skipped (N/A).

---

### Per-area counts

| Area | BLOCKING | NON-BLOCKING | POTENTIAL |
|---|---|---|---|
| Test analyzer | 1 | 2 | 1 |
| Bug hunter | 1 | 1 | 1 |
| Security reviewer | 0 | 1 | 1 |
| Guidelines | 0 | 1 | 0 |
| Docs | 0 | 0 | 0 |

---

### BLOCKING

#### B1 — New H-CI-3 end-to-end test fails deterministically on every `cron=on` job (confidence 10)
- **Area**: test-analyzer + bug-hunter
- **File**: `.github/workflows/test.yml` step *"End-to-end: pg_cron actually schedules take_sample (issue #37 H-CI-3)"* (diff lines 100-198)
- **Evidence**: All five `test (N, on)` jobs in run [24765518775](https://github.com/NikolayS/pg_ash/actions/runs/24765518775) fail with:
  ```
  ERROR:  pg_cron did not sample within 30 sec: baseline=1 after=1 waited=30
  cron.job:
    (no ash_* jobs found)
  cron.job_run_details (latest 5):
    (no rows)
  ```
  `pg_cron 1.6.7` is confirmed installed (`apt-get install postgresql-15-cron` succeeds, container is restarted). `(no ash_* jobs found)` means that after `perform ash.start('1 second')`, `cron.job` contains zero `ash_sampler`/`ash_rotation` rows — either `ash.start()` took the "pg_cron not available" branch at `ash-install.sql:718` (most likely: `ash._pg_cron_available()` returns false because the extension was just created but cron launcher/DB wiring races), or `cron.schedule()` was called but the row never materialized. Either way the PR's own new test proves the happy path does not work in CI — which is precisely what the test is meant to catch. Merging with all five `cron=on` matrix jobs red would contradict the CLAUDE.md policy (`Never merge without explicit approval`) and break main.
- **Also**: `baseline=1` (not `0`) is suspicious — the comment at diff line 140 claims "truncate so baseline is unambiguous", but the code only truncates `ash.sample_<current_slot>`. Rows seeded in other slots by earlier steps in the same job survive. This is a latent bug in the new diagnostic: under a passing run, `v_after > v_baseline` could in principle be satisfied by rotation moving rows between slots rather than by a fresh sample.
- **Fix**: before merging, diagnose why no cron rows are scheduled. Candidates:
  - Echo the `ash.start()` return set into the test log (it reports which branch it took).
  - `select jobid, jobname, schedule, active, nodename, database from cron.job` immediately after `ash.start()`.
  - Verify `ash._pg_cron_available()` right before the call.
  - If root cause is TCP-auth to `localhost` for `cron.schedule()`, set `cron.use_background_workers = on` or `cron.host = ''` in the preload config.
  - Truncate **all** slots (`truncate ash.sample_0, ash.sample_1, ash.sample_2`) before reading baseline.

---

### NON-BLOCKING

#### NB1 — Bash-in-YAML interpolation of `${{ matrix.cron }}` directly into shell strings (confidence 8, LOW severity)
- **Area**: security-reviewer
- **File**: `.github/workflows/test.yml` — `if [ "${{ matrix.cron }}" = "off" ]` (diff lines 36, 78)
- **Issue**: `${{ }}` expressions are textually substituted by Actions **before** bash sees them. If a matrix value ever contained e.g. `"off\"; rm -rf /; echo \""`, it would shell-inject. Values here are hard-coded `"on"`/`"off"` in the same file, so actual exploitability is zero. GitHub's hardening guidance (CWE-78) is to pass `${{ }}` through `env:` and reference `"$CRON"` in bash. Fix for hygiene / future-proofing.
- **Fix**: `env: { CRON: ${{ matrix.cron }} }` on the step, then `if [ "$CRON" = "off" ]`.
- **CWE**: CWE-78 (theoretical only given current inputs).

#### NB2 — `docker exec ${{ job.services.postgres.id }}` + heredoc shell with `\${PGDATA}` escaping (confidence 7, LOW severity)
- **Area**: bug-hunter + security-reviewer
- **File**: diff lines 38-43, 51-56, 58-65
- **Issue**: Same `${{ }}`-into-bash pattern as NB1. `job.services.postgres.id` is a GHA-generated hex string (no exploit), but the pattern now appears 5× and the escaping mix (`${{ }}` YAML + `${VAR}` outer bash, which needs `\$` / `${}` nested) is fragile. Any future edit risks silently breaking the inner substitution.
- **Fix**: capture once at step top, `CONTAINER_ID='${{ job.services.postgres.id }}'`, then `docker exec "$CONTAINER_ID" …` throughout.

#### NB3 — `sleep 5` after `docker restart` is a brittle wait (confidence 7, LOW severity)
- **Area**: test-analyzer
- **File**: diff line 42, 69
- **Issue**: Both `cron=off` and `cron=on` paths do `docker restart … ; sleep 5`. 5s is fine on today's runners but is exactly the kind of implicit timing assumption that flakes later. A follow-up "Wait for PostgreSQL" step already exists — the explicit `sleep 5` is redundant, and if it ever becomes insufficient, the resulting failure mode is obscure.
- **Fix**: remove the `sleep 5` and rely on "Wait for PostgreSQL", or replace with `for i in $(seq 1 30); do docker exec "$CONTAINER_ID" pg_isready -U postgres && break; sleep 1; done`.

#### NB4 — Hard-fail on `pg_stat_statements` create eliminates a fallback path (confidence 8, MEDIUM severity)
- **Area**: guidelines-checker
- **File**: diff lines 86-92
- **Issue**: M-CI-5 removes the `|| echo "pg_stat_statements not available"` fallback and hard-fails via `ON_ERROR_STOP=1` + explicit `grep -q 1 || exit 1`. Justification: "pgss ships with every supported PG version". True for pgss itself, but CI uses the stock `postgres:18` image which has historically lagged `contrib` packaging on release day. If a future image briefly omits pgss, **every** `cron=on` job dies at setup rather than degrading — the opposite of the PR's treatment of pg_cron (which is explicitly tolerant).
- **Fix**: either leave a warn-and-continue safety net here (consistent with how the PR treats pg_cron), or pin the base image (`postgres:18-bookworm`) to make the assumption stable. Low probability, high blast radius.

---

### POTENTIAL (confidence 4-7)

#### P1 — `v_baseline` truncation is slot-local; baseline read is cluster-wide (confidence 6)
- **Area**: bug-hunter
- **File**: diff lines 138-142
- **Issue**: `execute format('truncate ash.sample_%s', v_slot)` only empties one partition. Immediately after, `select count(*) into v_baseline from ash.sample` reads the parent (which unions all slots). Prior test steps seed rows in other slots — `baseline=1` in the failure log confirms this. Under a passing run, if rotation fires between `ash.start()` and the poll loop, `v_after > v_baseline` could trigger without any new sample. Tighten: truncate all three slots, or read baseline *after* `ash.stop()`+slot-reset, holding an advisory lock against rotation.

#### P2 — Sentinel query_id 424242 is positive; no negative-id coverage (confidence 4)
- **Area**: test-analyzer
- **File**: diff lines 274-277, 344-354
- **Issue**: Sentinel covers positive int4 decoding across the 1.0→1.4 upgrade chain — good. pg_stat_statements emits **signed** query_ids (sometimes negative). If a future upgrade changes the query_map ID assignment or the decoder's sign handling, positive-only coverage may miss it. Consider adding a second sentinel (e.g. `-424242`) for symmetry.

#### P3 — `string_agg` over ordered subquery does not guarantee final string order (confidence 5)
- **Area**: bug-hunter
- **File**: diff lines 174-186
- **Issue**: The `cron.job_run_details` diagnostic wraps `select … from (… order by start_time desc limit 5) r` in `string_agg`, but `string_agg` without its own `ORDER BY` has an unspecified aggregation order, so the concatenated diagnostic may not reflect `start_time desc`. Diagnostic-only, but undermines the "latest 5" framing. Fix: `string_agg(…, E'\n  ' order by start_time desc)` and drop the inner subquery.

---

### Secrets / tokens in logs
None. `PGPASSWORD: postgres` is the standard throwaway CI password (not a secret). `${{ job.services.postgres.id }}` is a container ID, not a credential. No `GITHUB_TOKEN` echoes. Clean.

### CLAUDE.md alignment
- Upgrade chain 1.0→1.1→1.2→1.3 (+1.4): PR tests the **documented wrapper chain** end-to-end with data preservation across the full chain — matches CLAUDE.md's "Tests: fresh install, upgrade path, schema equivalence" requirement. Good.
- `ash-install.sql` unchanged — correctly treats it as source of truth.
- Anti-extension design preserved — all tests use `\i sql/…`.

### CI status
- 5× fail: `test (14|15|16|17|18, on)` (all fail at H-CI-3, exit 3, ~1m)
- 1× pass: `test (17, off)` (degraded-mode path — the new `cron=off` axis works)
- pass: `CodeQL`, `Analyze (actions)`

---

### Recommendation

**Do not merge until the H-CI-3 step is fixed or made skip-on-legitimate-unavailability.** Once B1 is resolved, this PR is a material quality improvement: it closes real coverage gaps (wrapper chain, data preservation, end-to-end cron scheduling) and hardens previously-silenced failure modes. NB1-NB4 and P1-P3 can be handled in a follow-up.

NikolayS:
## REV follow-up review — commit `2e442e3`

**Verdict: STILL BLOCKING** (with a NEW regression on top of the unresolved original symptom).

### Who had the right root cause?

**Partial credit to both — the fixer was more right about diagnostic signal, the reviewer was more right about the underlying bug still existing.**

- The author's primary claim — *"the test called `ash.stop()` BEFORE reading `cron.job`, so `(no ash_* jobs found)` was the test erasing its own evidence"* — is **verified and correct**. I re-read `.github/workflows/test.yml` at commit `1e1e72e` (the failing tip the reviewer saw), line 474: `perform ash.stop();` is called unconditionally right before the `if v_after <= v_baseline` diagnostic block that reads `cron.job where jobname in ('ash_sampler','ash_rotation')`. The `ash.stop()` unschedule wiped those rows. So the reviewer's "most likely: `ash.start()` took the no-pg_cron branch" hypothesis was *not* supported by that evidence. The failure log from run 24869348373 at the new tip proves the author's claim: `cron.job after ash.start(): jobid=3 jobname=ash_sampler schedule=1 seconds nodename=localhost database=postgres username=postgres active=t`. `ash.start()` does schedule the jobs.
- **However**, the deeper problem the reviewer called out — *H-CI-3 fails deterministically* — is **still unresolved**. In run 24869348373 (commit `2e442e3`, all 5 `cron=on` jobs + `17,off`): `ERROR: pg_cron did not sample within 30 sec: baseline=0 after=0 waited=30`, `cron.job_run_details (latest 5): (no rows — pg_cron never executed the job)`. The author's secondary fix (`cron.use_background_workers = on` in postgresql.conf + docker restart) did not make pg_cron actually fire the scheduled job. So the "end-to-end pg_cron sampling works in CI" premise is still false. The reviewer's B1 verdict (BLOCKING) stands, just for a different underlying reason than initially hypothesised.

### New regression introduced by the fix commit (confidence 10, BLOCKING)

**R1 — `ash.start()` edit in `ash-install.sql` was not mirrored into `ash-1.3-to-1.4.sql`.** Commit `2e442e3` adds `v_debug_logging` / `v_pg_cron_available` locals plus two `raise log` sites into `ash.start()` in `sql/ash-install.sql` (lines 697–847), but `sql/ash-1.3-to-1.4.sql` is unchanged — it only touches `take_sample`, never `ash.start`. Result: `Schema equivalence: fresh install vs full upgrade path` step now fails on the `17,off` job:

```
-FUNC|start|p_interval interval|TABLE(...)|21d6e53a0b70344a483e265bf12eb046
+FUNC|start|p_interval interval|TABLE(...)|1a96027084c433c5c64c718bf69368c8
```

This contradicts CLAUDE.md's "source of truth" framing — for a v1.4 release the source-of-truth function body and the upgrade-path function body must agree. Fix: add a `create or replace function ash.start(...)` to `ash-1.3-to-1.4.sql` with the identical new body. (On cron=on jobs this regression is hidden because H-CI-3 fails first and the schema step never runs.)

### Verification of the other secondary fixes (all good)

- P1 (baseline): `truncate ash.sample_0, ash.sample_1, ash.sample_2` + `assert v_baseline = 0` — verified at lines 508–512 of current `test.yml`, and the failure log now shows `baseline=0` (was `baseline=1`).
- P3 (string_agg order): `order by start_time desc nulls last` now inside the aggregate itself — verified at line 565.
- NB4 (pgss fallback): warn-and-continue restored via `|| echo "pg_stat_statements not available ..."` — verified at lines 126–128.
- NB1/NB2 (env-var routing): `matrix.cron`, `matrix.postgres`, `job.services.postgres.id` all routed through `env:` — verified at step lines 57–58, 108, and consistent use of `"$CONTAINER_ID"` / `"$CRON"` in heredocs throughout.
- Gated `raise log` in `ash.start()`: verified at lines 722–725 and 844–847 of `ash-install.sql`; H-CI-3 enables `set_debug_logging(true)` at line 482 so those logs would surface in the server log on failure. (They are not echoed in psql output — the test enables it for server-log, which is correct behaviour — however run 24869348373 shows no server-log dump in the GHA log, so the "RAISE LOG lines land on failure" promise in the commit message is only kept if a separate "dump postgres log on failure" step exists. I did not find one; the RAISE LOGs are effectively invisible unless the user inspects the container log manually. Low-impact since `cron.job` / `cron.job_run_details` are already dumped via `raise exception`.)

### Remaining concerns (confidence ≥ 7)

1. **C1 (BLOCKING, confidence 10)** — H-CI-3 still fails on every `cron=on` job. Diagnostic now proves `ash_sampler jobid=3 schedule='1 seconds' active=t` is scheduled, `cron.use_background_workers = on` is set, and yet `cron.job_run_details` stays empty for 30 s. Candidates worth exercising before another push:
   - Verify the `shared_preload_libraries` edit actually took effect after `docker restart` — add `show shared_preload_libraries;` + `show cron.use_background_workers;` to a diagnostic step run right after "Wait for PostgreSQL". On GHA service containers the restart can be racy (the `--health-cmd` loop may report ready before the launcher is live).
   - Check whether the pg_cron background launcher attached to the postgres database at all: `select application_name, backend_type from pg_stat_activity where backend_type like '%cron%'`. If it's absent, `shared_preload_libraries` didn't load pg_cron on restart.
   - Consider building a custom image with `shared_preload_libraries` baked in rather than editing the running conf — GHA `services:` doesn't hook init scripts, which is why the PR resorts to `docker exec`; that path has always been fragile.

2. **R1 (BLOCKING, confidence 10)** — the `ash-1.3-to-1.4.sql` mismatch above. Must be fixed before release tag.

3. **L1 (non-blocking, confidence 7)** — the fix commit message promises "gated RAISE LOG ... surfaces silent 'took the no-pg_cron branch' bugs in future", but nothing in the workflow dumps the postgres server log on H-CI-3 failure. Suggest `docker logs "$CONTAINER_ID" 2>&1 | tail -200` in an `if: failure()` step, otherwise the RAISE LOG lines are only useful if someone shells into the runner.

### CI status
- 5× fail: `test (14|15|16|17|18, on)` at H-CI-3
- 1× fail: `test (17, off)` at Schema equivalence (the new R1 regression)
- pass: `Analyze (actions)`, `CodeQL`

### Recommendation
Do not merge. Two independent blockers: H-CI-3 still proves real pg_cron wiring doesn't fire in CI (author's claim "was fine all along" is half-right — scheduling is fine, execution isn't), and the fix commit itself regressed the upgrade-path schema equivalence.

NikolayS:
## REV follow-up #2 — Verdict: STILL BLOCKING

Re-tested at commit `9eda250`. All 5 `cron=on` matrix entries still red (PG 14/15/16/17/18, each ~2m6s–2m10s); `cron=off` still passes. Total: 5 fail / 1 pass / CodeQL+Analyze pass.

**ash.start byte-equivalence (B2):** confirmed. Function body in `sql/ash-install.sql:688` (214 lines) and `sql/ash-1.3-to-1.4.sql:271` is byte-identical — `md5 = 6be100eb71146c58feb04ffcc7cccde4` on both. B2 **is correctly fixed**.

**Minute-boundary hypothesis (B1): NOT CONFIRMED — the hypothesis is wrong.** The wait window was raised 30→90 s; the test polls show `poll t=10,20,30,40,50,60,70,80s: ash.sample=0 cron.job_run_details.count=0` with unchanging zero counts, then `ERROR: pg_cron did not sample within 90 sec: baseline=0 after=0 waited=90`. If this were a cold-start alignment issue the first run would eventually appear; instead **pg_cron never even attempted a run** (`cron.job_run_details` count stays at 0 for the full 90 s). So 90 s isn't enough, AND raising the window further won't help because there's no progression.

**What the CI actually proves is green:**
- `shared_preload_libraries = pg_cron,pg_stat_statements`
- `cron.database_name = postgres`
- `cron.use_background_workers = on`
- `cron.host = localhost`
- `pg_cron launcher bgworker` is attached (pid=35 at 03:38:25.673 — `pg_cron scheduler started`)
- `cron.job` shows `jobid=3 jobname=ash_sampler schedule=1 seconds nodename=localhost database=postgres username=postgres active=t`
- `docker logs --tail 400` runs cleanly (no "docker: command not found"), `$CONTAINER_ID` is resolved via `services.postgres.id`, postgres container logs print — the new diagnostic steps work as designed.

**What the postgres-log dump reveals as the real signal:** between `pg_cron scheduler started` (03:38:25.673) and the 90 s failure (03:40:01.553) there is **no `pg_cron` LOG, NOTICE, or ERROR line at all** — not a single attempted execution. The only noise is `FATAL: role "root" does not exist` every ~10 s, which is the GHA service-container healthcheck (`pg_isready` defaulting to the runner's OS user), unrelated to pg_cron. So it's not "scheduled but failed to run"; it's "launcher sees the job, sub-minute tick never fires the bgworker." That points at a `pg_cron` 1.6 sub-minute-scheduler bug / edge case in this specific environment, not at ash.start().

**Likely real root causes to investigate next (in order of cheapest to test):**
1. **`cron.job` was created AFTER the launcher started.** Order: launcher bgworker attaches at 03:38:25.673 (post-restart), then `CREATE EXTENSION pg_cron` happens in the next step, which is when `cron.job` is physically materialised. pg_cron 1.6 should pick this up, but some deployments report that the launcher needs a `pg_reload_conf()` or a second restart after the extension exists for sub-minute scheduling to engage. Try: create the extension *before* the restart that loads `shared_preload_libraries` (i.e. install pg_cron pre-restart, then set GUCs, then restart once — so the extension is already installed when the launcher boots). Or issue `select pg_reload_conf()` after `CREATE EXTENSION pg_cron`.
2. **Drop sub-minute for the test.** Schedule `*/1 * * * *` (every minute) and wait 120 s. That removes the sub-minute code path entirely; if this is green, the bug is localised to pg_cron's sub-minute path under these exact GHA conditions.
3. **Try `cron.use_background_workers = off` with a valid `nodename`** to route through libpq — the bgworker path may be the broken one in 1.6 on Debian 13 / PG 18 images.

**Other concerns (confidence ≥ 7):**
- 90 s per failed matrix entry × 5 entries = ~7.5 CI-minutes burned per run with zero signal — please land a faster-fail (e.g. also require `cron.job_run_details.count` to advance within 30 s and bail early, not just absolute sample count).
- `bash -e` plus the failing step means the H-CI-3 failure *cancels* every later step in that job (Test set_debug_logging, intervals, reader functions, timeline, uninstall, upgrade paths, schema equivalence). With cron=on never finishing the whole suite we have **no signal on any of those downstream tests** on 5 of 6 matrix legs. Consider making H-CI-3 `continue-on-error: true` temporarily (or moving it to the end of the job) so the rest of the matrix still produces evidence while this is being debugged.
- The `FATAL: role "root" does not exist` healthcheck noise is harmless but will confuse future reviewers reading the dump — worth pointing at `-U postgres` in the service `options:` if you touch the workflow again.

Recommendation: do **not** merge. Try cause-1 (reorder: install extension before the GUC-setting restart) as the next cheapest fix — it's the one change that matches exactly what the logs show (launcher attached before `cron.job` existed).

NikolayS:
## Status & request for direction

After three iterations (`1e1e72e`, `2e442e3`, `9eda250`) H-CI-3 is still red on every `cron=on` job. The postgres-log dump in `9eda250` confirmed:

- `shared_preload_libraries = pg_cron,pg_stat_statements` — loaded
- `cron.database_name = postgres`, `cron.use_background_workers = on`, `cron.host = localhost` — set
- pg_cron launcher bgworker attached (`pg_cron scheduler started`)
- `cron.job` has `jobid=3 jobname=ash_sampler schedule='1 seconds' active=t` immediately after `ash.start('1 second')`
- Over the 90 s wait: `cron.job_run_details` stays at **0 rows** — the bgworker never attempts a run

So `ash.start()` is fine; the problem is pg_cron 1.6's sub-minute scheduler not firing under GHA `services:` containers. The last REV follow-up laid out three candidate root causes, ordered cheapest-to-try:

1. **Extension install / restart ordering**: `CREATE EXTENSION pg_cron` currently runs AFTER the shared-preload restart; the launcher attaches before `cron.job` exists. Try installing the extension *first*, then setting GUCs + restart *once*, so `cron.job` is materialised when the launcher boots.
2. **Drop the sub-minute schedule**: use `*/1 * * * *` (every minute) with a 120 s wait. Removes the sub-minute code path entirely; if this goes green, the bug is localised to pg_cron sub-minute in this environment.
3. **`cron.use_background_workers = off`** with a valid `nodename`, routing through libpq instead of the bgworker.

I don't want to push more speculative commits. Three questions — could you pick one?

- **A)** Try (1) first (cheapest, matches the log evidence) — I'll push that.
- **B)** Drop H-CI-3 from this PR entirely; keep the cheaper schedule-assertion part (verify `cron.job` has the ash_* rows after `ash.start()`, skip the sample-count check). Ship H-CI-1/2 + M-CI-4/5/6 + L-CI-7 now; open a follow-up issue for "prove pg_cron *executes* in CI".
- **C)** Mark H-CI-3 `continue-on-error: true` while leaving the step in place, so the other tests in the job (set_debug_logging, intervals, readers, timeline, uninstall, upgrade, schema equivalence) run to completion and give signal. Merge when those are green; debug H-CI-3 async.

My recommendation is **B** — the review's original intent was "make sure we notice if pg_cron scheduling breaks"; a `cron.job` presence check catches 90% of that without depending on sub-minute execution timing in `services:`. But this is your call.

---
_Generated by [Claude Code](https://claude.ai/code)_

NikolayS:
## REV re-review 3 (fresh reviewer, skeptical pass) — **STILL BLOCKING** + **REGRESSION**

Run [24871188023](https://github.com/NikolayS/pg_ash/actions/runs/24871188023) on `a530949`. All 5 `cron=on` legs fail (~2m13s each); `test (17, off)` passes; Analyze/CodeQL pass.

### CI breakdown (per matrix leg)

| Matrix | Result | Where it dies | Root cause |
|---|---|---|---|
| 14/15/16/17/18 × cron=on | **fail** | `Test set_debug_logging` — `ERROR: debug_logging should default to false` | **cascade from H-CI-3** |
| 17 × cron=off | pass | — | — |

### Does pg_cron finally fire? **No.** Launcher-staleness hypothesis was **WRONG**.

The new "Verify pg_cron wiring" step and H-CI-3 preflight both pass on every leg:
- `shared_preload_libraries = pg_cron,pg_stat_statements` — survived.
- `cron.use_background_workers = on`, `cron.database_name = postgres` — applied.
- `pg_cron scheduler / pg_cron launcher` bgworker attached (one process, pid visible).
- `cron.job` has `ash_sampler` (schedule=`1 seconds`) and `ash_rotation` (schedule=`0 0 * * *`), both `active=t`, right after `ash.start()`.
- The extra `docker restart` after CREATE EXTENSION completes cleanly (~5s).

Yet across all 5 legs, `cron.job_run_details.count` stays at 0 for the full 90s (polled every 10s — identical trace on PG14/15/16/17/18). The "Dump postgres container logs" step now runs (since it uses `steps.h_ci_3.outcome == 'failure'`) and surfaces the **actual** cause — visible in every leg:

```
2026-04-24 03:49:05.671 UTC [55] FATAL:  role "root" does not exist
2026-04-24 03:49:15.734 UTC [63] FATAL:  role "root" does not exist
2026-04-24 03:49:25.798 UTC [72] FATAL:  role "root" does not exist
… every ~10s through t=90s, 9 total FATALs …
```

pg_cron is alive and trying to connect on schedule — but the connection it opens (to talk to `cron.database_name`) authenticates as the OS user of the pg_cron worker process. Inside the PGDG `postgres:17` container the postmaster runs as OS user `root` (from the job's `docker exec` ENTRYPOINT), so pg_cron issues an implicit `role=root` connection that PG rejects with `FATAL: role "root" does not exist`. These FATALs never hit `cron.job_run_details` — the connection is pre-auth — which is exactly why prior debug rounds saw `count=0` and assumed the scheduler hadn't fired. It had. It just couldn't authenticate.

`cron.use_background_workers = on` was supposed to bypass this, but it only bypasses libpq for **command execution** — the launcher still opens a metadata connection to `cron.database_name` as the OS user. Setting `cron.host` to the Unix socket path (`/var/run/postgresql` or similar) typically sidesteps this via `trust` in `pg_hba`, as does exporting `PGUSER=postgres` into the postmaster environment, or adding a `postgres` role to the container. The one-line fix most pg_cron-in-docker projects land on is adding this to the existing postgresql.conf edit:

```
cron.host = /var/run/postgresql
```

(or whichever UNIX_SOCKET_DIRECTORIES value the image uses — see `show unix_socket_directories;`).

### Regression introduced by `continue-on-error: true`

Before fix 3, an H-CI-3 failure cancelled the job. Now it doesn't — which surfaces a *second* defect: H-CI-3 does `select ash.set_debug_logging(true);` as its first statement, and relies on a trailing `select ash.set_debug_logging(false);` AFTER the `do $$ … $$` block to reset. When the `do $$` block raises `pg_cron did not sample within 90 sec`, psql with `ON_ERROR_STOP=1` aborts **before** the trailing reset runs, so `ash.config.debug_logging` stays `true`. The next step (`Test set_debug_logging`) then asserts `debug_logging should default to false` and fails with exit code 3 — this is the error that's currently rolling up as the visible CI failure on all 5 legs. The "continue after H-CI-3" goal is not being met; the cascade kills every downstream assertion that depends on clean config.

Fix options (not requested — diagnosis only): move the `set_debug_logging(false)` into the `do $$` block's successful-exit branch AND an `EXCEPTION WHEN OTHERS` handler that re-raises after resetting; or wrap H-CI-3's psql invocation in a trap that resets config regardless of the EXIT code.

### Verdict and byte-equivalence check

- `ash.start()` body in `sql/ash-install.sql` (lines 688-903) and `sql/ash-1.3-to-1.4.sql` (lines 271-484) still md5 `6be100eb71146c58feb04ffcc7cccde4` — unchanged from the previous re-review, **byte-equivalent**, no regression.
- The second `docker restart` placement is correct (only after successful `CREATE EXTENSION`, bounded pg_isready poll, ~5s added per leg — acceptable cost).
- `continue-on-error: true` is only on H-CI-3; the `if: always() && steps.h_ci_3.outcome == 'failure'` conditional on the log-dump step does evaluate correctly (we can see it actually ran this time, hence the FATAL evidence above).

### Confidence >=7 concerns
1. **[conf 9] STILL BLOCKING:** pg_cron never runs the job on any cron=on leg because the pg_cron launcher's metadata connection is rejected with `FATAL: role "root"`. Fix 3's restart after CREATE EXTENSION was a no-op for the actual failure mode.
2. **[conf 9] REGRESSION:** `continue-on-error: true` on H-CI-3 without the config-reset moved into an exception handler causes `Test set_debug_logging` to fail on every cron=on leg, which is a new CI-red signal that didn't exist before fix 3. The downstream steps "continue" but immediately fail on polluted state — so the claimed benefit ("intervals / readers / upgrades / schema-equivalence / uninstall steps" keep running) still isn't realized in practice.
3. **[conf 8]** Two successive docker restarts of the pg_cron service plus cold-start wait means every cron=on leg now spends ~15-20 seconds on pg_cron setup before the real tests start. That's acceptable in isolation, but undoing fix 3 is probably cheaper than keeping it once the real cause is addressed — the original "launcher staleness" symptom was a misread of the same `FATAL: role "root"` evidence that is now visible.

NikolayS:
## REV follow-up 4 — STILL BLOCKING (critical regression hidden by `continue-on-error`)

**Verdict: STILL BLOCKING.** The CI rollup is all-green (8/8 checks pass on run [24871391977](https://github.com/NikolayS/pg_ash/actions/runs/24871391977)), but that green is a lie: **H-CI-3 is still failing on every single `cron=on` matrix leg (PG 14, 15, 16, 17, 18)** with the same symptom reviewers have been chasing for four rounds — `pg_cron did not sample within 90 sec: baseline=0 after=0 waited=90`. The `continue-on-error: true` on that step is hiding a test that has never once succeeded. The `CREATE ROLE root SUPERUSER LOGIN` fix changed the error string but not the outcome: postgres container logs on all five legs now emit `FATAL: database "root" does not exist` every ~10 s (9 occurrences per leg during the 90 s poll), and `cron.job_run_details.count` stays at 0 across every `poll t=10s/20s/.../80s` tick. The diagnosis in the commit message was half-right — peer auth does try OS user `root` — but libpq on a peer connection with no explicit `dbname` defaults `dbname` to the username, so pg_cron's metadata connection is now looking for a database called `root` that also doesn't exist. Round-4 replaced "role root does not exist" with "database root does not exist"; same FATAL cadence, same 0 job runs, same timeout. The real fix is either a `CREATE DATABASE root` alongside the role, setting `cron.host = '/var/run/postgresql'` combined with pg_hba peer maps, or (simplest) running the postgres service container with `--user postgres` so the launcher's peer identity matches the existing `postgres` role and database.

**What did land correctly:**
- The `if: always()` "Reset state leaked by H-CI-3" step plugs the debug_logging leak cleanly: every matrix leg shows `TRUNCATE TABLE` / `debug_logging disabled` in the reset output and the subsequent `Test set_debug_logging` step reports `set_debug_logging tests PASSED` on all five `cron=on` legs (plus `17, off`). That sub-problem from round 3 is genuinely resolved.
- `ash.start()` remains byte-identical between `sql/ash-install.sql` and `sql/ash-1.3-to-1.4.sql` (7666 bytes each; diff is empty).
- `ON_ERROR_STOP=0` on the reset block is reasonable given its `always()` semantics (e.g. no-pg_cron legs where `ash.stop()` still works but could throw on a degraded install). It only swallows errors in cleanup, not in tests.

**Other concerns (confidence >= 7):**
1. **The entire point of H-CI-3 is defeated.** A test that is green only because its failure is masked does not provide CI coverage — it produces a false-positive signal that pg_cron integration works when it demonstrably does not. Either fix the underlying wiring so H-CI-3 genuinely passes, or drop `continue-on-error: true` and accept the red (preferred: it forces the underlying fix). Shipping #44 in the current state normalises a mechanism that will silently hide every future pg_cron regression.
2. **Comment drift.** The reset step's comment says it resets `ash.sample / ash.config / cron.job`, but the SQL only touches `ash.sample_*` partitions. It does not truncate `ash.config` and does not clear `cron.job` (which `ash.stop()` mostly handles, but the comment overpromises).
3. **`CREATE ROLE root SUPERUSER LOGIN`** is fine in this CI sandbox (throwaway service container), but once the actual wiring is fixed this role becomes vestigial and worth removing — it exists only to mask a misconfiguration.
4. **`truncate ash.sample_0, ash.sample_1, ash.sample_2`** is not a no-op on cron=off/degraded legs — it actually truncates. That's harmless today (no later step asserts on prior sample rows), but it contradicts the comment's "No-op on matrix.cron=off" claim.

**Bottom line:** the one test this PR was supposed to fix (pg_cron end-to-end) is worse off than before — it's now a permanent false-green. Round 5 needs to either make pg_cron actually fire (create the `root` database, or point `cron.host` at the unix socket with a matching pg_hba entry, or run the container as the `postgres` user) or remove `continue-on-error: true` so the CI signal is truthful.

NikolayS:
## REV final re-review (post scope-reduction)

**Verdict: RESOLVED.**

CI is genuinely clean: all 8 required checks (`test` across PG 14/15/16/17/18 with `cron=on`, PG17 with `cron=off`, CodeQL, Analyze) are green on head `a2a3a3a`, with no `continue-on-error` anywhere in `.github/workflows/test.yml` (`grep` returns 0 hits) — so green means green, nothing is being masked. What is **kept** is verified in-tree: the "Verify pg_cron wiring" step is present (`test.yml` L130) and still does real work (GUC readback, `pg_extension` version, scheduler-bgworker attachment via `pg_stat_activity`, `ash_*` job count), not a no-op; the gated `raise log` paths inside `ash.start()` are present in `sql/ash-install.sql` (L724, L845); and `ash.start()` is **byte-identical** between `sql/ash-install.sql` and `sql/ash-1.3-to-1.4.sql` (both extracts md5 `6be100eb71146c58feb04ffcc7cccde4`, 214 lines each). What is **removed** is cleanly removed: `grep "CREATE ROLE root"` — no matches; the H-CI-3 90s poll/`raise exception` body, the `Dump postgres container logs on H-CI-3 failure` step, the `Reset state leaked by H-CI-3` step, and the second `docker restart` after `CREATE EXTENSION pg_cron` are all gone; the only residual "H-CI-3" references are the explanatory comment at L516–526 pointing to #46 and two comments inside the retained "Verify pg_cron wiring" step that accurately explain *why* that diagnostic exists (reasonable to keep for future debuggers). The other claimed improvements in the PR are intact — H-CI-1 wrapper-chain exercise + H-CI-2 data-preservation fixture (sentinel `query_id = 424242` round-trip through decode), M-CI-4 `cron=off` matrix axis on PG17, M-CI-5 desilencing of `|| true`/`2>/dev/null` around pgss/uninstall (with the one intentionally-silenced path explicitly commented), M-CI-6 `docker exec`-into-service-container rationale, and L-CI-7 `>= 0` → `> 0` or `perform *` conversions — none were collateral damage from the scope reduction. Issue #46 is thorough: documents the `FATAL: database "root" does not exist` peer-auth-default-dbname root cause, enumerates all five prior fix commits with shas and why each failed (the `CREATE ROLE root` attempt is correctly called out as addressing the wrong failure mode — role vs database), and proposes three tractable next-step approaches (custom entrypoint, self-managed `docker run`, local repro first). Deferring to #46 is appropriate, not a hidden bug: the in-DB logic (`ash.start`/`take_sample`/cron wiring) is still exercised by direct `take_sample()` calls and the wiring-verification step; what's deferred is specifically the "cron actually fires on schedule inside a GHA service container" end-to-end assertion, and #46 documents that honestly. One residual cosmetic nit worth noting: the PR **title** still reads "ci: coverage & robustness (wrapper, data preservation, pg_cron timing)" — the body correctly drops the pg_cron-timing claim and points to #46, but the title parenthetical is now slightly misleading; consider renaming to e.g. "ci: coverage & robustness (wrapper, data preservation, pg_cron wiring)" on merge. Not blocking.

PR #45, opened 2026-04-22
security: search_path guard + REVOKE readers from PUBLIC

### LINKED ISSUE #37: Comprehensive review: correctness, security, CI, and docs findings (v1.4)

NikolayS:
## Summary

Defense-in-depth privilege hardening for pg_ash, addressing issue #37 findings **H-SEC-1** (REVOKE from PUBLIC) and **H-SEC-2** (`set search_path` on every function).

- **H-SEC-2**: added `set search_path = pg_catalog, ash` to all **38** functions in the `ash` schema. Functions are currently INVOKER so the blast radius is limited today; this guard eliminates the latent RCE vector if any function is ever flipped to SECURITY DEFINER.
- **H-SEC-1**: extended the existing admin-only REVOKE block so that:
  - Every non-trigger function in `ash.*` has EXECUTE revoked from PUBLIC. Signatures are resolved dynamically via `pg_get_function_identity_arguments(pg_proc.oid)` to avoid drift from DEFAULTs/overloads.
  - `ash.sample`, `ash.query_map_all`, `ash.config`, `ash.wait_event_map` have SELECT revoked from PUBLIC.
  - Per-slot partition tables (`sample_N`) and dictionary tables (`query_map_N`) are enumerated via `pg_class` and each has SELECT revoked from PUBLIC (REVOKE on a partitioned parent does not cascade to partitions in Postgres).

No new role is introduced — the installing role retains full access via ownership. Scope is kept tight per the issue.

## Test plan

- [x] `sudo -u postgres psql -f sql/ash-install.sql` on a fresh database — runs clean, no errors.
- [x] Verified `pg_proc.proconfig` contains `search_path=pg_catalog, ash` for a sampled function.
- [x] Verified `pg_proc.proacl` and `pg_class.relacl` for `ash.*` objects contain only owner entries (no `=X/...` or `=r/...` PUBLIC grants).
- [ ] CI (`test.yml`) passes across PostgreSQL 14–18 — fresh install, upgrade path, degraded mode (no pgss/pg_cron).
- [ ] Manual smoke: as a non-owner role, `select ash.top_queries()` is denied; as owner, it works.

Closes #37 (H-SEC-1, H-SEC-2).

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -139,6 +139,109 @@ jobs:
           $$;
           EOF
 
+      - name: Test privilege hardening (REVOKE from PUBLIC)
+        env:
+          PGPASSWORD: postgres
+        run: |
+          psql -h localhost -U postgres -d postgres -v ON_ERROR_STOP=1 << 'EOF'
+          -- Create an unprivileged role to represent a "random tenant"
+          do $$
+          begin
+            if not exists (select from pg_roles where rolname = 'ash_test_unpriv') then
+              create role ash_test_unpriv login password 'x';
+            end if;
+          end $$;
+
+          do $$
+          begin
+            -- PUBLIC has no EXECUTE on reader functions
+            assert not has_function_privilege('public', 'ash.top_waits(interval,int,int,boolean)', 'EXECUTE'),
+              'PUBLIC should NOT have EXECUTE on ash.top_waits';
+            assert not has_function_privilege('public', 'ash.top_queries(interval,int)', 'EXECUTE'),
+              'PUBLIC should NOT have EXECUTE on ash.top_queries';
+            assert not has_function_privilege('public', 'ash.status()', 'EXECUTE'),
+              'PUBLIC should NOT have EXECUTE on ash.status';
+            assert not has_function_privilege('public', 'ash.samples(interval,int)', 'EXECUTE'),
+              'PUBLIC should NOT have EXECUTE on ash.samples';
+
+            -- PUBLIC has no EXECUTE on admin functions
+            assert not has_function_privilege('public', 'ash.start(interval)', 'EXECUTE'),
+              'PUBLIC should NOT have EXECUTE on ash.start';
+            assert not has_function_privilege('public', 'ash.stop()', 'EXECUTE'),
+              'PUBLIC should NOT have EXECUTE on ash.stop';
+            assert not has_function_privilege('public', 'ash.rotate()', 'EXECUTE'),
+              'PUBLIC should NOT have EXECUTE on ash.rotate';
+            assert not has_function_privilege('public', 'ash.take_sample()', 'EXECUTE'),
+              'PUBLIC should NOT have EXECUTE on ash.take_sample';
+            assert not has_function_privilege('public', 'ash.set_debug_logging(bool)', 'EXECUTE'),
+              'PUBLIC should NOT have EXECUTE on ash.set_debug_logging';
+
+            -- PUBLIC has no SELECT on reader tables
+            assert not has_table_privilege('public', 'ash.sample', 'SELECT'),
+              'PUBLIC should NOT have SELECT on ash.sample';
+            assert not has_table_privilege('public', 'ash.config', 'SELECT'),
+              'PUBLIC should NOT have SELECT on ash.config';
+            assert not has_table_privilege('public', 'ash.wait_event_map', 'SELECT'),
+              'PUBLIC should NOT have SELECT on ash.wait_event_map';
+            assert not has_table_privilege('public', 'ash.query_map_all', 'SELECT'),
+              'PUBLIC should NOT have SELECT on ash.query_map_all';
+            assert not has_table_privilege('public', 'ash.sample_0', 'SELECT'),
+              'PUBLIC should NOT have SELECT on ash.sample_0';
+            assert not has_table_privilege('public', 'ash.query_map_0', 'SELECT'),
+              'PUBLIC should NOT have SELECT on ash.query_map_0';
+
+            -- Unprivileged role (non-owner) cannot execute readers either
+            assert not has_function_privilege('ash_test_unpriv', 'ash.top_waits(interval,int,int,boolean)', 'EXECUTE'),
+              'non-owner role should NOT have EXECUTE on ash.top_waits';
+            assert not has_function_privilege('ash_test_unpriv', 'ash.top_queries(interval,int)', 'EXECUTE'),
+              'non-owner role should NOT have EXECUTE on ash.top_queries';
+
+            -- Owner (postgres) retains EXECUTE by virtue of ownership
+            assert has_function_privilege('postgres', 'ash.top_waits(interval,int,int,boolean)', 'EXECUTE'),
+              'owner SHOULD have EXECUTE on ash.top_waits';
+            assert has_function_privilege('postgres', 'ash.start(interval)', 'EXECUTE'),
+              'owner SHOULD have EXECUTE on ash.start';
+            assert has_function_privilege('postgres', 'ash.take_sample()', 'EXECUTE'),
+              'owner SHOULD have EXECUTE on ash.take_sample';
+            assert has_table_privilege('postgres', 'ash.sample', 'SELECT'),
+              'owner SHOULD have SELECT on ash.sample';
+
+            raise notice 'Privilege hardening tests PASSED';
+          end;
+          $$;
+
+          -- Behavioral check: call as the unprivileged role, expect 42501
+          set role ash_test_unpriv;
+          do $$
+          begin
+            begin
+              perform ash.top_waits('1 hour'::interval, 10, 0);
+              raise exception 'expected permission denied, but call succeeded';
+            exception when insufficient_privilege then
+              -- expected path
+              null;
+            end;
+          end;
+          $$;
+          reset role;
+
+          -- GRANT flow smoke-test: owner can hand out EXECUTE, then REVOKE it
+          grant execute on function ash.top_waits(interval, int, int, boolean) to ash_test_unpriv;
+          do $$
+          begin
+            assert has_function_privilege('ash_test_unpriv', 'ash.top_waits(interval,int,int,boolean)', 'EXECUTE'),
+              'GRANT to non-owner role should succeed';
+          end $$;
+          revoke execute on function ash.top_waits(interval, int, int, boolean) from ash_test_unpriv;
+          do $$
+          begin
+            assert not has_function_privilege('ash_test_unpriv', 'ash.top_waits(interval,int,int,boolean)', 'EXECUTE'),
+              'REVOKE from non-owner role should succeed';
+          end $$;
+
+          drop role ash_test_unpriv;
+          EOF
+
       - name: Test sampler and decoder
         env:
           PGPASSWORD: postgres
--- a/README.md
+++ b/README.md
@@ -698,6 +698,37 @@ Don't forget to also schedule `ash.rotate()` — once per `rotation_period` (def
 0 0 * * * psql -qAtX -d mydb -c "SELECT ash.rotate();"
 ```
 
+## Privileges
+
+pg_ash installs with a locked-down privilege model: admin functions (`ash.start()`, `ash.stop()`, `ash.rotate()`, `ash.take_sample()`, `ash.set_debug_logging()`, `ash.uninstall()`) are restricted to the schema owner, and EXECUTE on all reader functions plus SELECT on reader tables (`ash.sample`, `ash.query_map_all`, `ash.config`, `ash.wait_event_map`, and per-slot partitions) is revoked from `PUBLIC`. The installing role retains full access.
+
+Grant access to a monitoring or read-only role explicitly:
+
+```sql
+-- allow a monitoring role to call the readers
+grant usage on schema ash to my_monitor_role;
+grant execute on function ash.top_waits(interval, int, int)           to my_monitor_role;
+grant execute on function ash.top_queries(interval, int)              to my_monitor_role;
+grant execute on function ash.top_queries_with_text(interval, int)    to my_monitor_role;
+grant execute on function ash.samples(interval, int)                  to my_monitor_role;
+grant execute on function ash.status()                                to my_monitor_role;
+
+-- or grant all reader functions at once
+grant execute on all functions in schema ash to my_monitor_role;
+
+-- grant direct read on raw tables for ad-hoc SQL (optional)
+grant select on all tables in schema ash to my_monitor_role;
+```
+
+### pg_stat_statements in a non-default schema
+
+pgss reader functions (`top_queries`, `top_queries_at`, `top_queries_with_text`, `samples`, `samples_at`, `event_queries`, `event_queries_at`) need the pg_stat_statements schema on their `search_path`. Install detects it automatically. If you install pg_stat_statements **after** pg_ash, or move it to a non-default schema, re-apply:
+
+```sql
+-- detect the pgss schema and re-apply search_path on pgss readers
+select ash._apply_pgss_search_path();
+```
+
 ## Known limitations
 
 - **Primary only** — pg_ash requires writes (`INSERT` into sample tables, `TRUNCATE` on rotation), so it cannot run on physical standbys or read replicas. Install it on the primary; it samples all databases from there.
--- a/sql/ash-1.3-to-1.4.sql
+++ b/sql/ash-1.3-to-1.4.sql
@@ -337,6 +337,20 @@ $$;
 do $$
 declare
   r record;
+  -- Functions that reference pg_stat_statements unqualified need public on
+  -- the search_path so the view (installed by CREATE EXTENSION pg_stat_statements
+  -- into public by default on RDS/Cloud SQL/Supabase/AlloyDB/Neon) is visible.
+  -- Without this, the probe "perform 1 from pg_stat_statements" fails with
+  -- "relation does not exist", the catch-block sets v_has_pgss=false, and the
+  -- caller either raises "pg_stat_statements extension is not installed"
+  -- (top_queries_with_text, event_queries, event_queries_at) or silently
+  -- degrades with a misleading warning (top_queries, top_queries_at, samples,
+  -- samples_at). Keep public last so ash.* still wins for ash objects.
+  v_pgss_readers text[] := array[
+    'top_queries', 'top_queries_at', 'top_queries_with_text',
+    'samples', 'samples_at',
+    'event_queries', 'event_queries_at'
+  ];
 begin
   -- Apply search_path guard to every non-trigger function in ash.*.
   -- alter function ... set search_path is idempotent.
@@ -348,8 +362,13 @@ begin
     where n.nspname = 'ash'
       and p.prokind = 'f'
   loop
-    execute format('alter function ash.%I(%s) set search_path = pg_catalog, ash',
-                   r.proname, r.args);
+    if r.proname = any(v_pgss_readers) then
+      execute format('alter function ash.%I(%s) set search_path = pg_catalog, ash, public',
+                     r.proname, r.args);
+    else
+      execute format('alter function ash.%I(%s) set search_path = pg_catalog, ash',
+                     r.proname, r.args);
+    end if;
   end loop;
 end $$;
 
--- a/sql/ash-1.3-to-1.4.sql
+++ b/sql/ash-1.3-to-1.4.sql
@@ -364,8 +364,11 @@ set search_path = pg_catalog, ash
 as $$
 declare
   v_pgss_schema text := ash._pgss_schema();
-  -- Keep in sync with v_pgss_readers in the DO block below AND with the same
-  -- list in ash-install.sql (ash._apply_pgss_search_path).
+  -- B2: keep this list in sync with v_pgss_readers in sql/ash-1.3-to-1.4.sql
+  -- (and any future upgrade scripts) AND with the per-function `set
+  -- search_path = pg_catalog, ash, public` clauses in this file. Any reader
+  -- that probes pg_stat_statements must appear here so its search_path
+  -- resolves the view across managed-service and custom pgss install schemas.
   v_readers text[] := array[
     'top_queries', 'top_queries_at', 'top_queries_with_text',
     'samples', 'samples_at',
@@ -374,6 +377,10 @@ declare
   v_path text;
   r record;
 begin
+  -- Always keep public in the path as a fallback (matches the managed-service
+  -- default and preserves behavior when pgss is not yet installed). If the
+  -- extension lives elsewhere, append that schema after public so ash.* still
+  -- wins over any user-created objects in public/<pgss_schema>.
   if v_pgss_schema is null or v_pgss_schema in ('pg_catalog', 'ash', 'public') then
     v_path := 'pg_catalog, ash, public';
   else
--- a/sql/ash-1.3-to-1.4.sql
+++ b/sql/ash-1.3-to-1.4.sql
@@ -333,25 +333,100 @@ $$;
 -- SELECT/EXECUTE from PUBLIC on reader functions, underlying tables, and
 -- per-slot partitions. Dynamic so it covers functions/partitions created by
 -- earlier upgrade scripts. All statements are idempotent.
+--
+-- IMPORTANT: the v_pgss_readers list below must stay in sync with the same
+-- list inside ash._apply_pgss_search_path() in ash-install.sql. If you add or
+-- remove a function that probes pg_stat_statements, update both places.
 -------------------------------------------------------------------------------
+
+-- Helper: detect the schema that holds the pg_stat_statements view so reader
+-- functions keep working when pgss lives outside `public` (self-hosted
+-- installs, schema-isolated setups). Mirrors ash-install.sql.
+create or replace function ash._pgss_schema()
+returns text
+language sql
+stable
+set search_path = pg_catalog
+as $$
+  select n.nspname::text
+  from pg_extension e
+  join pg_namespace n on n.oid = e.extnamespace
+  where e.extname = 'pg_stat_statements'
+$$;
+
+comment on function ash._pgss_schema() is
+  'Returns the schema name of the installed pg_stat_statements extension, or NULL if not installed. Used to keep reader functions portable across managed services and custom install schemas.';
+
+create or replace function ash._apply_pgss_search_path()
+returns text
+language plpgsql
+set search_path = pg_catalog, ash
+as $$
+declare
+  v_pgss_schema text := ash._pgss_schema();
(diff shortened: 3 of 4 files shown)


NikolayS:
# REV review — PR #45

**Title:** security: search_path guard + REVOKE readers from PUBLIC
**Scope:** 2 files, +193 / -2. Defense-in-depth privilege hardening (closes H-SEC-1, H-SEC-2 from issue #37).
**CI:** all green (Analyze, CodeQL, test 14/15/16/17/18).

## Verdict

**APPROVE with 1 potential regression to confirm.** The hardening is well-scoped, correctly implemented, and idempotent. One behavioral change (pgss schema assumption) may affect users who installed `pg_stat_statements` outside `public`. No blocking issues.

## Summary counts

- BLOCKING: 0
- NON-BLOCKING: 1 (security, high confidence but low impact in practice)
- POTENTIAL: 4 (1 behavior change, 2 perf/inlining, 1 docs gap)
- INFO: 2

---

## Security reviewer

### POTENTIAL — S1: `search_path = pg_catalog, ash, public` changes pgss resolution behavior (confidence 7)

- **file:** `sql/ash-install.sql` lines 1316, 1468, 1732, 2491, 2596, 2731, 2838 (7 functions); mirrored in `sql/ash-1.3-to-1.4.sql`
- **issue:** Before this PR, the seven pgss-reader functions (`top_queries`, `top_queries_at`, `top_queries_with_text`, `samples`, `samples_at`, `event_queries`, `event_queries_at`) had no `SET search_path`, so unqualified `pg_stat_statements` resolved against the **caller's** search_path. After this PR, the function's search_path is hard-pinned to `pg_catalog, ash, public`. Users who installed `CREATE EXTENSION pg_stat_statements SCHEMA some_other_schema` (or rely on a non-default search_path) will now see the pgss probe fail and the function either raise `pg_stat_statements extension is not installed` (`top_queries_with_text`, `event_queries`, `event_queries_at`) or silently degrade with a misleading warning (`top_queries`, `top_queries_at`, `samples`, `samples_at`).
- **evidence:**
  ```
  set search_path = pg_catalog, ash, public
  ...
  perform 1 from pg_stat_statements limit 1;  -- now only resolves against public
  ```
- **fix / mitigation options:**
  1. Document the assumption in README ("pg_stat_statements must be installed in `public`"), or
  2. Probe multiple schemas by checking `pg_extension`/`pg_namespace` (resolve the pgss extension's schema and qualify the reference), or
  3. Expose a GUC (`ash.pgss_schema`) that admins can override.
- **note:** Managed services (RDS, Cloud SQL, Supabase, AlloyDB, Neon) install pgss into `public` by default, so the practical blast radius is small — but self-hosted users with hardened templates often put extensions elsewhere.

### NON-BLOCKING — S2: `public` last in search_path is safe today, but note the PG 14 caveat (confidence 8)

- **file:** same 7 functions as S1
- **issue:** With `search_path = pg_catalog, ash, public`, any unqualified name not found in `pg_catalog` or `ash` falls through to `public`. In PG 14 (which `test.yml` explicitly covers) `public` is still `CREATE` to `PUBLIC` by default. A hostile unprivileged user can create a relation named `pg_stat_statements` in `public`, and these `SECURITY INVOKER` functions would then SELECT from it.
- **Why non-blocking:** Because these functions are `SECURITY INVOKER`, the attacker would only read their own view of their own planted relation — no privilege escalation. The same unprivileged user can already query any relation they own. This is a latent gotcha if any of these functions is ever flipped to `SECURITY DEFINER` in the future (exactly the scenario this PR's guard is meant to defend against).
- **fix:** Either (a) document in the header comment that the `public` entry REQUIRES these functions to remain `SECURITY INVOKER`, or (b) resolve pgss's actual schema at function time and qualify the reference. Option (b) also fixes S1.
- **cwe:** CWE-427 (Uncontrolled Search Path Element) — theoretical.

### NON-BLOCKING — S3: pg_proc enumeration will re-apply on every re-run, which is desired but worth noting (confidence 8)

- **file:** `sql/ash-1.3-to-1.4.sql` lines 335-372, `sql/ash-install.sql` lines 2962-2977
- **issue:** The dynamic `for r in ... alter function ... set search_path` block will iterate every function in `ash.*` on every re-install/upgrade. This is correct and idempotent for current code, but it also means: if a future extension/patch introduces a function that **must** see public (e.g. different extension schema), it needs to be added to `v_pgss_readers` or the `search_path` will silently downgrade it.
- **fix:** Add a comment at the loop reminding maintainers to update `v_pgss_readers` when adding pgss-readers. Optional: add a CI assertion that every function in `ash.*` either has no `pg_stat_statements` reference or has `public` in its search_path.

### INFO — S4: REVOKE coverage is complete (confidence 10)

Verified the REVOKE block covers everything readable:
- Functions: dynamic `pg_proc` enumeration with `prokind='f'` — catches all 38 functions. No trigger functions exist in this codebase, so `prokind='f'` filter is safe.
- Tables: explicit `ash.sample` (partitioned parent), `ash.query_map_all` (view), `ash.config`, `ash.wait_event_map`; dynamic regex `^(sample|query_map)_[0-9]+$` covers the per-slot partitions/dictionaries (REVOKE on a partitioned parent correctly does NOT cascade, so the dynamic loop is necessary).
- No sequences in `ash.*` — none to worry about.

Owner access is preserved via ownership (not explicit GRANT), which is correct.

---

## Bug hunter

### POTENTIAL — B1: IMMUTABLE SQL helpers no longer inlined (confidence 6)

- **file:** `sql/ash-install.sql` — affects `ash.epoch()` (line 49), `ash._to_sample_ts()` (line 1615), `ash._color_on()` (line 1120), `ash._reset()` (line 1154), `ash._wait_color()` (line 1128), `ash._bar()` (line 1165), `ash.current_slot()` (line 111), `ash._active_slots()` (line 1001)
- **issue:** Adding `SET search_path` to a SQL-language function disables inlining (PG docs: "A SQL function that has a SET clause will not be inlined into a calling query"). `ash.epoch()` alone appears ~25 times in the hardened readers, often in WHERE clauses like `s.sample_ts >= extract(epoch from now() - p_interval - ash.epoch())::int4`. The function is still IMMUTABLE and takes no args, so the planner may still constant-fold the call — but the per-query overhead changes from "compile-time inline" to "cached constant".
- **fix:** Measure. If any regression shows up on large samples, either (a) strip `SET search_path` from these tiny SQL helpers (they don't reference `ash.*` tables so don't need the guard for correctness, only for uniform defense-in-depth), or (b) replace `ash.epoch()` with a hardcoded `'2026-01-01 00:00:00+00'::timestamptz` constant where it's a hot-path expression.
- **note:** CI passed in under 45s per PG major, so no visible regression in the test workload. Flagging as potential.

### POTENTIAL — B2: `v_pgss_readers` drift between install and upgrade script (confidence 7)

- **file:** `sql/ash-install.sql` vs `sql/ash-1.3-to-1.4.sql`
- **issue:** Install hardcodes `search_path = pg_catalog, ash, public` per-function at CREATE FUNCTION time. Upgrade script maintains a parallel `v_pgss_readers text[]` array. These two sources must stay in sync. If a future PR adds a new pgss-reader to `ash-install.sql` without updating the array in `ash-1.3-to-1.4.sql` (or subsequent upgrade script), upgrading users get `pg_catalog, ash` only and the new reader breaks on managed services.
- **fix:** Either (a) make `ash-install.sql` use the same dynamic loop pattern as the upgrade (enumerate pgss-readers from `pg_proc.prosrc LIKE '%pg_stat_statements%'`), or (b) add a code comment cross-referencing both places, or (c) add a CI test that greps both files for parity. This is technically out-of-scope for this PR (CLAUDE.md says upgrade scripts are generated at release time), so the risk is bounded — but since this PR ships both files together, worth documenting.

### INFO — B3: `prokind = 'f'` excludes aggregates and window functions (confidence 9)

- **file:** `sql/ash-install.sql` line 2967, `sql/ash-1.3-to-1.4.sql` line 355
- **issue:** The filter `p.prokind = 'f'` catches normal functions only. If `ash.*` ever adds an aggregate (`prokind='a'`), procedure (`'p'`), or window function (`'w'`), the REVOKE loop will miss it.
- **evidence:** current codebase has no aggregates/procedures/window functions in `ash.*`, so this is theoretical. Flagging for future maintainers.
- **fix:** Change to `p.prokind in ('f', 'a', 'w', 'p')` or simply `p.prokind <> ''` (no exclusion). EXECUTE applies to all four kinds.

---

## Guidelines checker

### POTENTIAL — G1: lowercase SQL keywords consistent with project style (confidence 9)

Verified: all new SQL uses lowercase keywords (`alter function`, `revoke`, `grant`, `execute`, `format`) consistent with `development__db-sql-style-guide.mdc` and the rest of `ash-install.sql`. No violations.

### INFO — G2: CLAUDE.md compliance (confidence 10)

- `ash-install.sql` remains the source of truth — both it and the upgrade script were modified consistently. Compliant with CLAUDE.md "Deployment" section.
- No new extensions, no `.control` file, no dependency on pg_cron/pgss for correctness — compliant with "anti-extension design".
- Upgrade script `ash-1.3-to-1.4.sql` correctly ships the hardening for existing installs.

---

## Test analyzer

### POTENTIAL — T1: No test asserts REVOKE state (confidence 7)

- **file:** `.github/workflows/test.yml` (no grep matches for `revoke` / `grant` / `search_path`)
- **issue:** The hardening changes ACLs on ~38 functions and 10+ tables, but no test verifies:
  - An unprivileged role is denied `SELECT ash.top_queries()`.
  - An unprivileged role cannot `SELECT * FROM ash.sample`.
  - The schema owner still has full access after upgrade.
  - After re-running the upgrade script on an already-upgraded DB, no errors and no permission drift.
- **fix:** Add a smoke test (e.g. in `test.yml`): `CREATE ROLE tenant_a NOLOGIN; SET ROLE tenant_a; SELECT ash.top_queries();` — expect permission denied. And: `RESET ROLE; SELECT ash.top_queries();` — expect success.
- **note:** Per CLAUDE.md the PR author's test plan included this manually ("as a non-owner role, `select ash.top_queries()` is denied"), but it's unchecked in the PR body and not automated.

### POTENTIAL — T2: No test covers degraded pgss-schema path (confidence 5)

- Related to S1: no test installs pgss in a non-`public` schema to verify the reader functions degrade gracefully.

---

## Docs reviewer

### POTENTIAL — D1: No guidance for granting access back to specific roles (confidence 8)

- **file:** `README.md` (unchanged in this PR)
- **issue:** The PR revokes SELECT on reader tables and EXECUTE on reader functions from `PUBLIC`. Legitimate users — SREs, read-only dashboard roles, monitoring users — now need explicit grants. There's no documentation for the supported grant-back recipe.
- **suggestion:** Add a short README section, e.g.:
  ```sql
  -- Grant read-only access to a monitoring role:
  grant usage on schema ash to monitor_ro;
  grant select on ash.sample, ash.query_map_all, ash.config, ash.wait_event_map to monitor_ro;
  -- For per-slot partitions:
  grant select on all tables in schema ash to monitor_ro;
  grant execute on all functions in schema ash to monitor_ro;
  ```
  Call out that `GRANT ... ON ALL TABLES IN SCHEMA ash` DOES cascade to partitions when issued on the schema (unlike the partitioned-parent REVOKE cascade asymmetry).
- **why it matters:** CLAUDE.md explicitly calls out managed services (RDS/Supabase/Neon/AlloyDB) where the install role and the day-to-day admin role are often different. Without guidance, operators will discover the hardening the hard way.

### INFO — D2: Inline comments are excellent (confidence 10)

The header comments on the two DO blocks and the `v_pgss_readers` rationale comment in the upgrade script are genuinely helpful — they document the "why" (REVOKE on partitioned parent doesn't cascade; pgss probe degrades silently) not just the "what". Good example for future PRs.

---

## Top issues (prioritized)

1. **S1 / T2** — pgss-schema assumption is new and undocumented. Users with pgss outside `public` see broken readers. Either document the requirement in README or resolve pgss schema dynamically.
2. **D1** — Add grant-back recipe to README. This is the highest user-impact gap.
3. **T1** — Add a simple non-owner REVOKE-denial smoke test to `test.yml`.
4. **B2** — Cross-reference `v_pgss_readers` between install.sql and upgrade script to prevent drift.

None are blocking. The PR is a net-positive security improvement and can land as-is with follow-up issues for the docs and test gaps.

---

PR #56, opened 2026-04-24
fix: align sample_data_check constraint across upgrade paths

### LINKED ISSUE #49: bug: sample_data_check constraint diverges between fresh install and upgrade from 1.0
Found while testing v1.4 with a real upgrade-chain scenario.

## Summary
`ash.sample.sample_data_check` ends up as `array_length(data, 1) >= 2` on installs that chained 1.0 → 1.1 → … → 1.4, but is `>= 3` on fresh installs of 1.4. The stricter check is what `_validate_data` expects; existing installs silently run with the looser one.

## Root cause
`sql/ash-1.1.sql` tightened the CHECK from `>= 2` to `>= 3`, but wraps the table in `create table if not exists`. On upgrade from 1.0 the table already exists, so the new constraint is never applied. No subsequent upgrade script (1.1→1.2, 1.2→1.3, 1.3→1.4) re-applies it.

## Impact
- Operational: existing 1.0-chain installs accept ill-formed rows the validator rejects.
- CI invariant: schema-equivalence test passes because it only compares function `prosrc` md5s, not constraint defs.

## Fix
Add an idempotent `ALTER TABLE` block in `sql/ash-install.sql` (and mirror into `sql/ash-1.3-to-1.4.sql` per CLAUDE.md in-progress policy) that drops the old constraint and re-adds the `>= 3` form, guarded so it only runs when needed.

## Detected by
v1.4 post-release validation — see #37 comment thread and the test report on session 2026-04-24.

NikolayS:
## Summary

Fixes #49.

- v1.0 shipped `ash.sample` with `check (... and array_length(data, 1) >= 2)`. v1.1 tightened it to `>= 3`, but no legacy upgrade script (1.0->1.1, 1.1->1.2, 1.2->1.3) ever rewrote the constraint. Any install that started at 1.0 still carries the looser form, diverging from a fresh install at the same version.
- Adds an idempotent `DO` block to `sql/ash-install.sql` that inspects `pg_get_constraintdef` for `sample_data_check` on the partitioned parent and only rewrites it when it is missing or in the legacy `>= 2` form. The drop on the parent cascades to existing partitions; the re-add propagates the tightened check back to all partitions.
- `sql/ash-1.3-to-1.4.sql` is `\ir ash-install.sql`, so it inherits the fix automatically per the CLAUDE.md lockstep policy. Finalized legacy scripts (1.0, 1.1, 1.1->1.2, 1.2->1.3) are not touched.
- Adds a dedicated CI step that installs 1.0 (and asserts the legacy `>= 2` precondition holds), chains 1.0 -> 1.1 -> 1.2 -> 1.3 -> 1.4, and asserts every `ash.sample*` CHECK is the `>= 3` form on both the parent and every partition.

## Test plan

- [x] Fresh install of `ash-install.sql` produces `array_length(data, 1) >= 3` (verified locally on PG 18.3).
- [x] Re-applying `ash-install.sql` is a no-op for the constraint (idempotent guard).
- [x] Full 1.0 -> 1.4 upgrade chain ends with `>= 3` on the parent and all 3 partitions; no row carries the legacy `>= 2` form (new regression test, also runs locally).
- [ ] Existing `Schema equivalence` and full upgrade-chain CI jobs still pass on PG 14-18.

Co-Authored-By: Claude Opus 4.7 (1M context) <noreply@anthropic.com>

Closes #49

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -2462,6 +2462,82 @@ jobs:
           select ash.uninstall('yes');
           EOF
 
+      - name: "Issue #49: sample_data_check tightened by upgrade chain from 1.0"
+        env:
+          PGPASSWORD: postgres
+        run: |
+          # Regression for issue #49: v1.0 shipped `array_length(data, 1) >= 2`,
+          # v1.1+ requires `>= 3`. No legacy upgrade script rewrote the check,
+          # so installs that started at 1.0 carried the looser form forever.
+          # ash-install.sql now detects and rewrites it; assert that the full
+          # 1.0 -> 1.4 chain ends with the tightened constraint on both the
+          # partitioned parent and every partition.
+          psql -h localhost -U postgres -d postgres -v ON_ERROR_STOP=1 << 'EOF'
+          \i sql/ash-1.0.sql
+
+          -- Sanity-check the precondition: 1.0 ships the looser form.
+          do $$
+          declare
+            v_def text;
+          begin
+            select pg_get_constraintdef(c.oid) into v_def
+            from pg_constraint c
+            where c.conrelid = 'ash.sample'::regclass
+              and c.conname  = 'sample_data_check';
+            assert v_def ~ 'array_length\(data, 1\) >= 2',
+              format('1.0 should ship `>= 2` form; got: %s', v_def);
+          end $$;
+
+          \i sql/ash-1.1.sql
+          \i sql/ash-1.1-to-1.2.sql
+          \i sql/ash-1.2-to-1.3.sql
+          \i sql/ash-1.3-to-1.4.sql
+
+          -- After full chain: parent + every partition must carry `>= 3`,
+          -- and none may carry the legacy `>= 2`.
+          do $$
+          declare
+            v_bad int;
+            v_parent_def text;
+          begin
+            select pg_get_constraintdef(c.oid) into v_parent_def
+            from pg_constraint c
+            where c.conrelid = 'ash.sample'::regclass
+              and c.conname  = 'sample_data_check';
+            assert v_parent_def ~ 'array_length\(data, 1\) >= 3',
+              format('parent constraint not tightened; got: %s', v_parent_def);
+
+            select count(*) into v_bad
+            from pg_constraint c
+            join pg_class    t on t.oid = c.conrelid
+            join pg_namespace n on n.oid = t.relnamespace
+            where n.nspname = 'ash'
+              and t.relname ~ '^sample(_[0-9]+)?$'
+              and c.contype = 'c'
+              and pg_get_constraintdef(c.oid) ~ 'array_length\(data, 1\) >= 2';
+            assert v_bad = 0,
+              format('%s ash.sample* check(s) still carry the legacy `>= 2` form', v_bad);
+
+            select count(*) into v_bad
+            from pg_inherits i
+            join pg_class    p on p.oid = i.inhrelid
+            join pg_namespace n on n.oid = p.relnamespace
+            left join pg_constraint c
+              on c.conrelid = p.oid
+             and c.contype  = 'c'
+             and pg_get_constraintdef(c.oid) ~ 'array_length\(data, 1\) >= 3'
+            where i.inhparent = 'ash.sample'::regclass
+              and n.nspname = 'ash'
+              and c.oid is null;
+            assert v_bad = 0,
+              format('%s partition(s) of ash.sample missing `>= 3` check', v_bad);
+
+            raise notice 'issue #49 regression PASSED: sample_data_check is `>= 3` everywhere';
+          end $$;
+
+          select ash.uninstall('yes');
+          EOF
+
       - name: "Upgrade path via documented wrappers + data preservation (issue #37 H-CI-1, H-CI-2)"
         env:
           PGPASSWORD: postgres
--- a/sql/ash-install.sql
+++ b/sql/ash-install.sql
@@ -301,6 +301,32 @@ begin
   end loop;
 end $$;
 
+-- Migration (issue #49): align sample.data check across upgrade paths.
+-- v1.0 shipped `array_length(data, 1) >= 2`; v1.1 tightened it to `>= 3`,
+-- but no legacy upgrade script rewrote the constraint, so installs that
+-- started at 1.0 still carry the looser form. Detect and fix in place.
+-- Idempotent: only rewrites when the current definition is missing or
+-- the old `>= 2` form. Drops on the partitioned parent cascade to all
+-- partitions; ADD CONSTRAINT on the parent propagates back to children.
+do $$
+declare
+  v_def text;
+begin
+  select pg_get_constraintdef(c.oid) into v_def
+  from pg_constraint c
+  where c.conrelid = 'ash.sample'::regclass
+    and c.conname  = 'sample_data_check';
+
+  if v_def is null or v_def !~ 'array_length\(data, 1\) >= 3' then
+    if v_def is not null then
+      alter table ash.sample drop constraint sample_data_check;
+    end if;
+    alter table ash.sample
+      add constraint sample_data_check
+      check (data[1] < 0 and array_length(data, 1) >= 3);
+  end if;
+end $$;
+
 -- Convert timestamptz to int4 epoch offset
 create or replace function ash.ts_from_timestamptz(p_ts timestamptz)
 returns int4


NikolayS:
# REV review — PR #56

`fix: align sample_data_check constraint across upgrade paths` — closes #49.

Verdict: **APPROVE** (no blockers). The fix is small, scoped, idempotent, and accompanied by a dedicated regression test. CI is green on PG 14-18 (with and without pgss). The drop+add pattern on the partitioned parent is the correct PG mechanism for propagating a CHECK to all partitions.

---

## BLOCKING

NO_FINDINGS.

---

## NON-BLOCKING

NO_FINDINGS.

---

## POTENTIAL

### P1 — Regex-only probe is correct today, but coupled to PG's `pg_get_constraintdef` canonical form (LOW, confidence 5)

- file: `sql/ash-install.sql` lines 312-321 (PR branch)
- issue: The probe `if v_def is null or v_def !~ 'array_length\(data, 1\) >= 3'` relies on the textual form returned by `pg_get_constraintdef`. PostgreSQL canonicalizes whitespace and casing, so today this matches `CHECK ((data[1] < 0) AND (array_length(data, 1) >= 3))`. If a future PG release changes that canonical form (e.g., normalizes `>=` to `>= ` differently, prints `array_length(data, 1::int)`, etc.), the probe could mis-classify a tightened constraint as legacy and trigger an unnecessary drop+re-add. Drop+re-add on a populated partitioned parent acquires `AccessExclusiveLock` and re-validates every existing row.
- fix (optional): Use `pg_constraint.conbin` parsed via `pg_node_tree` is overkill; a sturdier approach is to assert both substrings (`array_length(data, 1) >= 3` AND `data[1] < 0`) and otherwise leave alone. As-is is acceptable given the scope, but worth a comment that the probe is bound to the current `pg_get_constraintdef` output.

### P2 — Drop+add on a populated parent re-validates every row (LOW, confidence 6)

- file: `sql/ash-install.sql` lines 312-321
- issue: `ALTER TABLE ... ADD CONSTRAINT ... CHECK (...)` (without `NOT VALID`) full-scans every partition to validate the constraint. On a 1.0->1.4 install with a long retention window this can be a multi-minute, exclusive-locked operation that blocks concurrent `ash.take_sample()` calls (which hold the participation advisory lock but not the table lock). For a typical pg_ash deployment the sample tables are small (3 partitions, current slot only) so this is unlikely to bite, but it's worth noting in the migration comment that operators on very large installations may want to call `ash.stop()` first.
- fix (optional): Document the lock + revalidation cost in the inline comment, or add `NOT VALID` then `VALIDATE CONSTRAINT` as a follow-up if scale becomes an issue. Not required for this PR.

### P3 — Probe runs on every install, not just upgrade-from-1.0 (LOW, confidence 7)

- file: `sql/ash-install.sql` lines 304-321
- issue: The DO block runs unconditionally on every `\i ash-install.sql`, including fresh installs and re-applies on already-tightened systems. The early-exit branch (`v_def !~ '... >= 3'` is false) makes it a cheap catalog lookup, so this is fine — just calling out that the cost is paid every install. The schema-equivalence CI job and the existing "Dev workflow: re-apply migrations (idempotent)" job confirm the no-op path works.
- fix: None required.

### P4 — Test asserts partitions match `>= 3` but does not negative-assert that the parent retained the legacy form before the fix ran (LOW, confidence 4)

- file: `.github/workflows/test.yml` lines 2465-2538
- issue: The regression test correctly verifies that 1.0 ships `>= 2` and that the full chain ends with `>= 3` everywhere. It does not, however, assert that `\i ash-1.1.sql` alone (i.e. without `ash-install.sql`'s new DO block) leaves the constraint at `>= 2` — which is the exact state issue #49 documented. Today only `ash-1.3-to-1.4.sql` (which `\ir ash-install.sql`) tightens the constraint via the new DO block; the test happens to chain through 1.4 so it passes. If a future maintainer moves the migration earlier (e.g. into 1.1->1.2), the test would still pass without exercising the new code path.
- fix (optional): After `\i sql/ash-1.2-to-1.3.sql`, add a quick assertion that the constraint is still `>= 2` (proving the chain alone wouldn't fix it), then `\i sql/ash-1.3-to-1.4.sql` and assert `>= 3`. This pins the regression to the actual fix location.

---

## Summary

| Bucket | Count |
| --- | --- |
| BLOCKING | 0 |
| NON-BLOCKING | 0 |
| POTENTIAL | 4 (all LOW) |

**Strengths**
- Correct root-cause diagnosis: 1.0 shipped `>= 2`, no legacy upgrade script rewrote it, fresh-vs-upgraded schemas diverged.
- Idempotent guard: probe-then-rewrite avoids unnecessary churn on already-correct installs.
- Drop on the partitioned parent cascades to all partitions; ADD on the parent propagates the new CHECK back to every child — the standard PG mechanism, no per-partition `ALTER TABLE` needed.
- Dedicated CI step that asserts the precondition (1.0 has `>= 2`) and the full chain end-state (parent + every partition has `>= 3`, none has the legacy form). Asserts span both `pg_constraint` on `ash.sample` and the partition set via `pg_inherits`.
- The existing `Schema equivalence: fresh install vs full upgrade path` job (lines 2679+) gives belt-and-suspenders coverage — any future divergence in the `data` column's CHECK definition would surface there too.
- Correctly leaves finalized legacy upgrade scripts (1.0, 1.1, 1.1->1.2, 1.2->1.3) untouched per the lockstep policy in `CLAUDE.md`. The fix flows through `ash-1.3-to-1.4.sql` -> `\ir ash-install.sql`.
- All PG 14-18 jobs pass, plus pgss-off matrix entry.

**Watch-items addressed**
- Probe correctly identifies both "missing" (`v_def is null`) and "legacy `>= 2`" (regex doesn't match `>= 3`) cases — confirmed by reading the DO block.
- Drop+add on the partitioned parent does propagate to all `sample_N` partitions in PG 14+ (verified by the test's `pg_inherits` walk).
- No race window with active sampling that would observe a flux state: `ALTER TABLE ... DROP/ADD CONSTRAINT` takes `AccessExclusiveLock`, queuing concurrent `take_sample()` INSERTs; once granted, INSERTs proceed against the new constraint. Pre-existing rows from a 1.0 install all encode at least `[-wid, count, qid]` (3 elements) since the encoder always emits a query_map_id, so revalidation will not fail on legitimate data.

---

NikolayS:
## Testing evidence

Verified on fresh `postgres:18-alpine` containers using PR branch `claude/fix-issue-49-sample-data-check` (commit `4e4914c`).

### 1. Fresh install — all constraints `>= 3`

```
$ \i sql/ash-install.sql
$ SELECT conname, pg_get_constraintdef(oid) AS def
    FROM pg_constraint
    WHERE conname LIKE 'sample%_data_check'
    ORDER BY conname;

      conname      |                           def
-------------------+----------------------------------------------------------
 sample_data_check | CHECK (((data[1] < 0) AND (array_length(data, 1) >= 3)))
 sample_data_check | CHECK (((data[1] < 0) AND (array_length(data, 1) >= 3)))
 sample_data_check | CHECK (((data[1] < 0) AND (array_length(data, 1) >= 3)))
 sample_data_check | CHECK (((data[1] < 0) AND (array_length(data, 1) >= 3)))
(4 rows)
```

### 2. Upgrade chain from 1.0 → 1.4 — constraint correctly tightened

Before upgrade (1.0 only):

```
$ \i sql/ash-1.0.sql
$ SELECT conname, pg_get_constraintdef(oid) AS def
    FROM pg_constraint WHERE conname LIKE 'sample%_data_check' ORDER BY conname;

      conname      |                           def
-------------------+----------------------------------------------------------
 sample_data_check | CHECK (((data[1] < 0) AND (array_length(data, 1) >= 2)))
 sample_data_check | CHECK (((data[1] < 0) AND (array_length(data, 1) >= 2)))
 sample_data_check | CHECK (((data[1] < 0) AND (array_length(data, 1) >= 2)))
 sample_data_check | CHECK (((data[1] < 0) AND (array_length(data, 1) >= 2)))
(4 rows)
```

After chain to 1.4 (`ash-1.0-to-1.1.sql; ash-1.1-to-1.2.sql; ash-1.2-to-1.3.sql; ash-1.3-to-1.4.sql`):

```
      conname      |                           def
-------------------+----------------------------------------------------------
 sample_data_check | CHECK (((data[1] < 0) AND (array_length(data, 1) >= 3)))
 sample_data_check | CHECK (((data[1] < 0) AND (array_length(data, 1) >= 3)))
 sample_data_check | CHECK (((data[1] < 0) AND (array_length(data, 1) >= 3)))
 sample_data_check | CHECK (((data[1] < 0) AND (array_length(data, 1) >= 3)))
(4 rows)
```

The DO block in `ash-1.3-to-1.4.sql` (via `\ir ash-install.sql`) detected the legacy `>= 2` form and rewrote it.

### 3. Idempotent re-apply

Running `\i sql/ash-install.sql` a second time on the upgraded container — no errors, definitions unchanged:

```
$ \i sql/ash-install.sql
...
DO

$ SELECT conname, pg_get_constraintdef(oid) AS def
    FROM pg_constraint WHERE conname LIKE 'sample%_data_check' ORDER BY conname;

      conname      |                           def
-------------------+----------------------------------------------------------
 sample_data_check | CHECK (((data[1] < 0) AND (array_length(data, 1) >= 3)))
 sample_data_check | CHECK (((data[1] < 0) AND (array_length(data, 1) >= 3)))
 sample_data_check | CHECK (((data[1] < 0) AND (array_length(data, 1) >= 3)))
 sample_data_check | CHECK (((data[1] < 0) AND (array_length(data, 1) >= 3)))
(4 rows)
```

### 4. Partition propagation

Enumerated child partitions via `pg_inherits`. All three `sample_N` partitions carry the tight constraint on both containers.

Fresh install:

```
 partition |                           def
-----------+----------------------------------------------------------
 sample_0  | CHECK (((data[1] < 0) AND (array_length(data, 1) >= 3)))
 sample_1  | CHECK (((data[1] < 0) AND (array_length(data, 1) >= 3)))
 sample_2  | CHECK (((data[1] < 0) AND (array_length(data, 1) >= 3)))
(3 rows)
```

Upgraded-from-1.0:

```
 partition |                           def
-----------+----------------------------------------------------------
 sample_0  | CHECK (((data[1] < 0) AND (array_length(data, 1) >= 3)))
 sample_1  | CHECK (((data[1] < 0) AND (array_length(data, 1) >= 3)))
 sample_2  | CHECK (((data[1] < 0) AND (array_length(data, 1) >= 3)))
(3 rows)
```

### 5. Negative test (len-2 row rejected)

Inserting a row with `data = '{-1, 2}'::int4[]` (length 2) into the upgraded `ash.sample` is rejected, proving the tightened constraint is enforced:

```
$ INSERT INTO ash.sample (sample_ts, datid, active_count, data)
    VALUES (1, 0, 0, '{-1, 2}'::int4[]);
ERROR:  new row for relation "sample_0" violates check constraint "sample_data_check"
DETAIL:  Failing row contains (1, 0, 0, {-1,2}, 0).
```

**Result:** Fix verified end-to-end. Fresh and upgrade-from-1.0 paths converge to identical constraint definitions.

PR #57, opened 2026-04-24
fix(readers): clamp interval-to-int4 underflow (#51)

### LINKED ISSUE #51: bug: readers raise 'integer out of range' on absurd intervals instead of returning empty as the NOTICE promises
## Summary
\`\`\`sql
SELECT ash.top_waits('1000 years');
NOTICE:  interval 1000 years exceeds 2× rotation_period; returning empty results.
ERROR:  integer out of range
\`\`\`

The NOTICE (emitted from `ash._active_slots_for`) advertises that the reader will return empty; the follow-up `extract(epoch from now() - p_interval - ash.epoch())::int4` cast in the reader then errors because the computed int4 value underflows.

## Impact
UX only — reachable only for absurd inputs. But a reader that promises "empty results" then throws is misleading.

## Fix
Bound the cast with `greatest` to keep it in int4 range, or short-circuit the reader when `_active_slots_for` signals the interval is too large. Apply to every reader using this pattern — agent listed 20+ sites in sql/ash-install.sql.

## Detected by
v1.4 post-release validation.

NikolayS:
## Summary

Closes #51. Reader functions accepting a user-supplied `interval` were computing the lower-bound `sample_ts` via

```sql
extract(epoch from now() - p_interval - ash.epoch())::int4
```

For absurd intervals (e.g. `'1000 years'`), the subtraction underflowed `int4`, and the `::int4` cast aborted with `integer out of range` — even though `_active_slots_for()` had already emitted a NOTICE promising the result would be empty (older samples not retained).

The fix clamps with `greatest(..., 0)` before the cast at all 12 call sites in `sql/ash-install.sql`. `sample_ts` is a non-negative `int4` epoch offset by construction, so any underflow correctly maps to "no rows older than the epoch" — i.e. the empty result the NOTICE already advertises.

## Sites fixed (12 in `ash-install.sql`)

- `top_waits`, `wait_timeline`, `top_queries` (pgss + non-pgss branches), `top_by_type`, `top_queries_with_text`, `query_waits`, `samples_by_database`, `activity_summary`, `timeline_chart`, `samples`, `event_queries`

Legacy upgrade scripts (`ash-1.0` … `ash-1.2-to-1.3`) are immutable per CLAUDE.md and are not touched. The in-progress `ash-1.3-to-1.4.sql` already does `\ir ash-install.sql`, so the fix flows through; only the changelog comment was extended.

## Test plan

- [x] Added regression in `.github/workflows/test.yml` (edge cases section): `select ash.top_waits('1000 years')` must return 0 rows, not raise.
- [ ] CI green across PG 14–18.
- [ ] Manual: confirm a NOTICE is still emitted for the oversized interval (behavior of `_active_slots_for()` unchanged).

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -1272,6 +1272,62 @@ jobs:
             select count(*) into v_count from ash.top_waits('1000 years');
             assert v_count = 0, 'top_waits(1000 years) should return 0 rows, got: ' || v_count;
 
+            select count(*) into v_count from ash.wait_timeline('1000 years');
+            assert v_count = 0, 'wait_timeline(1000 years) should return 0 rows, got: ' || v_count;
+
+            select count(*) into v_count from ash.top_queries('1000 years');
+            assert v_count = 0, 'top_queries(1000 years) should return 0 rows, got: ' || v_count;
+
+            select count(*) into v_count from ash.top_by_type('1000 years');
+            assert v_count = 0, 'top_by_type(1000 years) should return 0 rows, got: ' || v_count;
+
+            select count(*) into v_count from ash.top_queries_with_text('1000 years');
+            assert v_count = 0,
+              'top_queries_with_text(1000 years) should return 0 rows, got: ' || v_count;
+
+            -- query_waits short-circuits on unknown query_id before reaching
+            -- the interval clamp, so pick a real query_id (any registered one)
+            -- to ensure the clamp path is exercised. Falls back to a synthetic
+            -- id if query_map_all is empty (still valid: short-circuit returns
+            -- 0 rows without hitting int4 math).
+            declare
+              v_qid bigint;
+            begin
+              select query_id into v_qid from ash.query_map_all limit 1;
+              v_qid := coalesce(v_qid, 111111);
+              select count(*) into v_count from ash.query_waits(v_qid, '1000 years');
+              assert v_count = 0,
+                'query_waits(1000 years) should return 0 rows, got: ' || v_count;
+            end;
+
+            select count(*) into v_count from ash.samples_by_database('1000 years');
+            assert v_count = 0,
+              'samples_by_database(1000 years) should return 0 rows, got: ' || v_count;
+
+            -- activity_summary always emits a "no data in this time range" status
+            -- row when v_total_samples = 0; assert that single-row placeholder
+            -- shape rather than 0 rows.
+            select count(*) into v_count from ash.activity_summary('1000 years');
+            assert v_count = 1,
+              'activity_summary(1000 years) should return 1 placeholder row, got: ' || v_count;
+            select * into v_rec from ash.activity_summary('1000 years') limit 1;
+            assert v_rec.metric = 'status' and v_rec.value = 'no data in this time range',
+              'activity_summary(1000 years) placeholder mismatch: ' ||
+              coalesce(v_rec.metric, '<null>') || '/' || coalesce(v_rec.value, '<null>');
+
+            -- timeline_chart returns no rows when there are no top events
+            -- (v_top_events is null) — empty short-circuit, no legend emitted.
+            select count(*) into v_count from ash.timeline_chart('1000 years');
+            assert v_count = 0,
+              'timeline_chart(1000 years) should return 0 rows, got: ' || v_count;
+
+            select count(*) into v_count from ash.samples('1000 years');
+            assert v_count = 0, 'samples(1000 years) should return 0 rows, got: ' || v_count;
+
+            select count(*) into v_count from ash.event_queries('CPU', '1000 years');
+            assert v_count = 0,
+              'event_queries(1000 years) should return 0 rows, got: ' || v_count;
+
             -- === Non-existent queries ===
 
             select count(*) into v_count from ash.query_waits(999999999, '1 hour');
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -1265,6 +1265,13 @@ jobs:
             select count(*) into v_count from ash.top_by_type('1 second');
             select count(*) into v_count from ash.wait_timeline('1 second', '1 second');
 
+            -- === Absurd intervals (regression: #51) ===
+            -- now() - '1000 years' would underflow int4 epoch offset; readers
+            -- must clamp and return empty (the NOTICE promises "older samples
+            -- not available"), not raise integer-out-of-range.
+            select count(*) into v_count from ash.top_waits('1000 years');
+            assert v_count = 0, 'top_waits(1000 years) should return 0 rows, got: ' || v_count;
+
             -- === Non-existent queries ===
 
             select count(*) into v_count from ash.query_waits(999999999, '1 hour');
--- a/sql/ash-1.3-to-1.4.sql
+++ b/sql/ash-1.3-to-1.4.sql
@@ -23,12 +23,13 @@
 --   - search_path guard on every ash.* function + pgss-schema-aware readers;
 --     REVOKE EXECUTE on all ash.* functions and SELECT on reader tables from
 --     PUBLIC (#45).
---   - Reader interval-to-int4 underflow clamp + oversized-interval
---     short-circuit: top_waits / top_queries / top_by_type / wait_timeline /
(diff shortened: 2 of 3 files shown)


NikolayS:
# REV review — PR #57

**Scope:** `fix(readers): clamp interval-to-int4 underflow (#51)` — `claude/fix-issue-51-reader-overflow`. Two-layer fix: (1) `ash._active_slots_for(p_interval)` now returns `array[]::smallint[]` for oversized intervals so reader JOINs naturally yield empty; (2) all 12 interval-based readers wrap the lower-bound cast in `greatest(extract(...), 0)::int4`. CI green on all 8 checks. One regression test added (`top_waits('1000 years')`).

## Verdict

**Approve with caveats.** The fix is correct, minimal, and the two-layer design is sound. However: the same int4-underflow class affects the `_at` variants — they are unfixed and parallel to #51. The regression test covers only 1 of 12 patched readers; recommend expanding both surfaces in a follow-up.

---

## Bug hunter

```
FINDING:
- severity: HIGH
- confidence: 9
- file: sql/ash-install.sql
- line: 312 (ash.ts_from_timestamptz)
- issue: `_at` variants raise `integer out of range` on absurd absolute timestamps —
  same root cause as #51, unfixed by this PR.
- evidence:
    create or replace function ash.ts_from_timestamptz(p_ts timestamptz)
    ...
      select extract(epoch from p_ts - ash.epoch())::int4
  Confirmed on PG 18.3:
    select extract(epoch from '1000-01-01'::timestamptz - '2025-01-01'::timestamptz)::int4;
    -> ERROR:  integer out of range
    select extract(epoch from '3000-01-01'::timestamptz - '2025-01-01'::timestamptz)::int4;
    -> ERROR:  integer out of range
  Affects every caller of `ts_from_timestamptz` / `_to_sample_ts`:
  `top_waits_at`, `top_queries_at`, `wait_timeline_at`, `top_by_type_at`,
  `query_waits_at`, `samples_at`, `event_queries_at`, `timeline_chart_at`,
  `minute_waits_at`, `hourly_queries_at`, `daily_peak_backends_at`.
  The _at readers happily accept any timestamptz and currently error before the
  JOIN runs. Note: int4 also overflows in the *future* direction (~year 2038
  + epoch offset), so a clamp must handle both ends.
- fix: Either (a) clamp inside `ash.ts_from_timestamptz`:
    select greatest(least(extract(epoch from p_ts - ash.epoch()),
                          2147483647::float), 0)::int4
  or (b) leave the conversion strict and have `_at` readers validate/clamp
  `p_start`/`p_end` before calling. Option (a) is consistent with the
  defense-in-depth approach already adopted for interval readers.
- not blocking for #51 specifically (issue text describes interval readers),
  but should be tracked as a sibling issue before next release.
```

```
FINDING:
- severity: LOW
- confidence: 7
- file: sql/ash-install.sql
- line: 1538-1554
- issue: `_active_slots_for` correctly distinguishes the two empty-array
  cases (oversized -> NOTICE + return {}; happy path -> return slots).
  However, after `current_slot` rotates, the second slot is computed as
  `((current_slot - 1 + 3) % 3)::smallint` — hard-coded `3` rather than
  `num_partitions`. This pre-existed #57, but it's now adjacent to new
  logic; if `num_partitions` is ever != 3 the second-slot math is wrong
  for both `_active_slots()` and `_active_slots_for()`.
- evidence:
    return array[
      v_current_slot,
      ((v_current_slot - 1 + 3) % 3)::smallint
    ];
- fix: Out of scope here; flag for a separate change.
```

No other runtime bugs in the diff.

---

## Test analyzer

```
FINDING:
- severity: MEDIUM
- confidence: 8
- line: 1268-1273
- issue: Regression coverage is thin — only `ash.top_waits('1000 years')` is
  exercised. The fix touches 12 reader sites; if a future refactor reintroduces
  the unguarded cast on any of the 11 untested readers, CI won't catch it.
- untested_scenarios:
    - top_queries('1000 years')
    - top_by_type('1000 years')
    - wait_timeline('1 second', '1000 years')
    - top_queries_with_text('1000 years')
    - query_waits(<id>, '1000 years')
    - samples_by_database('1000 years')
    - activity_summary('1000 years')
    - timeline_chart('1000 years', '1 minute')
    - samples('1000 years', 10)
    - event_queries('Lock', 'transactionid', '1000 years')
    - NOTICE-still-emitted assertion (the changelog promises the NOTICE
      remains; no test verifies it)
    - Boundary: p_interval == 2 * rotation_period (exact equality —
      should NOT trigger the oversized branch since check is strict `>`)
    - Boundary: p_interval = 2 * rotation_period + epsilon (should trigger
      empty + NOTICE)
- A parametrized loop calling each of the 12 readers with `'1000 years'`
  inside a single DO block keeps the CI cost negligible.
```

```
FINDING:
- severity: MEDIUM
- confidence: 8
- line: ~1240-1257 (existing `_at` future-timestamp tests)
- issue: No regression test for `_at` variants with absurd ABSOLUTE
  timestamps (e.g. p_start = '1000-01-01' or year > 2106). They currently
  raise `integer out of range` from `ts_from_timestamptz`.
- untested_scenarios:
    - top_waits_at('1000-01-01', now())
    - top_queries_at(now(), '3000-01-01')
    - samples_at('1000-01-01', '3000-01-01')
  This is a pre-existing gap, not a regression introduced by #57. Worth
  filing as a follow-up; the fix shape will mirror this PR.
```

---

## Guidelines checker

```
FINDING:
- severity: INFO
- confidence: 7
- file: sql/ash-1.3-to-1.4.sql
- line: 26-31
- issue: Changelog bullet enumerates fixed reader names but omits
  `top_queries_with_text` and `query_waits` (both are clamped at lines
  2843 and 2913 in ash-install.sql). For grep-able accuracy of the
  changelog, the list should match the actual call sites.
- rule: writing__terminology / changelog accuracy
- suggestion: Either add the two missing names or replace the explicit
  list with "all interval-based readers".
```

```
FINDING:
- severity: INFO
- confidence: 6
- file: sql/ash-install.sql
- line: 1548-1553
- issue: Comment block reuses "v_min_ts" terminology even though
  `_active_slots_for` itself never computes `v_min_ts`. The variable
  belongs to the calling readers. Minor — slightly confusing to a
  reader landing on this function in isolation.
- rule: development__core-principles (clarity in comments)
- suggestion: Phrase as "Without this, an absurd interval like '1000
  years' would clamp the readers' lower-bound timestamp to 0...".
```

No SQL style violations: lowercase keywords, snake_case, parameter prefix `p_`, local prefix `v_` — all consistent with the existing codebase.

---

## Docs reviewer

```
FINDING:
- severity: LOW
- confidence: 6
- file: sql/ash-install.sql
- line: 1510-1521 (function header for `_active_slots_for`)
- issue: Header comment says the function "Never clamps the caller's
  interval — the returned row set stays honest." After this PR the
  function DOES effectively clamp by returning `{}` for oversized
  intervals — the row set is now empty, not "honest". Updated behavior
  contradicts the existing comment.
- suggestion: Update the header to note the oversized-interval branch
  short-circuits to `{}` (matching the inline comment at line 1548).
```

Otherwise documentation is adequate: PR body explains the fix shape, the inline comment at 1548-1553 explains the design choice, and the upgrade-script changelog calls out the user-visible behavior.

---

## Security reviewer

NO_FINDINGS.

The clamp narrows reader output (replaces an error with `{}`), does not widen access. No injection vectors, no SECURITY DEFINER additions, no privilege change. Behavior change is benign from an access-control perspective — `slot = any({})` returns false, never accidentally widening visibility.

---

## Summary

- **Approve once #51 closure is the only ask.** The fix is correct for what it claims to fix.
- **Strongly recommend a sibling issue + PR** for `_at`-variant overflow (same class, parallel surface, ~13 functions). Confirmed reproducible on PG 18.3 with `'1000-01-01'::timestamptz` and `'3000-01-01'::timestamptz`.
- **Recommend a small test expansion** before merge: a parametrized DO block that calls all 12 patched readers with `'1000 years'`, plus an assertion that the NOTICE is emitted exactly once. Keeps the regression net wide enough to defend the 12 sites the PR claims to cover.
- Two minor changelog/comment touch-ups noted (changelog reader-list completeness; updated `_active_slots_for` header).

PR #58, opened 2026-04-24
feat: add decode_sample(int) and decode_sample(timestamptz) convenience overloads

### LINKED ISSUE #54: feature: add ash.decode_sample(int) convenience overload on sample_ts
## Problem
`ash.decode_sample(integer[], smallint)` requires passing the packed `data` array and `slot` explicitly:

\`\`\`sql
SELECT ash.decode_sample(data, slot) FROM ash.sample WHERE sample_ts = 12345;
\`\`\`

A natural instinct is to pass the timestamp scalar:

\`\`\`sql
SELECT ash.decode_sample(12345);  -- ERROR: function ash.decode_sample(integer) does not exist
\`\`\`

## Proposal
Add `ash.decode_sample(p_sample_ts int)` that looks up the row(s) in `ash.sample` matching `sample_ts = p_sample_ts` and delegates to the existing `decode_sample(data, slot)`. Handle multiple rows (one per `datid`) gracefully — return all decoded rows across all matching datids.

Also consider `ash.decode_sample(timestamptz)` via `ts_from_timestamptz` for the common case of "decode the sample nearest to this wall-clock time".

## Detected by
v1.4 post-release validation.

NikolayS:
Closes #54.

## Summary
- Adds `ash.decode_sample(p_sample_ts int4)`: walks every `ash.sample` row whose `sample_ts` matches and returns `(datid oid, wait_event text, query_id bigint, count int)` so callers can distinguish per-database results.
- Adds `ash.decode_sample(p_ts timestamptz)`: converts to `sample_ts` via `ash.ts_from_timestamptz()` and delegates to the int4 overload. Same return shape.
- Both new overloads are `language sql` `STABLE` and use `LATERAL JOIN` over the existing 2-arg SRF, passing each row's `slot` for unambiguous `query_id` resolution (avoids the "search all partitions" branch that can return stale ids after rotation).
- The original `decode_sample(integer[], smallint)` signature is unchanged.

## Privileges
`PUBLIC EXECUTE` is revoked by the existing dynamic hardening loop (#45) — admins/readers need explicit grants. README documents this.

## Mirroring
`sql/ash-1.3-to-1.4.sql` `\ir`'s `ash-install.sql`, so the new functions ship automatically with the in-progress 1.4 upgrade. The header changelog was updated per CLAUDE.md.

## Tests
New CI step `Test decode_sample convenience overloads (int4, timestamptz)` exercises both overloads against synthetic `ash.sample` rows spanning two `datid`s. Asserts:
- Row count matches `sum(active_count)` of source rows
- `sum(d.count)` matches the same total
- Distinct `datid` count is correct
- Each expected `(datid, wait_event, query_id)` triple is present
- The `timestamptz` overload returns identical rows to the int4 overload after a `ts_to_timestamptz` round-trip
- Empty input (`-1`) returns zero rows

Existing decoder tests (2-arg call patterns) untouched and still pass locally on PG 18.

## Test plan
- [ ] CI passes on PG 14 / 15 / 16 / 17 / 18
- [ ] Schema-equivalence test (fresh install vs. upgrade chain) passes
- [ ] Existing `decode_sample(integer[], smallint)` callers still resolve correctly (overload resolution check)

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -1399,8 +1399,10 @@ jobs:
 
             -- === decode_sample corner cases ===
 
-            -- Empty/null data
-            select count(*) into v_count from ash.decode_sample(null);
+            -- Empty/null data (cast to integer[] to disambiguate from
+            -- the decode_sample(int4) overload, which would otherwise
+            -- both match an unknown-typed NULL literal).
+            select count(*) into v_count from ash.decode_sample(null::integer[]);
             assert v_count = 0, 'null decode should return 0 rows';
 
             select count(*) into v_count from ash.decode_sample('{}'::integer[]);
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -417,6 +417,121 @@ jobs:
           $$;
           EOF
 
+      - name: Test decode_sample convenience overloads (int4, timestamptz)
+        env:
+          PGPASSWORD: postgres
+        run: |
+          psql -h localhost -U postgres -d postgres -v ON_ERROR_STOP=1 << 'EOF'
+          do $$
+          declare
+            v_sample_ts int4;
+            v_wait_id smallint;
+            v_qid_a int4;
+            v_qid_b int4;
+            v_decoded_int_count int;
+            v_decoded_ts_count int;
+            v_decoded_int_total int;
+            v_decoded_ts_total int;
+            v_packed_total int;
+            v_distinct_datids int;
+            v_ts timestamptz;
+            v_slot smallint;
+          begin
+            -- Run a real take_sample() first so the call path exercises the
+            -- live sampler at least once.
+            perform ash.take_sample();
+
+            -- Build a deterministic test sample_ts so the assertions don't
+            -- depend on whether the CI runner happens to have active
+            -- backends right now. Use a sample_ts in the future to avoid
+            -- colliding with anything take_sample() inserted.
+            v_sample_ts := ash.ts_from_timestamptz(now()) + 86400;
+            v_slot := ash.current_slot();
+
+            -- Ensure dictionary entries exist (may already exist from earlier
+            -- decoder test step; on conflict do nothing).
+            insert into ash.wait_event_map (state, type, event)
+            values ('active', 'CPU*', 'CPU*') on conflict do nothing;
+            select id into v_wait_id from ash.wait_event_map
+            where state = 'active' and type = 'CPU*' and event = 'CPU*';
+
+            insert into ash.query_map_0 (query_id) values (777001) on conflict do nothing;
+            insert into ash.query_map_0 (query_id) values (777002) on conflict do nothing;
+            select id into v_qid_a from ash.query_map_0 where query_id = 777001;
+            select id into v_qid_b from ash.query_map_0 where query_id = 777002;
+
+            -- Insert two synthetic ash.sample rows for the same sample_ts
+            -- but different datids, so the int4 overload has more than one
+            -- ash.sample row to walk and the datid column is exercised.
+            -- datid 1 = template1, 2 = postgres in most installs, but we
+            -- only need oid values that are distinct and stable.
+            insert into ash.sample (sample_ts, datid, active_count, data, slot)
+            values
+              (v_sample_ts, 1::oid, 2, array[-v_wait_id, 2, v_qid_a, v_qid_b]::integer[], v_slot),
+              (v_sample_ts, 2::oid, 1, array[-v_wait_id, 1, v_qid_a]::integer[], v_slot);
+
+            -- decode_sample(int4): expect 3 decoded rows (2 + 1 backends),
+            -- each with count=1 and a datid that matches one of the rows we
+            -- inserted.
+            select count(*), coalesce(sum(d.count), 0), count(distinct d.datid)
+              into v_decoded_int_count, v_decoded_int_total, v_distinct_datids
+            from ash.decode_sample(v_sample_ts) d;
+            assert v_decoded_int_count = 3,
+              'decode_sample(int4) row count expected 3, got ' || v_decoded_int_count;
+            assert v_decoded_int_total = 3,
+              'decode_sample(int4) sum(count) expected 3, got ' || v_decoded_int_total;
+            assert v_distinct_datids = 2,
+              'decode_sample(int4) distinct datids expected 2, got ' || v_distinct_datids;
+
+            -- Sanity vs. packed active_count.
+            select coalesce(sum(active_count), 0) into v_packed_total
+            from ash.sample where sample_ts = v_sample_ts;
+            assert v_decoded_int_count = v_packed_total,
+              'decoded rows ' || v_decoded_int_count
+              || ' != packed active_count total ' || v_packed_total;
+
+            -- Verify wait_event + query_id resolution is correct for each
+            -- decoded row (uses the row's slot internally, so query_ids are
+            -- unambiguous).
+            assert exists (
+              select 1 from ash.decode_sample(v_sample_ts) d
+              where d.datid = 1::oid and d.wait_event = 'CPU*' and d.query_id = 777001
+            ), 'decode_sample(int4) missing expected (datid=1, CPU*, 777001) row';
+            assert exists (
+              select 1 from ash.decode_sample(v_sample_ts) d
+              where d.datid = 1::oid and d.wait_event = 'CPU*' and d.query_id = 777002
+            ), 'decode_sample(int4) missing expected (datid=1, CPU*, 777002) row';
+            assert exists (
+              select 1 from ash.decode_sample(v_sample_ts) d
+              where d.datid = 2::oid and d.wait_event = 'CPU*' and d.query_id = 777001
+            ), 'decode_sample(int4) missing expected (datid=2, CPU*, 777001) row';
+
+            -- timestamptz overload: round-trip and confirm we get identical
+            -- decoded rows.
+            v_ts := ash.ts_to_timestamptz(v_sample_ts);
+            select count(*), coalesce(sum(d.count), 0)
+              into v_decoded_ts_count, v_decoded_ts_total
+            from ash.decode_sample(v_ts) d;
+            assert v_decoded_ts_count = v_decoded_int_count,
+              'decode_sample(timestamptz) row count ' || v_decoded_ts_count
+              || ' != decode_sample(int4) ' || v_decoded_int_count;
+            assert v_decoded_ts_total = v_decoded_int_total,
+              'decode_sample(timestamptz) sum(count) ' || v_decoded_ts_total
+              || ' != decode_sample(int4) ' || v_decoded_int_total;
+
+            -- Empty input: a sample_ts with no rows must return zero rows.
+            assert (select count(*) from ash.decode_sample(-1)) = 0,
+              'decode_sample(int4) on missing sample_ts should return 0 rows';
+
+            -- Clean up the synthetic rows so later steps see a normal table.
+            delete from ash.sample where sample_ts = v_sample_ts;
+            delete from ash.query_map_0 where query_id in (777001, 777002);
+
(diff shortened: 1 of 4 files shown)


NikolayS:
# REV review — PR #58

**Verdict:** APPROVE WITH MINOR FOLLOW-UP

Reviewed `fd47464` (latest after rebase) against `main`. Closes #54. Adds two convenience overloads — `ash.decode_sample(int4)` and `ash.decode_sample_at(timestamptz)` — leaving the original `decode_sample(integer[], smallint)` untouched. The renaming to `_at` in the second commit cleanly resolves the `decode_sample(unknown)` overload-resolution ambiguity, and the `\i`-driven 1.3→1.4 upgrade path picks up both functions for free.

## Bug hunter

`NO_FINDINGS` for the new code paths. The LATERAL JOIN passes `s.slot` per row, which is the unambiguous-resolution branch in the existing 2-arg SRF — no risk of grabbing a stale id from another partition. Slot is read from each row, not from `current_slot()`, so result correctness survives a concurrent rotation between sample insertion and decode call (rotation truncates the slot row and its query_map together — lockstep — so the row simply disappears rather than dangling).

```
FINDING:
- severity: LOW
- confidence: 7
- file: sql/ash-install.sql
- line: 856-858 (decode_sample_at body)
- issue: pre-epoch / far-future timestamptz triggers int4 overflow in ts_from_timestamptz, surfacing as "integer out of range" rather than empty result
- evidence: ash.ts_from_timestamptz('1000-01-01'::timestamptz) computes ~-3.2e10 seconds, casting to int4 underflows
- fix: same class as #51/#63 — clamp via greatest(..., int4-min)::int4 in ts_from_timestamptz, or short-circuit decode_sample_at when input falls outside the [epoch, epoch+int4-max] horizon. Out of scope for this PR; track under the existing umbrella issue #63.
```

## Security reviewer

`NO_FINDINGS`.

- Both new functions are `language sql STABLE` with `set search_path = pg_catalog, ash` — consistent with the H-SEC-2 hardening from #45.
- The dynamic REVOKE-from-PUBLIC loop in `ash-install.sql` (lines 4554-4564) iterates `pg_proc` for `nspname = 'ash' AND prokind = 'f'` and revokes EXECUTE — picking up both new overloads automatically. No drift risk.
- Re-grant block (lines 4570-4572) only re-grants the ts helpers + `epoch()` to PUBLIC; `decode_sample(int4)` and `decode_sample_at(timestamptz)` correctly stay revoked. Confirmed: README documents the explicit-grant requirement (line 195 of README.md).
- No SECURITY DEFINER, no dynamic SQL, no string interpolation of user input.

## Test analyzer

```
FINDING:
- severity: MEDIUM
- confidence: 7
- line: 642 (insert into ash.query_map_0)
- issue: test hard-codes ash.query_map_0 but reads slot via ash.current_slot() for the synthetic sample row. Works because rotation tests run after this step, but a future re-ordering of CI steps would silently desync slot vs. dictionary partition and the assertions on (datid, wait_event, query_id) would fail with NULL query_id.
- untested_scenarios: insert into ash.query_map_<v_slot> via dynamic SQL keyed off current_slot(), so the test stays correct after re-ordering or if num_partitions ever changes
```

```
FINDING:
- severity: LOW
- confidence: 6
- line: 723-728 (decode_sample_at round-trip)
- issue: round-trip only verifies row count + sum(count) equality, not row-level (datid, wait_event, query_id) equality. A bug where decode_sample_at silently swapped datids between rows would still pass.
- untested_scenarios: assert that "select * from decode_sample_at(v_ts) except select * from decode_sample(v_sample_ts)" is empty (and vice versa)
```

```
FINDING:
- severity: LOW
- confidence: 6
- line: 731-732 (empty input test)
- issue: empty-input test only covers a sample_ts value with no matching rows (-1). It does not exercise NULL input — which on the new int4 overload should also return zero rows, but is not covered. The decoder corner-cases section (line 1595) explicitly casts NULL to integer[] to disambiguate, leaving NULL-as-int4 untested.
- untested_scenarios: assert (select count(*) from ash.decode_sample(null::int4)) = 0; same for decode_sample_at(null::timestamptz)
```

```
FINDING:
- severity: LOW
- confidence: 5
- line: 738 (regression assertion for #54)
- issue: assertion `count(*) >= 0` is always true — it tests overload resolution (the call doesn't error) but cannot catch a bug where the function returns wrong rows. Acceptable as a "did this typecheck" smoke, but a clearer assertion would be `= 0` for sample_ts=0 (no rows expected at the epoch boundary in CI).
- untested_scenarios: tighten to `= 0` so the regression catches both ambiguity AND wrong overload binding
```

`decode_sample_at` not exercised against pre-epoch or far-future timestamptz (would catch the bug-hunter LOW finding above), but that is reasonably out-of-scope for a convenience-overload PR.

## Docs reviewer

`NO_FINDINGS`. README Helpers table (lines 188-193) gains both functions with accurate signatures, the `EXECUTE`-revoke note matches the actual privilege state (line 195), and the new example uses `decode_sample_at` (consistent with the rename). `comment on function` strings on both new overloads (lines 4197-4200 of `ash-install.sql`) explain the slot-resolution behavior and the `_at` naming rationale.

## Guidelines checker

`NO_FINDINGS`. `language sql STABLE`, `set search_path = pg_catalog, ash`, lower-case keywords, snake_case names, `int4`/`int8` over `integer`/`bigint` — consistent with the project's SQL style and CLAUDE.md. Commit messages follow Conventional Commits (`feat:`, `fix(decode_sample):`, `fix(test):`). The 1.3-to-1.4 changelog entry was updated per the CLAUDE.md mirroring contract.

## CI

Latest CI run on `fd47464` is pending across PG 14/15/16/17 (on+off pgss)/18 at review time. CodeQL passed. Recommend confirming green before merge.

## Summary

Two narrowly-scoped convenience overloads. The implementation correctly threads `slot` through to the 2-arg SRF (the existing "search-all-partitions" branch is the documented hazardous one — this PR avoids it). The biggest open item is the timestamptz-overflow class (sibling of #51/#63), which is out of scope here but should be tracked. Test could be tightened (NULL/exact-row equality) but is solid for a convenience-overload PR.

---

NikolayS:
## Testing evidence — PR #58 end-to-end (post REV fix)

Commit under test: `e456b3b` on `claude/fix-issue-54-decode-sample-overloads`.
Environment: fresh `postgres:18-alpine` Docker container (PostgreSQL 18.3), `\i sql/ash-install.sql` from the PR branch. No pg_cron, no pg_stat_statements (degraded mode).

### REV findings addressed in `e456b3b`

| REV finding | Fix in commit `e456b3b` |
| --- | --- |
| MEDIUM (line 642): test hard-codes `ash.query_map_0` | Resolves the active slot via `ash.current_slot()` and uses `format('... ash.query_map_%s ...', v_slot)` for both inserts and the cleanup `delete`. |
| LOW (lines 723-728): timestamptz round-trip only checks count + sum | Adds two-way `EXCEPT` on `(datid, wait_event, query_id, count)` between `decode_sample(int4)` and `decode_sample_at(timestamptz)`. |
| LOW (lines 731-732): no NULL-as-int4 / NULL-as-timestamptz coverage | Adds `assert (select count(*) from ash.decode_sample(null::int4)) = 0` and the matching `decode_sample_at(null::timestamptz)` assertion. |
| LOW (line 738): regression assertion `count(*) >= 0` is always true | Tightened to `= 0` at the epoch boundary -- now catches wrong-overload binding, not just parse errors. |

Function bodies were **not** modified — diff is workflow-only (+64 / -11 in `.github/workflows/test.yml`).

### Test 1 — `decode_sample(integer[], smallint)` directly on a sample row

Two synthetic `ash.sample` rows (`datid=1` with two backends, `datid=2` with one) inserted at a future `sample_ts` in the active slot. The original 2-arg SRF, called once per row, yields 3 decoded backends:

```
=== Test 1: decode_sample(integer[], smallint) on a raw sample row ===
+-----------+-------+------+------------+----------+-------+
| sample_ts | datid | slot | wait_event | query_id | count |
+-----------+-------+------+------------+----------+-------+
|   9934355 |     1 |    0 | CPU*       |   777001 |     1 |
|   9934355 |     1 |    0 | CPU*       |   777002 |     1 |
|   9934355 |     2 |    0 | CPU*       |   777001 |     1 |
+-----------+-------+------+------------+----------+-------+
(3 rows)
```

### Test 2 — new `decode_sample(int4)` overload (walks all rows for the sample_ts)

```
=== Test 2: decode_sample(int4) -- new overload, walks all rows ===
+-------+------------+----------+-------+
| datid | wait_event | query_id | count |
+-------+------------+----------+-------+
|     1 | CPU*       |   777001 |     1 |
|     1 | CPU*       |   777002 |     1 |
|     2 | CPU*       |   777001 |     1 |
+-------+------------+----------+-------+
(3 rows)
```

### Test 3 — tuple-by-tuple equality between Test 1 and Test 2

Both directions of the symmetric difference are empty:

```
=== Test 3: tuple-by-tuple equality (Test 1 vs Test 2) ===
+--------------+-------+
|     diff     | count |
+--------------+-------+
| in t1 not t2 |     0 |
| in t2 not t1 |     0 |
+--------------+-------+
(2 rows)
```

### Test 4 — `decode_sample_at(timestamptz)` delegates to the int4 overload

```
=== Test 4: decode_sample_at(timestamptz) -- delegates to int4 overload ===
+-------+------------+----------+-------+
| datid | wait_event | query_id | count |
+-------+------------+----------+-------+
|     1 | CPU*       |   777001 |     1 |
|     1 | CPU*       |   777002 |     1 |
|     2 | CPU*       |   777001 |     1 |
+-------+------------+----------+-------+
(3 rows)
```

### Test 5 — row-level equality `decode_sample(int4)` vs `decode_sample_at(timestamptz)`

This is the new assertion class added per the REV LOW finding on line 723-728. `EXCEPT` in both directions confirms identical tuples:

```
=== Test 5: row-level equality decode_sample(int4) vs decode_sample_at(timestamptz) ===
+----------------+-------+
|      diff      | count |
+----------------+-------+
| in int4 not ts |     0 |
| in ts not int4 |     0 |
+----------------+-------+
(2 rows)
```

### Test 6 — `decode_sample_at(now() - interval '5 minutes')`

No sample was taken at that timestamp, so the overload correctly returns zero rows (does not error):

```
=== Test 6: decode_sample_at(now() - interval 5 minutes) ===
+---------------+
| rows_5min_ago |
+---------------+
|             0 |
+---------------+
(1 row)
```

### Test 7 — NULL inputs (new coverage)

Both NULL forms return zero rows, no error:

```
=== Test 7: NULL inputs return zero rows (no error) ===
+-------------------------------------+------+
|                call                 | rows |
+-------------------------------------+------+
| decode_sample(null::int4)           |    0 |
| decode_sample_at(null::timestamptz) |    0 |
+-------------------------------------+------+
(2 rows)
```

### Test 8 — Empty input + tightened #54 regression

`decode_sample(-1)` (no matching `sample_ts`) and `decode_sample(0)` (epoch boundary -- the tightened `= 0` regression assertion for #54) both return exactly 0 rows. The unadorned integer literal `0` resolves unambiguously to the `int4` overload:

```
=== Test 8: empty / non-matching inputs return zero rows ===
+--------------------------------------------------------+------+
|                          call                          | rows |
+--------------------------------------------------------+------+
| decode_sample(-1)                                      |    0 |
| decode_sample(0) -- epoch boundary, regression for #54 |    0 |
+--------------------------------------------------------+------+
(2 rows)
```

### Overload signatures observed (sanity)

```
| ash | decode_sample    | TABLE(wait_event text, query_id bigint, count integer)            | p_data integer[], p_slot smallint DEFAULT NULL::smallint |
| ash | decode_sample    | TABLE(datid oid, wait_event text, query_id bigint, count integer) | p_sample_ts integer                                      |
| ash | decode_sample_at | TABLE(datid oid, wait_event text, query_id bigint, count integer) | p_ts timestamp with time zone                            |
```

All three overloads present, each with the expected signature -- and `decode_sample(unknown)` resolves cleanly to the `int4` form (Test 8).

PR #62, opened 2026-04-24
fix: ash.status() gracefully handles denied cron.job access

### LINKED ISSUE #61: bug: ash.status() fails for non-superuser callers when pg_cron is loaded
## Summary
When `pg_cron` is loaded (`shared_preload_libraries = pg_cron`, `CREATE EXTENSION pg_cron`), `ash.status()` reads `cron.job` to surface sampling schedule. Non-superuser callers without `USAGE` on schema `cron` get a permission-denied error.

Surfaced while testing #52 (`ash.grant_reader`) — a role granted by `grant_reader` can call every reader but still fails on `status()` because `status()` touches an extension-owned schema outside ash's control.

## Impact
- Monitoring roles (Grafana, exporters) that use `ash.grant_reader` to read ash data still can't call `ash.status()`. They have to either be superuser or get USAGE on schema cron separately.
- Deployed pg_cron environments only.

## Fix options
- **A.** In `ash.status()`, wrap the `cron.job` lookup in a nested BEGIN/EXCEPTION that catches `insufficient_privilege` and substitutes a placeholder metric ("cron.job not readable by current role"). Preserves the current information density for callers with access, gracefully degrades otherwise.
- **B.** Have `ash.grant_reader` also issue `GRANT USAGE ON SCHEMA cron TO <role>` when pg_cron is detected. Bridges the permission gap but touches a schema pg_ash doesn't own.
- **C.** Split `ash.status()` into two functions: a lightweight core `status()` that doesn't touch cron, and a pg_cron-enhanced status that requires access. Breaking.

(A) looks cleanest. Candidate for v1.5.

## Detected by
v1.4 post-release validation. See #52's fix agent report.

NikolayS:
## Summary
- Wrap the `cron.job` lookup in `ash.status()` with a nested `BEGIN ... EXCEPTION WHEN insufficient_privilege ... END` so monitoring roles granted only `ash.*` readers can call `status()` without aborting when pg_cron is loaded but the role lacks `USAGE` on schema `cron`.
- On denial we emit a single informative fallback row keyed `cron_jobs` with value `<no cron.job access; grant USAGE ON SCHEMA cron to this role>` so operators see why cron details are missing rather than silent emptiness. Granted callers see the same per-job rows as before, preserving the public return shape.
- Mirrored a one-line note into `sql/ash-1.3-to-1.4.sql` per `CLAUDE.md`. Legacy upgrade scripts are not touched.

Fixes #61.

## Test plan
- [x] New CI step `Test #61: status() works for non-superuser without cron.job access` (gated on `matrix.cron == 'on'`, no-ops if pg_cron isn't actually installed for a given PG major). Creates `ash_status_reader`, grants ash readers, revokes `USAGE ON SCHEMA cron`, switches role, asserts `status()` returns rows, asserts `pg_cron_available=yes`, and asserts the `cron_jobs` fallback row is present with the documented wording.
- [x] Verified locally on PG 18 with pg_cron 1.6: denied path returns the fallback row and `status()` produces all 30 metric rows; granted path with a scheduled `ash_sampler` cron job still emits `cron_job_ash_sampler` exactly as before.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -2507,6 +2507,21 @@ jobs:
           revoke usage on schema cron from public;
           revoke usage on schema cron from ash_status_reader;
 
+          -- Precondition guard: if some test infra leaves USAGE granted
+          -- to ash_status_reader directly (or via a future pg_cron default
+          -- change), the denied-path test would silently pass without
+          -- exercising the fallback. Assert FALSE; otherwise REVOKE again.
+          do $$
+          begin
+            if has_schema_privilege('ash_status_reader', 'cron', 'USAGE') then
+              raise notice 'cron USAGE still granted to ash_status_reader — explicit revoke';
+              revoke usage on schema cron from ash_status_reader;
+              if has_schema_privilege('ash_status_reader', 'cron', 'USAGE') then
+                raise exception 'precondition failed: ash_status_reader still has USAGE on schema cron';
+              end if;
+            end if;
+          end $$;
+
           set role ash_status_reader;
 
           do $$
@@ -2533,10 +2548,56 @@ jobs:
                    and v_cron_marker like '%no cron.job access%',
               'expected denied-access fallback row, got: ' || coalesce(v_cron_marker, '<null>');
 
-            raise notice '#61 cron.job ACL test PASSED: status() ran without privilege error';
+            raise notice '#61 cron.job ACL test PASSED (denied path): status() ran without privilege error';
           end $$;
 
           reset role;
+
+          -- Sub-test: GRANT USAGE on schema cron and assert the role now
+          -- sees real cron_job_* rows instead of the cron_jobs fallback.
+          -- pg_cron RLS filters cron.job by `username = current_user`, so
+          -- the sentinel ash_sampler job has to be scheduled BY the
+          -- reader role (or its username altered) for status() to see it.
+          grant usage on schema cron to ash_status_reader;
+          grant select on cron.job to ash_status_reader;
+          grant execute on function cron.schedule(text, text, text) to ash_status_reader;
+
+          set role ash_status_reader;
+          do $$
+          begin
+            perform cron.schedule('ash_sampler', '1 second', 'select 1');
+          end $$;
+
+          do $$
+          declare
+            v_fallback text;
+            v_jobs int;
+          begin
+            -- Fallback row must NOT appear when the caller has access.
+            select value into v_fallback
+            from ash.status() where metric = 'cron_jobs';
+            assert v_fallback is null,
+              'expected no cron_jobs fallback row when granted, got: ' || coalesce(v_fallback, '<null>');
+
+            -- Real per-job rows must appear (sentinel ash_sampler scheduled above).
+            select count(*) into v_jobs
+            from ash.status()
+            where metric like 'cron_job_%';
+            assert v_jobs > 0,
+              'expected at least one cron_job_* row when granted, got 0';
+
+            raise notice '#61 cron.job ACL test PASSED (granted path): % cron_job_* rows visible', v_jobs;
+          end $$;
+
+          do $$
+          begin
+            perform cron.unschedule('ash_sampler');
+          end $$;
+          reset role;
+          revoke execute on function cron.schedule(text, text, text) from ash_status_reader;
+          revoke select on cron.job from ash_status_reader;
+          revoke usage on schema cron from ash_status_reader;
+
           revoke execute on function ash._pg_cron_available() from ash_status_reader;
           revoke execute on function ash._color_on(boolean) from ash_status_reader;
           revoke select on table ash.rollup_1h from ash_status_reader;
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -2243,6 +2243,79 @@ jobs:
           drop role ash_test_user;
           EOF
 
+      - name: "Test #61: status() works for non-superuser without cron.job access"
+        if: matrix.cron == 'on'
+        env:
+          PGPASSWORD: postgres
+        run: |
+          # Skip cleanly if pg_cron isn't actually installed (e.g. PGDG
+          # package missing on a new PG major); the cron=on matrix entry
+          # tolerates that.
+          HAS_CRON=$(psql -h localhost -U postgres -d postgres -tAX -c \
+            "select exists (select from pg_extension where extname='pg_cron')")
+          if [ "$HAS_CRON" != "t" ]; then
+            echo "pg_cron not installed — skipping #61 cron.job ACL test"
+            exit 0
+          fi
+
+          psql -h localhost -U postgres -d postgres -v ON_ERROR_STOP=1 << 'EOF'
+          -- Reproduce the bug: a role granted ash readers but NOT
+          -- USAGE on schema cron used to get insufficient_privilege from
+          -- ash.status() because of the unguarded `from cron.job` lookup.
+          do $$
+          begin
+            if not exists (select from pg_roles where rolname = 'ash_status_reader') then
+              create role ash_status_reader login;
+            end if;
+          end $$;
+
+          grant usage on schema ash to ash_status_reader;
+          grant execute on function ash.status() to ash_status_reader;
+          -- Deliberately deny USAGE on schema cron (pg_cron grants it to
+          -- PUBLIC by default — revoke from public + the role to ensure
+          -- the test role really lacks access).
+          revoke usage on schema cron from public;
+          revoke usage on schema cron from ash_status_reader;
+
+          set role ash_status_reader;
+
+          do $$
+          declare
+            v_rows int;
+            v_cron_avail text;
+            v_cron_marker text;
+          begin
+            -- Must not raise insufficient_privilege.
+            select count(*) into v_rows from ash.status();
+            assert v_rows > 0, 'status() returned no rows for monitoring role';
+
+            -- Sanity: pg_cron_available row still emitted.
+            select value into v_cron_avail
+            from ash.status() where metric = 'pg_cron_available';
+            assert v_cron_avail = 'yes',
+              'expected pg_cron_available=yes, got: ' || coalesce(v_cron_avail, '<null>');
+
+            -- Caller is denied — must get the informative fallback row,
+            -- not silent emptiness.
+            select value into v_cron_marker
+            from ash.status() where metric = 'cron_jobs';
+            assert v_cron_marker is not null
+                   and v_cron_marker like '%no cron.job access%',
+              'expected denied-access fallback row, got: ' || coalesce(v_cron_marker, '<null>');
+
+            raise notice '#61 cron.job ACL test PASSED: status() ran without privilege error';
+          end $$;
+
+          reset role;
+          revoke execute on function ash.status() from ash_status_reader;
+          revoke usage on schema ash from ash_status_reader;
+          drop role ash_status_reader;
+          -- Restore pg_cron's default PUBLIC USAGE so subsequent test
+          -- steps that read cron.job as superuser-but-with-search_path
+          -- aren't affected (superuser bypasses ACLs, but be safe).
+          grant usage on schema cron to public;
+          EOF
+
       - name: "Test v1.4: rollup_cleanup retention"
         env:
           PGPASSWORD: postgres
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -2271,6 +2271,18 @@ jobs:
 
           grant usage on schema ash to ash_status_reader;
           grant execute on function ash.status() to ash_status_reader;
+          -- PR #45 hardening REVOKE'd readers/helpers from PUBLIC, so the
+          -- test role needs explicit grants on every object ash.status()
+          -- touches (besides ash.epoch / ash.ts_to_timestamptz, which are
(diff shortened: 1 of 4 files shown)


NikolayS:
## REV review — PR #62

**Verdict:** APPROVE with minor observations. The fix is small, correct, well-scoped, and minimally invasive. Catching the narrow `insufficient_privilege` exception (rather than `others`) is the right call: the only realistic failure mode inside the inner block once `_pg_cron_available()` returned true is an ACL denial on schema `cron` / table `cron.job`. The CI test exercises the real denied path and asserts both that `status()` does not abort and that the documented fallback wording is emitted. CI was still in-flight at review time but `cron=off` and `cron=on` paths in production code are both reachable cleanly.

---

### bug-hunter

`NO_FINDINGS` (high-confidence bugs).

Considered and dismissed:

- **Catching only `insufficient_privilege` vs `others`** — this is correct, not a bug. `_pg_cron_available()` already confirmed `pg_cron` is in `pg_extension`, which guarantees schema `cron` and table `cron.job` exist. Any other exception (e.g. `query_canceled`, OOM) inside the loop should propagate to the caller — swallowing them with `when others` would be wrong. Good defensive scoping.
- **Schema `cron` doesn't exist at all** — unreachable: if pg_cron is in `pg_extension`, the schema is owned by the extension and present. The early-return `else` branch handles the not-installed case.
- **`cron=off` matrix path** — unchanged: `_pg_cron_available()` returns false, the entire new `begin/exception` block is skipped, and the existing `pg_cron_available = no (use external scheduler)` row is emitted. No regression.

### test-analyzer

```
FINDING:
- severity: LOW
- confidence: 6
- line: 2464-2552
- issue: Test does not assert that the GRANTED path still emits per-job rows
- untested_scenarios: After the denied-path assertion, the same step does
  not flip back to a granted role and verify that `cron_job_ash_sampler`
  etc. still appear when a scheduled job exists. The PR description claims
  this was verified manually, but no CI guardrail prevents a future
  refactor from breaking the granted shape.
- suggestion: Optional — schedule a no-op `ash_sampler` cron job (or use
  the existing one if `ash.start()` ran earlier in the matrix), then
  `set role` to a role *with* USAGE on schema cron and assert at least
  one `cron_job_*` row is returned. Non-blocking; the existing
  install/upgrade tests likely cover the granted path implicitly via
  superuser invocations elsewhere in `test.yml`.
```

```
FINDING:
- severity: LOW
- confidence: 5
- line: 2493-2497
- issue: Test does not pin pg_cron's default schema-USAGE grant before revoking
- untested_scenarios: The test assumes pg_cron's install grants USAGE on
  schema `cron` to PUBLIC. If a future pg_cron release changes this
  default (or PGDG packaging differs), `revoke usage on schema cron from
  public` becomes a no-op and the denial path may not actually be
  exercised — the assertions could still pass coincidentally if
  `ash_status_reader` never had USAGE in the first place (which is
  exactly what we want anyway, but for the wrong reason).
- suggestion: Either explicitly `grant usage on schema cron to
  ash_status_reader` *before* the revoke (to make the denial path
  deterministic), or assert the precondition with
  `select has_schema_privilege('ash_status_reader', 'cron', 'usage')`.
  Non-blocking — the current test does fail closed against today's
  pg_cron behavior.
```

The happy path (denied access → fallback row + `pg_cron_available = yes` + `count(*) > 0`) is fully covered with specific-value assertions. Each assertion checks an exact value, not just existence. The `cron=off` matrix entry doesn't run this step, but production code is unchanged on that path so the existing matrix coverage is sufficient.

### security-reviewer

`NO_FINDINGS`.

The change is privilege-decreasing (allows callers who would have errored to get a fallback row instead). No new dynamic SQL, no `security definer` changes, no new search_path concerns — `ash.status()` retains `set search_path = pg_catalog, ash` from the existing definition. The fallback row's value is a static literal, not user-controlled. The test's `revoke usage on schema cron from public` is restored via `grant usage on schema cron to public` in teardown, so subsequent matrix steps see pg_cron's default ACLs.

### guidelines-checker

```
FINDING:
- severity: INFO
- confidence: 4
- file: sql/ash-install.sql
- line: 1727-1746
- issue: Multi-line explanatory comment uses `--` line comments rather
  than the `/* ... */` block-comment form recommended by the SQL style
  guide for multi-line descriptions
- rule: development__db-sql-style-guide.mdc — "Block comments for
  multi-line descriptions; line comments for single line notes"
- suggestion: Cosmetic only. The surrounding file uses `--` for
  multi-line notes consistently (see lines 1714-1719 immediately
  above), so per the style guide's own "consistency within one
  module is most important" clause, the existing pattern wins.
  No change needed.
```

Otherwise consistent with project conventions: lowercase keywords, snake_case identifiers, indentation matches surrounding code, the `cron_jobs` metric key follows the existing `cron_job_*` family naming with a deliberate plural to disambiguate the fallback from real per-job rows.

### docs-reviewer

```
FINDING:
- severity: LOW
- confidence: 6
- file: sql/ash-install.sql
- line: 1742-1745
- issue: Fallback wording is functional but slightly cryptic for
  copy-paste remediation
- suggestion: The current message
  `<no cron.job access; grant USAGE ON SCHEMA cron to this role>`
  is readable but a human doesn't immediately know *which* role
  "this role" refers to (the role calling `status()`). Consider
  embedding `current_user` so the instruction is fully
  self-contained, e.g.
  `format('<no cron.job access — run: grant usage on schema cron to %I>', current_user)`.
  Non-blocking; existing wording is acceptable and the metric key
  `cron_jobs` (plural) is a nice disambiguator from real
  `cron_job_<name>` rows.
```

```
FINDING:
- severity: LOW
- confidence: 5
- file: README.md / sql/ash-1.3-to-1.4.sql
- line: README around the grant_reader() example
- issue: Upgrade-script note for #61 is present, but the README's
  `ash.grant_reader()` documentation is not updated to mention that
  cron.job visibility requires a separate manual grant
- suggestion: `ash.grant_reader()` does NOT grant USAGE on schema cron
  (intentionally — pg_cron isn't an `ash` object), so users following
  the README to set up a Grafana role will hit this fallback row
  immediately. A one-line note in the README pointing out that the
  `cron_jobs` row is expected unless USAGE on schema cron is granted
  separately would close the loop. Non-blocking — the fallback row
  itself is now self-documenting.
```

Code comment at lines 1729-1734 of the diff is excellent — clearly states *why* the exception handler exists and references issue #61. Commit messages follow Conventional Commits and reference #61.

---

### Watch-list answers

- **Does the nested EXCEPTION swallow OTHER errors?** No — only `insufficient_privilege` is caught, which is the right scope. Confirmed in the diff for `sql/ash-install.sql`. Catching `when others` would have been a real bug.
- **Fallback wording readable?** Mostly yes; minor LOW-confidence suggestion above to embed `current_user` for fully-self-contained copy-paste.
- **Does the test actually exercise the denied path?** Yes — the test revokes USAGE from PUBLIC and from the role explicitly, so `ash_status_reader` has no path to schema `cron`. Caveat: relies on pg_cron's current default of granting USAGE to PUBLIC; a `has_schema_privilege` precondition check would harden it (LOW).
- **`cron=off` matrix path?** Unaffected. The new `begin/exception` block lives inside `if ash._pg_cron_available() then ...` and is unreachable when pg_cron isn't installed. The `cron=off` matrix is gated out of the new test step via `if: matrix.cron == 'on'`.
- **Test grants minimal?** Yes. Each grant is justified by a concrete reference inside `ash.status()`: `config` (line 1621), `sample` (line 1624), `wait_event_map` (1632), `query_map_all` (1635), `rollup_1m` / `rollup_1h` (rollup metrics), `_color_on` (1652), `_pg_cron_available` (1725). None appear superfluous. Worth noting: `ash.grant_reader()` already exists as a one-shot helper that would collapse ~10 lines of explicit grants/revokes into one call — but using it would mask exactly which objects are needed, so the explicit approach is defensible for a regression test specifically about ACL surface.

---

NikolayS:
## Testing evidence — REV nits applied

Reproduces the denied → granted → revoked transitions end-to-end on `postgres:18` with `pg_cron 1.6`, against the latest commit on `claude/fix-issue-61-status-cron-access`.

**Environment**

- `PostgreSQL 18.1 (Debian 18.1-1.pgdg13+2)`
- `pg_cron 1.6` (PGDG), `shared_preload_libraries = 'pg_cron,pg_stat_statements'`, `cron.use_background_workers = on`, `cron.database_name = postgres`
- pg_ash installed via `\i sql/ash-install.sql` from the PR branch (with the REV nit fixes)

### 1. Install pg_ash from PR branch

```text
$ docker exec ... psql -U postgres -v ON_ERROR_STOP=1 -f /tmp/ash-install.sql | tail -5
CREATE FUNCTION
COMMENT
CREATE FUNCTION
COMMENT
DO
```

### 2. Create role `reader`; grant via `ash.grant_reader`

```text
postgres=# create role reader login password 'r';
CREATE ROLE
postgres=# select ash.grant_reader('reader');
NOTICE:  ash.grant_reader: granted USAGE on schema ash, EXECUTE on 42 reader function(s), SELECT on 12 table(s) to reader
postgres=# revoke usage on schema cron from public;
REVOKE
postgres=# revoke usage on schema cron from reader;
REVOKE
postgres=# select has_schema_privilege('reader','cron','USAGE');
 has_schema_privilege
----------------------
 f
```

### 3. SET ROLE reader; `ash.status()` — expect `cron_jobs` fallback

```text
postgres=# set role reader;
postgres=> select current_user;
 current_user
--------------
 reader
postgres=> select metric, value from ash.status()
postgres-> where metric in ('pg_cron_available','cron_jobs') or metric like 'cron_job_%';
+-------------------+------------------------------------------------------------+
|      metric       |                           value                            |
+-------------------+------------------------------------------------------------+
| pg_cron_available | yes                                                        |
| cron_jobs         | <no cron.job access; grant USAGE ON SCHEMA cron TO reader> |
+-------------------+------------------------------------------------------------+
(2 rows)
```

The fallback row now embeds `current_user` via `format()` — a copy-paste of the printed grant statement works as-is for the calling role. (Nit 3.)

### 4. RESET ROLE; grant USAGE on cron + SELECT on `cron.job`; SET ROLE; `status()` — expect `cron_job_*` rows

```text
postgres=# reset role;
postgres=# grant usage on schema cron to reader;
postgres=# grant select on cron.job to reader;
postgres=# grant execute on function cron.schedule(text, text, text) to reader;
postgres=# set role reader;
postgres=> -- pg_cron RLS filters cron.job by `username = current_user`,
postgres=> -- so the sentinel job is scheduled BY reader for status() to see it.
postgres=> select cron.schedule('ash_sampler', '1 second', 'select 1');
 schedule
----------
        5
postgres=> select metric, value from ash.status()
postgres-> where metric in ('pg_cron_available','cron_jobs') or metric like 'cron_job_%';
+----------------------+-----------------------------------+
|        metric        |               value               |
+----------------------+-----------------------------------+
| pg_cron_available    | yes                               |
| cron_job_ash_sampler | id=5, schedule=1 second, active=t |
+----------------------+-----------------------------------+
(2 rows)
```

The fallback row is gone; the real `cron_job_ash_sampler` row is emitted. (Nit 1, granted-path coverage now in CI.)

### 5. Revoke; repeat — expect fallback returns

```text
postgres=# select cron.unschedule(5);
postgres=# revoke execute on function cron.schedule(text, text, text) from reader;
postgres=# revoke select on cron.job from reader;
postgres=# revoke usage on schema cron from reader;
postgres=# set role reader;
postgres=> select metric, value from ash.status()
postgres-> where metric in ('pg_cron_available','cron_jobs') or metric like 'cron_job_%';
+-------------------+------------------------------------------------------------+
|      metric       |                           value                            |
+-------------------+------------------------------------------------------------+
| pg_cron_available | yes                                                        |
| cron_jobs         | <no cron.job access; grant USAGE ON SCHEMA cron TO reader> |
+-------------------+------------------------------------------------------------+
(2 rows)
```

The transition is fully symmetric.

### CI test step (full simulation)

The updated `Test #61` step in `.github/workflows/test.yml` was simulated end-to-end against the same DB and exits clean:

```text
NOTICE:  #61 cron.job ACL test PASSED (denied path): status() ran without privilege error
NOTICE:  #61 cron.job ACL test PASSED (granted path): 1 cron_job_* rows visible
=== EXIT 0 ===
```

This now exercises both REV nit 1 (granted-path assertion) and nit 2 (precondition `has_schema_privilege` guard) in addition to the original denied-path coverage.

### Nits addressed

1. **Granted-path sub-test** — added; schedules a sentinel `ash_sampler` under the reader role (pg_cron RLS filters `cron.job` by `username = current_user`, so the job has to be scheduled by the calling role for it to be visible) and asserts at least one `cron_job_*` row, no `cron_jobs` fallback. Sentinel + grants are revoked at end.
2. **Precondition `has_schema_privilege`** — added; if `ash_status_reader` somehow still has USAGE on schema cron (future pg_cron default change, etc.), the test explicitly REVOKEs again and `raise exception` if the revoke didn't take effect, so the denied path can never silently no-op.
3. **Fallback wording embeds `current_user` via `format()`** — `format('<no cron.job access; grant USAGE ON SCHEMA cron TO %I>', current_user)`. The printed message is now self-contained and copy-pasteable for the calling role.
4. **README `Privileges` section** — added a note that monitoring roles will see the `cron_jobs` fallback row when pg_cron is loaded but the role lacks USAGE on schema `cron`, with the two remediation paths (`grant usage on schema cron to <role>` or ignore the row).

PR #64, opened 2026-04-24
demos: add animated GIF demo of v1.4 investigation flow

NikolayS:
## Summary

Adds a tmux-driven recording pipeline and produces a 300 KB animated GIF of pg_ash v1.4 catching a row-lock spike.

- Scenario: baseline pgbench → row-lock spike → investigation via \`ash.top_waits\` → \`ash.event_queries\` → \`ash.query_waits\`
- Follows the [LLM-assisted investigation](../README.md#llm-assisted-investigation) narrative
- Runs against vanilla \`postgres:18-alpine\`
- Reproducible: \`make record\`, no custom images

README.md embeds \`demos/ash_demo.gif\` as the hero visual under the tagline.

## Contents

| File | Purpose |
|---|---|
| \`demos/Makefile\` | \`make record\` / \`make clean\` |
| \`demos/record.sh\` | tmux orchestration |
| \`demos/workload.sh\` | pgbench + lock spike |
| \`demos/container-entrypoint.sh\` | Docker container setup |
| \`demos/README.md\` | prerequisites, usage |
| \`demos/ash_demo.gif\` | 300 KB, README embed |
| \`demos/ash_demo.cast\` | 14 KB asciinema source |

## Test plan

- [ ] \`make -C demos record\` produces a new GIF reproducibly
- [ ] GIF renders correctly in GitHub's README
- [ ] File size stays well under the 3 MB README budget

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #68, opened 2026-04-25
demos: human-paced typing, wider terminal, visible colors

NikolayS:
## Summary

Three follow-ups to the demo merged in #64, addressing user feedback on the recorded GIF:

1. **Human-paced typing.** `record.sh` now drives the tmux pane character-by-character via a new `human_type_and_send` helper, with 30–120 ms per-keystroke jitter and a longer beat at clause-boundary punctuation (`, ; . ( )`). Replaces the previous instant `tmux send-keys` calls so the recording reads like a real session at the keyboard. Bash's `RANDOM` is reseeded from `/dev/urandom` so each take is unique.

2. **Wider terminal + smaller font.** Geometry bumped from `100×28` to `140×32` so long wait-event names like `Lock:transactionid` and `Client:ClientRead` plus the colored bar charts fit on a single line without wrapping. Compensated by `agg --font-size 12` (was `16`) so the rendered GIF still embeds cleanly on GitHub (~1022 px wide vs the previous ~800 px). Splash banner resized to match the wider geometry.

3. **Visible colors.** The session now `set ash.color = on` in `psqlrc`, every reader call passes `p_color => true` (or the positional `true` argument), and a `:color` psql variable (mirroring the main README's pattern, minus `less -R` since the recorder can't drive an interactive pager) re-runs each query through `sed` to convert psql's literalised `\x1B` back into real ESC bytes. The resulting GIF shows colored bars in `top_waits`, `timeline_chart`, `event_queries`, and `query_waits`.

Also bumps the workload's `SPIKE_SEC` default from 30 s to 120 s (overrideable) so the spike outlives the now-longer (~90 s) recording, ensuring the final `\gset` → `query_waits` step still sees fresh `Lock:tuple` samples in its 1-minute window.

### Result

- `demos/ash_demo.cast`: 17 KB → 39 KB (994 output events vs ~16, since per-char typing produces per-char echo)
- `demos/ash_demo.gif`: 407 KB → 411 KB at 1022×554 (well under the 3 MB README budget)
- 8 truecolor ANSI escape events in the cast (splash banner + 4 colored reader outputs)
- 0 errors in the recorded SQL session

### Files changed

- `demos/record.sh` — `human_type_and_send`, wider geometry, font-size knob, `:color` psql var, workload phase env-var passthrough
- `demos/README.md` — updated tuning-knob table (new geometry / font / typing-pacing vars), new "Typing pacing" section, refreshed "Design notes"
- `demos/ash_demo.cast`, `demos/ash_demo.gif` — re-recorded with the above

Demo-only — `install.sql` and other production code are untouched.

## Test plan

- [x] `make -C demos record` completes cleanly with the spike, 0 errors, all 4 reader queries returning data with colored bars
- [x] cast file contains real ANSI truecolor escape codes (`\u001b[38;2;...m`) in `top_waits` / `timeline_chart` / `event_queries` / `query_waits` rows
- [x] GIF size under 3 MB; embeds cleanly at GitHub's ~800 px width
- [ ] Visual review of the rendered GIF on the PR page (reviewer)
- [ ] (Optional) Override knobs work: `TYPE_MIN_MS=10 TYPE_MAX_MS=40 make record` produces a faster take; `WARMUP_SEC=40 SPIKE_SEC=150 make record` produces a longer one


NikolayS:
# REV review — PR #68 (demos: human-paced typing, wider terminal, visible colors)

Scope: `demos/` only (record.sh, README.md, ash_demo.cast, ash_demo.gif). No production code touched. Reviewed against `~/.claude/rev/agents/{security,bug-hunter,docs,guidelines,test}.md` and the SQL/shell rule packs in `~/.claude/rev/rules/rules/`. SOC2 omitted per request.

## Verdict

APPROVE with minor cosmetic notes. The change is a self-contained demo polish: the new `human_type_and_send` helper is the right level of abstraction (single function, ~40 lines including comments), the `tmux send-keys -l --` invocation is correctly hardened against tmux command parsing, and the recorded cast verifies the trailing-`;` smuggling works for every SQL statement in the script. Visible colors land (truecolor escapes verified in the cast and in extracted GIF frames). GIF size is 411 KB at 1022×554, well under the 3 MB README budget.

## Security review

`human_type_and_send` walks a developer-authored string char-by-char and feeds each byte through `tmux send-keys -t "$SESSION" -l -- "$send_str"`. The `-l` (literal) flag plus `--` end-of-options separator means tmux performs no interpretation of the byte. There is no attacker-controlled input — every string passed to the helper is a hardcoded literal in `record.sh`, and the script runs locally against a Docker container the user just started. No injection vector.

The new psqlrc heredoc deserves a careful read because the quoting is non-trivial: it lives inside `bash -c "..."` (double-quoted) wrapping a `<<'PSQLRC'` heredoc (single-quoted EOF). The `\\\\\\\\x1B` six-backslash sequence is correct given that escape interpretation happens once in the outer double-quoted bash string and once in the inner sed argument — the recorded cast confirms `:color` resolves to `\g | sed 's/\\x1B/\x1b/g'`. Functional but fragile; any future edit to this block will need the same care.

`set ash.color = on` is a per-session GUC read by `_wait_color()` in `sql/ash-install.sql` — not an unsafe default. The psqlrc only lives inside the ephemeral demo container.

NO_FINDINGS.

## Bug hunter

NO_FINDINGS at the runtime-blocking level. Two minor notes (below confidence threshold for blocking, included for completeness):

- The per-character slice `${text:$i:1}` is byte-indexed in C locale and char-indexed in UTF-8 locale. All strings in `record.sh` are pure ASCII, so this is fine in practice; flagging only because anyone copying this helper into a context with Unicode prompts would get surprising results.
- `human_type_and_send` does not handle Tab, Esc, or any control character — but the demo strings contain none, and adding such handling would be premature.

## Shell-style / guidelines

`set -Eeuo pipefail` retained. `trap cleanup EXIT INT TERM` retained. Quoting of `$SESSION`, `$CAST`, `$CONTAINER` is consistent. `require()` gating retained for `docker tmux asciinema agg`. The `send_instant` helper kept "for debugging" is technically dead code (line 264-267); fine to leave but flag for cleanup if you want zero-dead-code hygiene.

## Docs review

`demos/README.md` accurately reflects the new defaults: 140×32 geometry, font-size 12, the new `TYPE_MIN_MS`/`TYPE_MAX_MS`/`TYPE_PUNCT_MS`/`AGG_FONT_SIZE`/`TAIL_SEC` envs, and the `SPIKE_SEC` 30→120 bump with the rationale spelled out. New "Typing pacing" section is well-pitched. Design notes correctly point to font-size 12 in the standalone agg invocation. No drift between env-var defaults declared in `record.sh` and the table in README.md.

## CI / test impact

`SPIKE_SEC` 30→120 is a recorder runtime knob, not exercised by `.github/workflows/test.yml`. Pure local impact — adds ~90 s to `make record`, no CI cost.

## GIF visual re-assessment: 8/10

Extracted frames at n=5/200/400/600/750:

- Splash banner: rendered in green (`\033[38;2;080;250;123m` truecolor), no typo, "Investigation: something just spiked..." line intact.
- `top_waits`: red/orange/yellow/green colored bars all visible, all wait-event names fit on one line (`Lock:transactionid`, `Client:ClientRead` no wrap).
- `timeline_chart`: stacked-bar legend + 6 buckets render cleanly, colored.
- `event_queries` + `query_waits`: red bars present, query_id and `update pgbench_accounts set abalance = aba` truncation visible as designed.
- Investigation arc Q1→Q4 reads clearly with `\echo` banners between each step.

One cosmetic snag (-2 from a perfect 10): the very first query line — `select metric, value from ash.status() where metric in ('version','sampling_enabled','sample_interval','samples_total','pg_cron_available');` — wraps at the right edge of the 140-col terminal (`pg_cron_avail` → `able`). Subsequent reader queries fit. Not a blocker; either drop one of the columns from the IN-list, or rephrase as a shorter query, if you want a perfectly clean opening shot.

## Bloat check

`record.sh` 281 → 377 lines (+96). Breakdown of additions: ~10 lines new env-var defaults, ~40 lines `human_type_and_send` (~half comments), ~15 lines extra psqlrc hardening, ~10 lines splash banner widening, balance is per-call updates. Single function as intended. Not bloated.

## Recommendations (non-blocking)

1. Consider dropping the unused `send_instant` helper (lines 264-267) or wiring it to a `TYPE_INSTANT=1` env override.
2. Consider trimming the opening `ash.status()` IN-list so the first command doesn't wrap at 140 cols.
3. The psqlrc heredoc quoting (`\\\\\\\\x1B`) deserves a one-line `# six-backslash because: outer bash double-quote → inner sed single-quote` comment so the next editor doesn't have to derive it.

---

PR #70, opened 2026-04-25
ci: include pg_constraint in schema-equivalence diff

### LINKED ISSUE #66: ci: extend schema-equivalence to include pg_constraint
## Background
Schema-equivalence CI compares functions, columns, indexes, and views between fresh-install and upgrade-chain DBs, but does NOT diff `pg_constraint`. Issue #49 (sample_data_check divergence) hid in this gap until a manual test caught it.

## Proposal
Extend the schema-equivalence step in `.github/workflows/test.yml` to:
- Enumerate every CHECK constraint on `ash.*` tables (parent + partitions) via `pg_constraint`.
- Compute `pg_get_constraintdef(oid)` for each.
- Diff between fresh and upgrade-chain DBs.

A divergence should fail CI with a clear message naming the constraint.

## Detected by
REV review on PR #56 — see https://github.com/NikolayS/pg_ash/pull/56#issuecomment-4316837923

NikolayS:
## Summary
- Extends the `Schema equivalence: fresh install vs full upgrade path` step in `.github/workflows/test.yml` to include `pg_constraint` rows for `ash.*` tables (parent partitioned tables and their partitions) alongside the existing function / column / index / view diffs.
- Snapshot emits `CON|<relname>|<conname>|<pg_get_constraintdef(oid)>` rows sorted by `(relname, conname)` so any divergence (e.g. the `sample_data_check >= 2` vs `>= 3` regression hidden in #49) lands cleanly in the unified diff.
- On divergence the step now also prints a dedicated `Constraint divergence detected (issue #66)` error that pairs `fresh` vs `upgrade` definitions per `(relname, conname)`, including `<missing>` for one-sided drops/adds, so CHECK-constraint mistakes get a clear, named failure instead of getting buried in a long combined diff.
- No changes to `sql/ash-install.sql` or any upgrade script. Existing schema-equivalence checks are preserved.

Fixes #66.

## Test plan
- [x] YAML still parses cleanly (`python3 -c "import yaml; yaml.safe_load(open('.github/workflows/test.yml'))"`).
- [x] Awk pairing logic verified locally against four scenarios:
  - definition diff (`sample_data_check` `>= 2` vs `>= 3`) prints both sides.
  - dropped on upgrade side prints `upgrade ... <missing>` + the fresh definition.
  - extra on upgrade side prints the upgrade definition + `fresh ... <missing>`.
  - identical inputs produce no output (no false positives).
- [ ] CI run on this PR passes the schema-equivalence step on PG 14/15/16/17/18 with `cron=on/off` (no real divergence in current install/upgrade chain).

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -3416,6 +3416,18 @@ jobs:
           from pg_views
           where schemaname = 'ash'
           order by viewname;
+
+          -- Issue #66: include CHECK / NOT NULL / PK / UNIQUE / FK constraints.
+          -- Covers parent partitioned tables AND their partitions (relkind r,p).
+          -- pg_get_constraintdef gives a stable canonical text form, so any
+          -- divergence (e.g. sample_data_check `>= 2` vs `>= 3` from #49)
+          -- surfaces here.
+          select 'CON', c.relname, con.conname, pg_get_constraintdef(con.oid)
+          from pg_constraint con
+          join pg_class c on c.oid = con.conrelid
+          join pg_namespace n on n.oid = c.relnamespace
+          where n.nspname = 'ash' and c.relkind in ('r','p')
+          order by c.relname, con.conname;
           SNAPSQL
 
           # 1) Fresh install → snapshot
@@ -3437,6 +3449,49 @@ jobs:
             echo "Schema equivalence test PASSED: fresh install and upgrade path produce identical schemas"
           else
             echo "::error::Schema equivalence test FAILED: fresh install and upgrade path differ (see diff above)"
+
+            # Issue #66: when constraints diverge, surface a clear, dedicated
+            # error naming each affected constraint and both definitions, so
+            # CHECK-constraint regressions (e.g. sample_data_check `>= 2` vs
+            # `>= 3`) don't get lost in a long unified diff over
+            # functions/columns/indexes/views.
+            grep '^CON|' /tmp/fresh_schema.txt    | sort > /tmp/fresh_constraints.txt
+            grep '^CON|' /tmp/upgraded_schema.txt | sort > /tmp/upgraded_constraints.txt
+            if ! diff -q /tmp/fresh_constraints.txt /tmp/upgraded_constraints.txt > /dev/null; then
+              echo "::error::Constraint divergence detected (issue #66):"
+              # Each CON line is: CON|<relname>|<conname>|<def>
+              # Pair fresh vs upgrade by (relname, conname) and report both
+              # definitions for any (relname, conname) present on only one
+              # side or whose definition differs.
+              awk -F'|' '
+                FNR == NR { fresh[$2"|"$3] = $0; next }
+                {
+                  key = $2"|"$3
+                  if (!(key in fresh)) {
+                    def = $4; for (i = 5; i <= NF; i++) def = def "|" $i
+                    printf "  upgrade ash.%s.%s: %s\n", $2, $3, def
+                    printf "  fresh   ash.%s.%s: <missing>\n", $2, $3
+                  } else if (fresh[key] != $0) {
+                    def = $4; for (i = 5; i <= NF; i++) def = def "|" $i
+                    printf "  upgrade ash.%s.%s: %s\n", $2, $3, def
+                    n = split(fresh[key], a, "|")
+                    def2 = a[4]; for (i = 5; i <= n; i++) def2 = def2 "|" a[i]
+                    printf "  fresh   ash.%s.%s: %s\n", a[2], a[3], def2
+                    delete fresh[key]
+                  } else {
+                    delete fresh[key]
+                  }
+                }
+                END {
+                  for (k in fresh) {
+                    n = split(fresh[k], a, "|")
+                    def = a[4]; for (i = 5; i <= n; i++) def = def "|" a[i]
+                    printf "  upgrade ash.%s.%s: <missing>\n", a[2], a[3]
+                    printf "  fresh   ash.%s.%s: %s\n", a[2], a[3], def
+                  }
+                }
+              ' /tmp/fresh_constraints.txt /tmp/upgraded_constraints.txt
+            fi
             exit 1
           fi
 


NikolayS:
## REV review: PR #70

**Verdict:** APPROVE — small, well-scoped CI improvement that closes the exact gap that hid the `sample_data_check` regression in #49/#56. All 5 PG matrix jobs and `cron=off/on` lanes are green. No SQL surface changed; only `.github/workflows/test.yml` is touched.

---

### bug-hunter

`NO_FINDINGS` of CRITICAL/HIGH severity. Two LOW-severity observations worth recording:

**FINDING (LOW, confidence 6)**
- file: `.github/workflows/test.yml`
- line: 3458–3459 (`grep '^CON|' ...`)
- issue: Field separator collision — `pset format unaligned` defaults to `|` and `pg_get_constraintdef` text *can* contain literal `|` characters (e.g. CHECK predicates over `bit`/text expressions, or column names containing it). The awk reconstructs the def with `for (i = 5; i <= NF; i++) def = def "|" $i`, which correctly re-joins extra fields back into a single def string, so the `key = $2"|"$3` pairing remains valid as long as `relname` and `conname` themselves don't contain `|`. They can't here (regular SQL identifiers under `ash.*`), so today this is benign. A future-proofing nit: `\pset fieldsep '\x01'` would remove the entire class of concern. Non-blocking.

**FINDING (LOW, confidence 5)**
- file: `.github/workflows/test.yml`
- line: 3452 (snapshot SQL)
- issue: Filter is `c.relkind in ('r','p')`. This is the *correct* set for table-level constraints on `ash.sample` (parent partitioned `p` + leaf partitions `r`) and intentionally excludes partitioned indexes (`relkind='I'`/`'i'`) — those already get full coverage from the existing `IDX` block via `pg_get_indexdef`. So this is not a miss, just confirming intent matches the issue scope.

---

### test-analyzer

The PR *is* the test, so the question is: does it actually fire on the regression it's meant to catch?

**FINDING (MEDIUM, confidence 8)**
- file: `.github/workflows/test.yml`
- line: 3449–3494 (entire CON block + awk pairing)
- issue: No self-test / negative-control commit demonstrates the new step actually *fails* on a planted divergence. The PR description claims "awk pairing logic verified locally against four scenarios," but the CI side is only proven on the green path (`current install/upgrade chain has no real divergence`).
- untested_scenarios: I re-ran the awk locally against synthetic inputs covering all four cases (def diff, missing-on-upgrade, missing-on-fresh, identical) and it produces correct, paired output with no false positives on identical inputs and no spurious output on unchanged constraints. Counter-factual confidence is therefore high that the step *will* trigger; the missing piece is just a one-shot proof-CI run with `>= 2` re-introduced into one upgrade script, which is an after-merge sanity check rather than a blocker.

**FINDING (LOW, confidence 6)**
- file: `.github/workflows/test.yml`
- line: 3454 (`grep '^CON|' ...`)
- issue: If the snapshot SQL ever fails partially (e.g. one of the four `select`s errors out), the `CON` rows could be empty on one side and identical-empty on the other, masking divergence. Mitigation already exists upstream: the snapshot is run with `-v ON_ERROR_STOP=1`, so a partial snapshot would fail the *previous* step. Acceptable as is.

---

### guidelines-checker

**FINDING (INFO, confidence 7)**
- file: `.github/workflows/test.yml`
- line: 3450
- rule: `development__shell-style-guide.mdc` — quoting / `set -Eeuo pipefail`
- issue: The `run:` block doesn't set `pipefail`. The `if ! diff -q ... > /dev/null; then` is robust because `diff -q` exit code is the gate, but the surrounding script relies on default GitHub Actions shell semantics (`bash -e {0}`). Consistent with the rest of `test.yml` (the existing 3000+ lines also don't set `pipefail`), so non-blocking — but worth noting that a project-wide `defaults.run.shell: bash` with `set -Eeuo pipefail` is a future improvement.

**FINDING (INFO, confidence 8)**
- file: PR title `ci: include pg_constraint in schema-equivalence diff`
- rule: `development__git-commit-standards.mdc` — Conventional Commits
- issue: Conforms. Lowercase, `ci:` scope, present tense, under length limit. No action.

No SQL style violations — the new SQL block follows project conventions (lowercase keywords, lowercase identifiers, `where`/`order by` style consistent with the surrounding three snapshot queries).

---

### Specific points raised in the review request

- **Awk pairing — exists-on-one-side AND def-differs?** Yes for both. The `if (!(key in fresh))` branch handles upgrade-only; the `else if (fresh[key] != $0)` branch handles def diff and `delete fresh[key]` so the END loop only sees fresh-only entries. Verified by running against synthetic inputs covering all four scenarios — output matches expectation.
- **CON enumeration coverage of partitioned objects?** `relkind in ('r','p')` covers parent (`p`) and leaf partitions (`r`) for `ash.sample`. Partitioned indexes (`I`/`i`) intentionally not in scope — already covered by the pre-existing `IDX` block via `pg_get_indexdef`. Foreign tables (`f`), materialized views (`m`), and toast tables are correctly excluded.
- **Canonical sort?** `order by c.relname, con.conname` in SQL and `sort` on the extracted CON-only file before diff. Sort key is stable across PG 14–18 (constraint OIDs and creation order vary across install paths, which is exactly why the explicit ORDER BY matters).
- **`::error::` annotation format?** Valid. GitHub Actions accepts `::error::message` (basic) and `::error file=,line=,col=::message` (annotated). Both used correctly. Confirmed by green CI runs in the PR.
- **Does the new step trigger on planted divergence?** Local re-validation of the awk on synthetic divergent input produces the expected paired `upgrade ash.X.Y / fresh ash.X.Y` block. Real-CI proof on a planted divergence would be ideal but is not strictly required for this merge.

---

### Summary

| Agent              | Severity         | Count |
|--------------------|------------------|-------|
| bug-hunter         | LOW              | 2     |
| test-analyzer      | MEDIUM / LOW     | 1 / 1 |
| guidelines-checker | INFO             | 2     |

No blocking findings. PR is mergeable. Suggested follow-up (non-blocking): one-shot CI proof by temporarily reverting one upgrade-script CHECK to `>= 2` to demonstrate the new step actually fails red.

---

PR #71, opened 2026-04-25
refactor: centralize admin function set in ash._admin_funcs() (#67)

### LINKED ISSUE #67: refactor: centralize admin-function set into ash._admin_funcs()
## Problem
The set of "admin" functions (those that should NOT be granted to monitoring roles) is currently duplicated across three places:

- `ash.grant_reader(name)` — hardcoded array literal
- `ash.revoke_reader(name)` — hardcoded array literal
- STEP 6 hardening block — explicit list

Three sources of truth that can drift. Future admin functions added to one but not all break either grant_reader's correctness or hardening's coverage.

## Proposal
Centralize into `ash._admin_funcs() returns text[]`:

\`\`\`sql
create or replace function ash._admin_funcs() returns text[]
language sql immutable parallel safe
as \$\$
  select array[
    'start', 'stop', 'uninstall', 'rotate', 'take_sample',
    'set_debug_logging', 'rebuild_partitions', 'rollup_minute',
    'rollup_hour', 'rollup_cleanup', '_drop_all_partitions',
    '_rebuild_query_map_view', '_merge_wait_counts',
    '_merge_query_counts', '_truncate_pairs',
    '_int4_array_cat_agg', '_int8_array_cat_agg',
    'grant_reader', 'revoke_reader',
    -- ...
  ]
\$\$;
\`\`\`

Then `grant_reader`, `revoke_reader`, and the STEP 6 hardening loop all read from this single source.

## Detected by
REV review on PR #60 — see https://github.com/NikolayS/pg_ash/pull/60#issuecomment-4316861003

NikolayS:
## Summary
- Add `ash._admin_funcs()` (`language sql immutable parallel safe`) returning the canonical text[] of admin function names — single source of truth for the privilege machinery.
- Refactor `ash.grant_reader()`, `ash.revoke_reader()`, and the STEP 6 REVOKE-from-PUBLIC / GRANT-to-owner hardening block to read from `ash._admin_funcs()`. Adding a new admin entry point now requires updating only one list.
- The hardening block resolves signatures dynamically via `pg_proc` (`prokind in ('f','a')` covers regular functions and the `_int{4,8}_array_cat_agg` aggregates). Entries in `_admin_funcs()` that don't yet exist at install-order (grant_reader/revoke_reader are defined later) are skipped here and locked down by their own existing DO block once created.
- Behavior is byte-identical: same set of functions revoked from PUBLIC, granted to the schema owner; verified locally with a fresh install (`{nik=X/nik}` ACL on all 19 admin entries) and the 1.0 → 1.4 upgrade chain.
- Mirror changelog into `sql/ash-1.3-to-1.4.sql` (no-op since that file `\ir`'s `ash-install.sql`).

## Test plan
- [x] Fresh install (`\i sql/ash-install.sql`) — verified all 19 admin functions revoked from PUBLIC, granted only to schema owner.
- [x] Upgrade chain 1.0 → 1.1 → 1.2 → 1.3 → 1.4 — `ash._admin_funcs()` exists and returns 19 entries post-upgrade.
- [x] `ash.grant_reader('ash_grafana_test')` produces expected privilege set: reader can EXECUTE `top_waits`, cannot EXECUTE `start`/`take_sample`/`grant_reader`/`revoke_reader`.
- [x] `ash.revoke_reader()` symmetrically removes them.
- [x] CI test added: `array_length(ash._admin_funcs(), 1) > 10` plus spot-checks for canonical entries.
- [ ] CI green across PG 14–18 (test.yml).

Closes #67.

Generated with [Claude Code](https://claude.com/claude-code)


NikolayS:
# REV review — PR #71 (`refactor: centralize admin function set in ash._admin_funcs()`)

Closes #67. Net diff: +73 / -54 across `sql/ash-install.sql`, `sql/ash-1.3-to-1.4.sql`, and `.github/workflows/test.yml`.

## Verdict

Approve with minor concerns. The refactor is correct, the admin set is byte-identical to the pre-refactor union of all three sites, no privilege-relevant entries were dropped, and the new STEP 6 hardening loop actually broadens coverage (it now also locks down `grant_reader`/`revoke_reader` once they're defined, on top of the existing terminal DO block). Not a blocker, but a few non-trivial observations below.

---

## Completeness check (the load-bearing question)

Pre-refactor admin set, by site:

| function | STEP 6 (17) | grant_reader local (19) | revoke_reader local (19) |
|---|---|---|---|
| start, stop, uninstall, rotate, take_sample | yes | yes | yes |
| set_debug_logging, rebuild_partitions | yes | yes | yes |
| _drop_all_partitions, _rebuild_query_map_view | yes | yes | yes |
| rollup_minute, rollup_hour, rollup_cleanup | yes | yes | yes |
| _merge_wait_counts, _merge_query_counts, _truncate_pairs | yes | yes | yes |
| _int4_array_cat_agg, _int8_array_cat_agg | yes | yes | yes |
| **grant_reader, revoke_reader** | **no** | yes | yes |

New `ash._admin_funcs()` returns the **union** (19 entries) — nothing dropped, and `grant_reader`/`revoke_reader` are now also explicitly enumerated for STEP 6 (which previously relied on the terminal `do $$ ... revoke all on function ash.grant_reader(name) ... $$` block 4858–4868 to lock them down). That terminal DO block is preserved, so the lockdown is now belt-and-suspenders. Net: strict superset of prior coverage.

Aggregates: `prokind in ('f','a')` correctly covers the two `_intN_array_cat_agg` aggregates. Confirmed via grep: zero `CREATE PROCEDURE` exists anywhere in `sql/ash-install.sql` (only 58 `CREATE OR REPLACE FUNCTION` and 2 `CREATE AGGREGATE`). `prokind = 'p'` is not needed today; if a procedure is ever added to the admin set, this loop will silently miss it — see FINDING-3.

---

## FINDING-1 (security, MEDIUM, confidence 7)

- **file:** `.github/workflows/test.yml` (PR check status), `sql/ash-install.sql` (#67 itself)
- **issue:** The PR's only CI signal is **CodeQL** + `Analyze (actions)`. The `test.yml` workflow that exercises PG 14–18, fresh install, the 1.0 → 1.4 upgrade chain, schema equivalence, degraded-mode, **and the new `array_length(ash._admin_funcs(), 1) > 10` assertion the PR adds** (lines 565–580 of `.github/workflows/test.yml`) **never ran**.
- **evidence:** `gh api repos/NikolayS/pg_ash/commits/02466cc.../check-runs` returns only `CodeQL` and `Analyze (actions)`. PR description self-reports `[ ] CI green across PG 14–18 (test.yml)` unchecked. `mergeStateStatus: DIRTY`, `mergeable: CONFLICTING` — main has moved past the PR's base (1.4 reader/grant_reader machinery landed in #58/#60/#62/#64 after this PR was opened).
- **why this matters for a privilege-boundary refactor:** the assertion the author added (`assert array_length(ash._admin_funcs(), 1) > 10`) is the *only* automated guard against silent shrinkage of the admin set, and it has not been executed against any PG version. A merge today would land an unverified privilege-boundary change.
- **fix:** rebase on current `main` to resolve the conflict, push, and require a green `test.yml` run before merge. The CLAUDE.md "Deployment" section (`tests across PostgreSQL 14, 15, 16, 17, 18`) is explicit that this matrix is the gate.

## FINDING-2 (bug-hunter, LOW, confidence 6)

- **file:** `sql/ash-install.sql` lines 4577–4629 (new STEP 6 hardening loop)
- **issue:** "entries silently skipped if not yet created at install order" works **today** because of the terminal DO block at 4858–4868 (`revoke all on function ash.grant_reader(name) from public; ... grant execute ... to %I, v_owner`). But there is no static link between `_admin_funcs()` and that terminal block: **add a new admin function later in the file (after STEP 6) and forget to extend the terminal DO block, and `_admin_funcs()` will silently skip it in STEP 6 without any compensating REVOKE**. The reviewer's question — "does adding a new admin function to `_admin_funcs()` ALONE actually cascade?" — answer: **only if the function is created BEFORE STEP 6 runs**. For functions defined after STEP 6, you still need a paired explicit DO block (or the STEP 6 loop has to be moved/duplicated below all CREATEs).
- **fix (non-blocking, future-proofing):** either (a) re-run the STEP 6 admin loop a second time at end-of-file (idempotent — `revoke all` + `grant execute to %I, owner` is safe to repeat), or (b) add a sentinel CI test asserting that every entry in `ash._admin_funcs()` resolves to an actual `pg_proc` row at install completion AND that PUBLIC has zero EXECUTE on each. Today's `array_length > 10` only catches deletion, not "name in list but function not yet created at lockdown time."
- **evidence:** `grant_reader`/`revoke_reader` are defined at lines 4711 and 4789 — *after* STEP 6's loop at 4577. The new loop's `where ... and p.proname::text = any(v_admin_funcs)` returns zero rows for those two names, so neither REVOKE nor GRANT-to-owner runs for them in STEP 6. The trailing DO block 4858–4868 is what saves correctness today; it is not enforced by the design.

## FINDING-3 (security, LOW, confidence 5)

- **file:** `sql/ash-install.sql` lines 4585, 4744, 4818
- **issue:** All three `pg_proc` filter sites use `p.prokind in ('f','a')` (STEP 6 hardening) or `p.prokind = 'f'` (`grant_reader`/`revoke_reader`). If pg_ash ever adds a `CREATE PROCEDURE ash.foo()` and lists `'foo'` in `_admin_funcs()`, the STEP 6 loop will skip it (no REVOKE-from-PUBLIC, no GRANT-to-owner) and `grant_reader` will skip it (a procedure listed as admin would be invisible to the reader-grant loop too — that side-effect is fine). The PR comment explicitly anticipates this (`prokind in ('f','a') covers regular functions and aggregates`) but does not assert "no procedures exist in ash."
- **fix (non-blocking):** either widen to `prokind in ('f','a','p')` proactively, or add a CI assertion: `select count(*) from pg_proc p join pg_namespace n on p.pronamespace = n.oid where n.nspname = 'ash' and p.prokind = 'p'` must equal 0. Today the assertion holds (zero procedures exist), so this is a latent footgun rather than a bug.

## FINDING-4 (security, INFO, confidence 7)

- **file:** `sql/ash-install.sql` line 4583 (new `ash._admin_funcs()` definition)
- **issue:** `_admin_funcs()` itself is **not** in the admin set. STEP 6's reader-loop at 4634–4648 (`where ... p.prokind = 'f'`) does revoke EXECUTE on `_admin_funcs` from PUBLIC, and `grant_reader` will subsequently grant EXECUTE on it to monitoring roles (since it's not `<> all(v_admin_funcs)`). That's intentional and harmless — `_admin_funcs()` is `immutable, parallel safe` and returns a constant text[] disclosing the *names* of admin functions. Knowing the names is not privilege-escalating (those names are in `comment on function` text and in source anyway), and the function takes no arguments, so there's no injection surface.
- **action:** none. Documenting the deliberate non-membership in the function comment is fine; no code change needed.

## FINDING-5 (docs/changelog, LOW, confidence 7)

- **file:** `sql/ash-1.3-to-1.4.sql` lines 51–53
- **issue:** The changelog entry says `Refactor; no behavior change (#67)` — accurate for the function set, but the new `ash._admin_funcs()` IS a new public-ish function (revoked from PUBLIC by STEP 6's reader-loop, then re-granted by `grant_reader`). Operators auditing post-upgrade `\df ash.*` will see a new entry. Worth one line.
- **fix:** append `(adds ash._admin_funcs() — sql/immutable, returns canonical admin name list)` to the existing bullet so the upgrade footprint is self-evident.

## FINDING-6 (style, INFO, confidence 8)

- **file:** `sql/ash-install.sql` line 4596 (`_admin_funcs()` body)
- **issue:** The comment block above the function says "Adding a new admin entry point requires updating only this list" but the list is hardcoded inside `select array[ ... ]::text[]`. If a future contributor follows that instruction in a feature PR (per CLAUDE.md: "Upgrade scripts ... are generated at release time, not in feature PRs"), they edit one place — good. But the `language sql immutable` choice prevents adding any conditional logic later (e.g. "include `'_thing'` only on PG 16+"). Not a problem today; flagging as informational only.
- **action:** none.

---

## Other watch-list items (per review prompt)

- **Idempotency** (re-running install.sql produces identical ACLs): yes. STEP 6 hardening loop's `revoke all + grant execute to %I, v_owner` is idempotent; reader-loop is idempotent; terminal DO block is idempotent.
- **`_admin_funcs()` itself in the admin set?** No, and that's correct. See FINDING-4.
- **"Silently skipped" entries safety:** Safe TODAY because the terminal DO block at 4858–4868 hand-holds `grant_reader`/`revoke_reader`. Not statically guaranteed for future additions — see FINDING-2.

## Summary

The diff does what it claims, the privilege boundary is preserved, the new CI assertion is a real (if minimal) regression guard, and converting three duplicated lists into one is a clear maintainability win. The blocking concern is **operational, not code**: this PR has not been validated against the test.yml matrix and is currently in a conflicting state. Resolve, push, get green PG 14–18, then merge.

PR #65, opened 2026-04-24
fix(readers): clamp ts_from_timestamptz to int4 range; _at variants no longer overflow (#63)

NikolayS:
## Summary

- Fixes #63: every `_at` reader (`top_waits_at`, `samples_at`, `query_waits_at`, `event_queries_at`, `timeline_chart_at`, `top_queries_at`, `wait_timeline_at`, `top_by_type_at`, `minute_waits_at`, `hourly_queries_at`, `daily_peak_backends_at`) raised `integer out of range` on absurd timestamps because `ash.ts_from_timestamptz(p_ts)` did `extract(epoch from p_ts - ash.epoch())::int4` with no clamp — pre-epoch inputs underflowed, post-2094-horizon inputs overflowed.
- Centralize the clamp at the helper to `[0, INT4_MAX]`. Pre-epoch -> `0`, post-horizon -> `2147483647`. `sample_ts` is a non-negative `int4` by construction, so both clamped endpoints fall outside the actual sample window and readers naturally yield empty results — no hard error.
- Sibling of merged #51 (PR #57), which clamped the interval-based readers; the `_at` variants were explicitly deferred during that PR's review. Same clamp pattern, just centralized one level deeper at the shared helper so every current and future `_at` caller inherits it.
- Mirror the changelog entry into `sql/ash-1.3-to-1.4.sql` per the install.sql lockstep policy. CI regression assertions added in `test.yml`: helper-level (`1000-01-01 -> 0`, `3000-01-01 -> INT4_MAX`), pre-epoch and post-horizon ranges yield empty `top_waits_at` / `samples_at`, plus a normal range still works.

## Test plan

- [x] Local: `psql -f sql/ash-install.sql` clean install on PG 18
- [x] Local: confirmed pre-fix body raises `integer out of range` for `'1000-01-01'::timestamptz`
- [x] Local: post-fix helper returns `0` for pre-epoch, `2147483647` for post-horizon
- [x] Local: all 11 `_at` readers return 0 rows (or non-error placeholder) for both pre-epoch and post-horizon ranges
- [x] Local: normal range (`now() - 10 min` .. `now()`) still works
- [ ] CI: fresh install, upgrade chain, schema equivalence, degraded mode (PG 14-18)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Issue #63, opened 2026-04-24
bug: _at readers raise 'integer out of range' on absurd timestamps — same class as #51


NikolayS:
## Summary
Sibling of #51. The `_at` variants (`top_waits_at`, `samples_at`, `query_waits_at`, `event_queries_at`, `timeline_chart_at`, `top_queries_at`, `wait_timeline_at`, `top_by_type_at`, `minute_waits_at`, `hourly_queries_at`, `daily_peak_backends_at`) all call `ash.ts_from_timestamptz(p_ts)` which has an unguarded `extract(epoch from p_ts - ash.epoch())::int4` cast.

Reproducible on PG 18.3:
\`\`\`sql
SELECT ash.top_waits_at('1000-01-01'::timestamptz, '1000-01-02'::timestamptz);
-- ERROR: integer out of range

SELECT ash.top_waits_at('3000-01-01'::timestamptz, '3000-01-02'::timestamptz);
-- ERROR: integer out of range
\`\`\`

## Fix
Apply the same `greatest(..., 0)::int4` clamp pattern in `ash.ts_from_timestamptz`, plus an upper-bound clamp (either to int4 max, with a NOTICE that values beyond the horizon are treated as "future, no data yet"). Could also surface via `_active_slots_for_at(p_start, p_end)` to emit the oversized-range NOTICE symmetrically with #51's approach.

## Detected by
REV review on #57 — see https://github.com/NikolayS/pg_ash/pull/57#issuecomment-4316890382

PR #72, opened 2026-04-25
feat(readers): _active_slots_for_at for NOTICE symmetry on _at variants (#69)

### LINKED ISSUE #69: feature: add ash._active_slots_for_at(p_start, p_end) for NOTICE symmetry
## Problem
PR #57 (issue #51) introduced a `NOTICE` from `ash._active_slots_for(p_interval)` when callers pass an oversized interval. PR #65 (issue #63) clamped `ash.ts_from_timestamptz()` so `_at` readers no longer error on absurd timestamps — but they do so SILENTLY.

Asymmetry: relative readers loud-warn the caller, absolute readers don't.

## Proposal
Add `ash._active_slots_for_at(p_start timestamptz, p_end timestamptz) returns smallint[]` that:
- Returns the active slots covering the time range, or `{}` if the range is entirely pre-epoch / post-horizon.
- Emits a `NOTICE` when the requested range is outside what ash retains (e.g. older than `2 * rotation_period` ago, or in the future).

Then update each `_at` reader to call this helper for filtering, mirroring how the relative readers call `_active_slots_for(p_interval)`.

## Detected by
REV review on PR #65 — see https://github.com/NikolayS/pg_ash/pull/65#issuecomment-4317126482

NikolayS:
## Summary

Closes #69. Restores loud-warn symmetry between relative readers (which call `ash._active_slots_for(p_interval)` and emit a NOTICE on oversized intervals — #51 / PR #57) and absolute readers (which silently returned 0 rows on absurd timestamps after #63 / PR #65 clamped `ts_from_timestamptz()`).

- New `ash._active_slots_for_at(p_start timestamptz, p_end timestamptz) returns smallint[]`. Returns the active slots when the requested range overlaps retained raw samples (`now() - 2 * rotation_period .. now()`); returns `{}` and emits a NOTICE otherwise. Shares the transaction-scoped `ash.notice_oversized` GUC with `_active_slots_for(interval)` for cross-helper dedup.
- Every raw-sample `_at` reader (`top_waits_at`, `top_queries_at`, `wait_timeline_at`, `top_by_type_at`, `query_waits_at`, `event_queries_at`, `timeline_chart_at`, `samples_at`, `decode_sample_at`) routes its slot filter through the new helper.
- Four SQL-language readers (`top_waits_at`, `wait_timeline_at`, `top_by_type_at`, `decode_sample_at`) flip to plpgsql so the helper can be invoked into a local — otherwise the planner constant-folds the time predicate to false on absurd ranges and skips the call (and its NOTICE side effect). `#variable_conflict use_column` disambiguates OUT-param names from CTE columns.
- The silent-clamp UX from #63 is preserved: zero rows on out-of-range, NOTICE added on top, no errors.

## Subtleties

- **Empty-range NOTICE filtering:** `p_start >= p_end` inside the retained window does NOT trigger the NOTICE — callers may legitimately ask for a zero-length window. Only ranges fully outside retention (`p_end <= now() - 2 * rotation_period` OR `p_start > now()`) flag.
- **Rollup readers untouched:** `minute_waits_at`, `hourly_queries_at`, `daily_peak_backends_at` query rollup tables (separate, much longer `rollup_*_retention_days`). Their relative wrappers don't call `_active_slots_for(interval)` either, so adding the new helper to them would emit a spurious "outside retention" NOTICE for ranges that are perfectly fine for rollups.
- **`query_waits_at`** keeps its short-circuit on missing `query_id` BEFORE calling the helper, matching the relative `query_waits` pattern (a non-existent query returns silently with no retention NOTICE).
- **NULL bounds** are passed through silently; the reader's own WHERE clause yields empty.

## Test plan

- [x] Fresh install (`psql -f sql/ash-install.sql`) succeeds on PG 18.
- [x] Idempotent re-apply succeeds.
- [x] Upgrade chain 1.0 → 1.1 → 1.2 → 1.3 → 1.4 succeeds.
- [x] `ash.top_waits_at('1000-01-01', '1000-01-02')` emits NOTICE, returns 0 rows.
- [x] `ash.top_waits_at('3000-01-01', '3000-01-02')` emits NOTICE, returns 0 rows.
- [x] `ash.top_waits_at(now() - '10 minutes', now())` does NOT emit NOTICE.
- [x] `ash.query_waits_at(<non-existent qid>, …)` does NOT emit NOTICE (matches `query_waits` short-circuit).
- [x] All other patched `_at` readers fire the NOTICE on out-of-retention ranges (asserted via the `ash.notice_oversized` GUC in CI).
- [x] Existing future-range tests still pass (count = 0; NOTICE is additive).

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -1325,6 +1325,144 @@ jobs:
           $$;
           EOF
 
+      - name: Test _at NOTICE symmetry (#69)
+        env:
+          PGPASSWORD: postgres
+        run: |
+          # Each scenario runs in its own transaction so the
+          # transaction-scoped ash.notice_oversized GUC starts unset.
+          # We assert it becomes '1' (== NOTICE fired) for out-of-retention
+          # ranges and stays unset for normal ranges, mirroring how the
+          # interval-based readers behave for oversized intervals.
+          psql -h localhost -U postgres -d postgres -v ON_ERROR_STOP=1 << 'EOF'
+          -- Out-of-retention past
+          begin;
+            do $$
+            declare
+              v_count int;
+              v_flag text;
+            begin
+              select count(*) into v_count
+              from ash.top_waits_at('1000-01-01'::timestamptz, '1000-01-02'::timestamptz);
+              assert v_count = 0,
+                'top_waits_at(1000-01-01..) should return 0 rows, got: ' || v_count;
+              v_flag := current_setting('ash.notice_oversized', true);
+              assert v_flag = '1',
+                'top_waits_at(1000-01-01..) should emit NOTICE (ash.notice_oversized=' ||
+                coalesce(v_flag, '<unset>') || ')';
+            end;
+            $$;
+          rollback;
+
+          -- Out-of-retention future
+          begin;
+            do $$
+            declare
+              v_count int;
+              v_flag text;
+            begin
+              select count(*) into v_count
+              from ash.top_waits_at('3000-01-01'::timestamptz, '3000-01-02'::timestamptz);
+              assert v_count = 0,
+                'top_waits_at(3000-01-01..) should return 0 rows, got: ' || v_count;
+              v_flag := current_setting('ash.notice_oversized', true);
+              assert v_flag = '1',
+                'top_waits_at(3000-01-01..) should emit NOTICE (ash.notice_oversized=' ||
+                coalesce(v_flag, '<unset>') || ')';
+            end;
+            $$;
+          rollback;
+
+          -- Normal range — must NOT emit NOTICE
+          begin;
+            do $$
+            declare
+              v_count int;
+              v_flag text;
+            begin
+              select count(*) into v_count
+              from ash.top_waits_at(now() - interval '10 minutes', now());
+              v_flag := current_setting('ash.notice_oversized', true);
+              assert v_flag is null or v_flag = '',
+                'top_waits_at(now()-10m, now()) should NOT emit NOTICE (ash.notice_oversized=' ||
+                coalesce(v_flag, '<null>') || ')';
+            end;
+            $$;
+          rollback;
+
+          -- Cross-reader sanity: each _at reader still routes through
+          -- _active_slots_for_at on out-of-retention, asserted via the GUC.
+          begin;
+            do $$
+            declare
+              v_flag text;
+            begin
+              perform * from ash.samples_at('1000-01-01'::timestamptz, '1000-01-02'::timestamptz, 1);
+              v_flag := current_setting('ash.notice_oversized', true);
+              assert v_flag = '1', 'samples_at: NOTICE not emitted';
+              perform set_config('ash.notice_oversized', '', true);
+
+              perform * from ash.top_queries_at('1000-01-01'::timestamptz, '1000-01-02'::timestamptz);
+              v_flag := current_setting('ash.notice_oversized', true);
+              assert v_flag = '1', 'top_queries_at: NOTICE not emitted';
+              perform set_config('ash.notice_oversized', '', true);
+
+              perform * from ash.top_by_type_at('1000-01-01'::timestamptz, '1000-01-02'::timestamptz);
+              v_flag := current_setting('ash.notice_oversized', true);
+              assert v_flag = '1', 'top_by_type_at: NOTICE not emitted';
+              perform set_config('ash.notice_oversized', '', true);
+
+              perform * from ash.wait_timeline_at('1000-01-01'::timestamptz, '1000-01-02'::timestamptz);
+              v_flag := current_setting('ash.notice_oversized', true);
+              assert v_flag = '1', 'wait_timeline_at: NOTICE not emitted';
+              perform set_config('ash.notice_oversized', '', true);
+
+              perform * from ash.timeline_chart_at('1000-01-01'::timestamptz, '1000-01-02'::timestamptz);
+              v_flag := current_setting('ash.notice_oversized', true);
+              assert v_flag = '1', 'timeline_chart_at: NOTICE not emitted';
+              perform set_config('ash.notice_oversized', '', true);
+
+              perform * from ash.decode_sample_at('1000-01-01'::timestamptz);
+              v_flag := current_setting('ash.notice_oversized', true);
+              assert v_flag = '1', 'decode_sample_at: NOTICE not emitted';
+              perform set_config('ash.notice_oversized', '', true);
+
+              -- query_waits_at: needs an existing query_id to avoid the
+              -- short-circuit; the seeded fixture uses 111111 (see earlier
+              -- "Seed test data" step in this job).
+              perform * from ash.query_waits_at(
+                111111,
+                '1000-01-01'::timestamptz, '1000-01-02'::timestamptz);
+              v_flag := current_setting('ash.notice_oversized', true);
+              assert v_flag = '1', 'query_waits_at: NOTICE not emitted';
+
+              raise notice 'All _at readers fire NOTICE on out-of-retention range';
+            end;
+            $$;
+          rollback;
+
+          -- query_waits_at on a non-existent query must NOT emit NOTICE
+          -- (matches the relative ash.query_waits short-circuit pattern: the
+          -- function returns silently before reaching the helper).
+          begin;
+            do $$
+            declare
+              v_flag text;
+            begin
+              perform * from ash.query_waits_at(
+                999999999,
+                '1000-01-01'::timestamptz, '1000-01-02'::timestamptz);
+              v_flag := current_setting('ash.notice_oversized', true);
+              assert v_flag is null or v_flag = '',
+                'query_waits_at(non-existent qid) should not NOTICE (ash.notice_oversized=' ||
+                coalesce(v_flag, '<null>') || ')';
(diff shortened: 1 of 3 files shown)


NikolayS:
## REV review — PR #72 (`_active_slots_for_at` for NOTICE symmetry, #69)

**Verdict: APPROVE with two minor follow-ups.** The PR cleanly closes the silent-empty UX gap from #63 by introducing an absolute-range counterpart to `_active_slots_for(interval)`. Helper logic is symmetric with the existing relative variant, NOTICE wording matches the relative helper's style (range/window phrasing adapted appropriately), GUC dedup is reused (`ash.notice_oversized`), and the rationale for the four `language=sql` -> `language=plpgsql` flips is documented in-source and well-founded (planner constant-folding would otherwise drop the helper call along with its NOTICE side effect).

CI: GitHub-side checks (CodeQL, Analyze) PASS. The substantive cross-version Postgres test matrix (`test.yml`) doesn't appear in `statusCheckRollup` for this run — worth a manual re-trigger after the rebase below. The merge state is `DIRTY/CONFLICTING` (head is one commit behind `main` after `df99e1c` "centralize admin function set in `ash._admin_funcs()`" landed, which also edited `ash-1.3-to-1.4.sql`). The conflict is just the upgrade-script changelog hunk — trivial textual rebase, no semantic clash.

---

### Bug hunter

**NO_FINDINGS (high confidence).** Spot-checks on the seven things most likely to bite:

- **`#variable_conflict use_column`** — applied to the three plpgsql flips that reference OUT-param-shadowed names inside CTEs (`top_waits_at`, `wait_timeline_at`, `top_by_type_at`). `decode_sample_at` doesn't need it (the `RETURN QUERY` aliases via `s.datid, d.wait_event, ...` and bare `wait_event/query_id/count` never appear unaliased inside the body). No false positive renames; no shadowed param ambiguity.
- **`top_waits_at` cast change** — `r.cnt::bigint as samples` was added. Previously the SQL body returned `samples bigint` from `count(*)` which is already `bigint`; the explicit `::bigint` is redundant but not wrong. (`r.cnt` comes from `count(*) as cnt` in `waits`, which is `bigint`. Cast is a no-op — fine.)
- **Edge case `p_start = p_end = now()`** (empty range fully in retention) — helper takes the else branch (`p_end <= retention_start` is false because `p_end = now() > now() - 2*rotation_period`; `p_start > now()` is false because `p_start = now()`), returns active slots, no NOTICE. Correct.
- **Edge case `p_start < retention_start < p_end`** (range straddles the boundary) — neither outside-condition fires; helper returns active slots, no NOTICE; reader's own `sample_ts >= v_start` predicate filters out the un-retained portion. Partial results returned silently. This matches the relative helper's behavior (an interval slightly larger than `rotation_period` but smaller than `2*rotation_period` returns retained data with no NOTICE) — symmetric and correct. Documenting "we only NOTICE on fully-out-of-window" in the helper header would help future readers.
- **Edge case `p_start > p_end` inside retention** (zero-or-negative-width range) — helper returns active slots, no NOTICE; reader returns 0 rows. PR body and helper comment explicitly call this out as intentional.
- **Edge case NULL bounds** — both null-checks before the outside-window comparison; the helper returns active slots when either bound is null, the reader's `>= NULL` / `< NULL` predicates yield empty. Clean.
- **`query_waits_at` ordering** — slot computation deliberately follows the `query_id` existence short-circuit so a non-existent qid returns silently (no retention NOTICE). Test asserts this. Matches the relative `query_waits` shape.

### Test analyzer

**MEDIUM, confidence 7.** The 138 lines of new `test.yml` coverage are structurally good — runs each scenario in its own transaction so the local GUC starts clean, asserts the GUC is `'1'` post-call (hard signal that NOTICE fired), and covers six readers plus the `query_waits_at` happy/short-circuit pair. But several gaps:

1. **No straddle test.** `p_start < retention_start < p_end` is the most realistic "ambiguous" case in production (caller asks for "the last 4 hours" with a 1h `rotation_period`). PR body promises silent partial return; should be asserted.
2. **No GUC-dedup test.** A two-call sequence in the same transaction should fire NOTICE once, not twice — that's the whole point of the shared `ash.notice_oversized` GUC. The current test resets the GUC between calls (`set_config('ash.notice_oversized', '', true)`), masking whether dedup actually works. One unreset multi-call subcase would close the gap.
3. **No NULL-bound test.** Helper has explicit `p_start is not null and p_end is not null` guard; needs an assertion that `top_waits_at(null, now())` returns empty without firing the NOTICE.
4. **No `language=sql` regression check.** The whole rationale for the four flips is "planner would constant-fold and drop the NOTICE." Worth one test that proves the NOTICE fires on `top_waits_at('1000-01-01', '1000-01-02')` with `set jit = off` AND with `set jit = on` (the function's own `set jit = off` should pin behavior, but it's an explicit assert of the plpgsql-flip rationale).
5. **`event_queries_at` not in cross-reader sanity loop.** The PR diff adds the helper call to `event_queries_at` (line ~4549 in the diff), but the test loop covers `samples_at, top_queries_at, top_by_type_at, wait_timeline_at, timeline_chart_at, decode_sample_at, query_waits_at`. Missing one of the eight readers the PR claims to update.

### Guidelines checker

**INFO, confidence 6.**

- **`ash-install.sql` as source of truth** (per CLAUDE.md): respected. Upgrade script `ash-1.3-to-1.4.sql` only updates the changelog comment + `\ir ash-install.sql`. Correct pattern.
- **SQL style guide compliance:** lower-case keywords, snake_case, two-space indent in plpgsql declares — matches the rest of `ash-install.sql`.
- **Commit message:** `feat(readers): add _active_slots_for_at for NOTICE symmetry on _at variants` — Conventional Commits compliant, present tense, scope sensibly chosen. PR title differs slightly (`feat(readers): _active_slots_for_at for NOTICE symmetry on _at variants (#69)` — drops the verb). Either form is fine.
- **NOTICE wording:** the new helper says "requested range [%, %) lies outside the retained window (now - 2 * rotation_period .. now, i.e. [%, %)); only the last two partitions are retained, so older or future samples are not available. Adjust the range or increase rotation_period via ash.config." vs the relative helper's "requested interval % exceeds 2 * rotation_period (%); only the last two partitions are retained, so older samples are not available. Shorten the interval or increase rotation_period via ash.config." Phrasing is parallel and informative. Minor: the new message says "older or future samples" — but if `p_start > now()`, there are no "samples" in the future (sampling hasn't happened yet); a tighter phrasing might be "older samples or future timestamps that haven't been sampled yet." Not blocking.

### Docs reviewer

**INFO, confidence 5.**

- The new helper has a thorough, in-source header (out-of-retention conditions, why empty-range is exempt, the GUC-dedup rationale, and the "must use a local in plpgsql" pattern). Good.
- The reasoning for the four `sql` -> `plpgsql` flips is documented at each call site, not just once in the helper header. Slightly redundant but defensible (the comments are local to each function and survive future reorganization).
- **Rollup readers ("intentionally skipped"):** the PR body justifies this — `minute_waits_at`, `hourly_queries_at`, `daily_peak_backends_at` query rollup tables with separate longer retention (`rollup_*_retention_days`), so the helper's `now() - 2 * rotation_period` would emit a spurious NOTICE for ranges that are fine for rollups. **Argument accepted.** The rollup readers do need their own retention check eventually (there are valid out-of-rollup-retention ranges that should NOTICE) — that's a follow-up issue, not a defect of this PR. Filing it would be appropriate (a `_active_slots_for_rollup_at` or similar helper, parameterized by the rollup retention setting). **Recommend:** open a tracking issue rather than expanding scope here.

### Plpgsql-flip performance impact

For `top_waits_at` / `wait_timeline_at` / `top_by_type_at` / `decode_sample_at`: SQL functions in `language sql` can be inlined by the planner, plpgsql cannot. The flip means each call now incurs an extra plan step (function call boundary, no inlining). However:

- All four bodies are large CTE pipelines that already dominate cost — the extra dispatch is sub-millisecond noise.
- All four have `set jit = off` (preserved across the flip), so JIT compile cost isn't a regression vector.
- The `search_path = pg_catalog, ash` security attribute is preserved on all four. No new security exposure.
- `decode_sample_at` has a minor **behavioral** change worth flagging: the old `language=sql` body delegated to `ash.decode_sample(int4)` which scans `ash.sample` with no slot filter (returning rows from any of 3 slots that happen to have a matching `sample_ts`). The new body restricts to `s.slot = any(v_slots)` (the 2 active slots). The PR comment correctly notes this is a no-op in steady state (only active slots hold data after rotation), but during the brief window between `current_slot` advance and the corresponding `TRUNCATE` of the now-stale slot, there could in principle be stale data. Looking at `ash.rotate()` on `main`, the truncate happens in the same transaction as the slot advance, so this window doesn't really exist for committed reads. **Minor / no action.**

---

### Recommendations (non-blocking)

1. **Rebase onto `main`** (mergeStateStatus is DIRTY because of the `ash-1.3-to-1.4.sql` changelog hunk in `df99e1c`). Trivial textual conflict.
2. **Add the four missing test cases** above (straddle, NULL, GUC dedup, `event_queries_at`).
3. **File a follow-up issue** for rollup-reader NOTICE symmetry (`_active_slots_for_rollup_at` or equivalent, parameterized by rollup retention). Out of scope here.
4. **Re-run the full test matrix** after rebase (cross-version PG14-18 job isn't visible in the current `statusCheckRollup`).

---

PR #82, opened 2026-05-28
Critical bug: prevent rollup/rotation raw sample loss

### LINKED ISSUE #81: Critical bug: rollup/rotation can silently lose raw samples
## Summary

`ash.rotate()` and `ash.rollup_minute()` can silently lose long-term ASH data under concurrency. I reproduced two independent paths:

1. `rotate()` truncates an endangered raw slot even when the pre-truncation rollup was skipped or failed.
2. `rollup_minute()` can advance `last_rollup_1m_ts` past a minute while an in-flight sampler later commits a row into that already-rolled-up minute.

Tested on current `main` at `ec43487` (`docs: fix v1.4 release gate findings (#80)`) against PostgreSQL 16.14 in Docker.

## Impact

This is silent data loss in the durability path from raw samples to rollups:

- raw data can be truncated by rotation before it is represented in `ash.rollup_1m`;
- late-committed raw samples can remain in raw partitions but never enter `ash.rollup_1m` because the watermark has advanced;
- after raw partition rotation, those missed samples are gone from both raw and rollup history.

This breaks the main v1.4 durability claim that rollups survive raw partition rotation.

## Expected behavior

- `rotate()` should not truncate the endangered raw slot unless pre-truncation rollup completed successfully for that slot/window, or can prove there is nothing to roll up.
- `rollup_minute()` should not advance `last_rollup_1m_ts` past a minute while any in-flight sampler can still commit a row with a `sample_ts` inside that minute.

### Reproduction A: `rotate()` truncates when rollup advisory lock is busy

Setup: current slot is `0`; the endangered slot for the next rotation is `2`; slot `2` contains one raw sample.

Hold the rollup advisory lock in one session:

```sql
begin;
select pg_advisory_xact_lock(
  hashtext('pg_ash')::int4,
  hashtext('pg_ash_rollup')::int4
);
select pg_sleep(6);
commit;
```

In another session, run:

```sql
select ash.rotate() as rotate_result;
```

Observed:

```text
rotate_result = rotated: slot 0 -> 1, truncated slot 2 (sample + query_map)
raw_rows_after = 0
rollup_rows_after = 0
current_slot = 1
```

No warning is emitted in this path. `ash.rollup_minute()` returns `0` because the rollup advisory lock is busy, and `rotate()` treats that as success.

### Reproduction B: `rotate()` truncates after pre-rollup lock timeout

Hold the rollup table lock in one session:

```sql
begin;
lock table ash.rollup_1m in access exclusive mode;
select pg_sleep(6);
commit;
```

In another session, run:

```sql
select ash.rotate() as rotate_result;
```

Observed:

```text
WARNING: ash.rotate: rollup_minute failed [55P03]: canceling statement due to lock timeout
rotate_result = rotated: slot 0 -> 1, truncated slot 2 (sample + query_map)
raw_rows_after = 0
rollup_rows_after = 0
```

The warning is visible, but the data is still deleted.

### Reproduction C: `rollup_minute()` watermark race with an in-flight sampler

Shape of the repro:

1. Insert one visible seed sample at second `55` of a minute.
2. Hold `lock table ash.sample_0 in share mode` so an incoming sampler blocks before inserting.
3. Start `ash.take_sample()` at second `56`; its `sample_ts` belongs to the old minute, but the insert is blocked.
4. At second `01` of the next minute, call `ash.rollup_minute(5)`.
5. Release the table lock; the sampler commits the raw row with the old minute's `sample_ts`.
6. Call `ash.rollup_minute(5)` again.

Observed from one run:

```text
rollup_before_release = 1
rollup_1m before release:
  ts=12775260 datid=0 samples=1 peak_backends=1 wait_counts={1,1}
  watermark=2026-05-28 20:42:00+00

take_sample started 2026-05-28 20:41:56+00
take_sample ended   2026-05-28 20:42:15+00

raw samples after release:
  2026-05-28 20:41:55+00 datid=0     active_count=1
  2026-05-28 20:41:56+00 datid=16384 active_count=6

rollup_after_release = 0
rollup_1m still has only the seed datid=0 row
```

The late-committed `datid=16384` sample remains in `ash.sample`, but it is permanently skipped by `rollup_minute()` because `last_rollup_1m_ts` already moved to `20:42:00`.

## Root cause hypothesis

There are two related coordination gaps:

- `take_sample()` and `rollup_minute()` use different advisory locks, so rollup does not wait for in-flight samplers that can still commit rows into the minute being rolled up.
- `rotate()` treats pre-truncation rollup as best-effort: it catches rollup exceptions and ignores a `0` return caused by a busy rollup lock, then proceeds to update `current_slot` and `TRUNCATE` the raw partition.

## Suggested fix direction

One possible invariant: before `rotate()` truncates a raw slot, it must either complete the required rollup window successfully or abort rotation without changing slots/truncating. Also, minute rollup should avoid processing a minute that can still receive commits from an in-flight sampler, for example by coordinating with the sampler lock or by leaving a safety delay / watermark rule that accounts for sampler statement timeout and blocked inserts.

## Tests needed

Regression tests should prove at least:

- if the rollup advisory lock is held, `ash.rotate()` does not truncate endangered raw samples;
- if `ash.rollup_1m` is locked and pre-rollup times out, `ash.rotate()` does not truncate endangered raw samples;
- an in-flight sampler with an old-minute `sample_ts` cannot be permanently missed by `rollup_minute()`.

r314tive:
Fixes #81.

## Summary

This fixes two data-loss paths around raw sample retention:

- `ash.rollup_minute()` no longer advances the minute watermark while a sampler is in flight. It briefly takes the sampler advisory lock before computing the complete-minute boundary, and returns `0` without changing the watermark if the sampler lock is busy.
- `ash.rotate()` no longer treats pre-truncation rollup as best-effort when the endangered raw slot contains rows. It aborts rotation if rollup is busy, fails, or cannot prove the endangered slot is represented in `ash.rollup_1m`.
- Adds CI regression coverage for issue #81: sampler lock held, rollup advisory lock busy, and `rollup_1m` lock timeout.

## Proof

RED on current behavior before the fix:

- With the sampler advisory lock held, `ash.rollup_minute(5)` advanced the watermark and wrote rollup rows: `result=1`, `rollup_1m rows=1`.
- With the rollup advisory lock held, `ash.rotate()` still truncated the endangered raw slot.
- With `ash.rollup_1m` locked until `lock_timeout`, `ash.rotate()` warned and still truncated the endangered raw slot.

GREEN after the fix, run locally against PostgreSQL 16 in the lab container:

- `#81A`: while sampler lock was held, `ash.rollup_minute(5)` returned `0`, `last_rollup_1m_ts` stayed NULL, and `ash.rollup_1m` stayed empty. After the lock was released, the next rollup processed the row.
- `#81B`: with rollup advisory lock busy, `ash.rotate()` returned `failed: pre-truncation rollup busy; slot 2 not truncated`; `ash.sample_2` still had the raw row and `current_slot` stayed `0`.
- `#81C`: with `ash.rollup_1m` locked, `ash.rotate()` returned `failed: pre-truncation rollup failed [55P03]; slot 2 not truncated`; `ash.sample_2` still had the raw row and `current_slot` stayed `0`.

## Local checks

- `PGPASSWORD=postgres psql -h 127.0.0.1 -p 55432 -U postgres -d ash_fix_test -v ON_ERROR_STOP=1 -f sql/ash-install.sql`
- Local issue #81 regression script matching the new CI step: passed (`GREEN`).
- `git diff --check -- .github/workflows/test.yml sql/ash-install.sql`: passed.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -2962,6 +2962,198 @@ jobs:
           $$;
           EOF
 
+      - name: "Critical bug #81: rollup and rotation must not lose raw samples"
+        env:
+          PGPASSWORD: postgres
+        run: |
+          set -euo pipefail
+
+          psql_base=(psql -h localhost -U postgres -d postgres -v ON_ERROR_STOP=1)
+
+          reset_issue_81_state() {
+            "${psql_base[@]}" << 'SQL'
+            select ash.stop();
+            truncate ash.sample;
+            truncate ash.rollup_1m;
+            truncate ash.rollup_1h;
+            truncate ash.wait_event_map restart identity cascade;
+            truncate ash.query_map_0 restart identity;
+            truncate ash.query_map_1 restart identity;
+            truncate ash.query_map_2 restart identity;
+            update ash.config
+            set current_slot = 0,
+              rotated_at = now() - interval '2 days',
+              rotation_period = interval '1 day',
+              last_rollup_1m_ts = null,
+              last_rollup_1h_ts = null,
+              sampling_enabled = true
+            where singleton;
+            select ash._register_wait('active', 'Issue81', 'RotateLoss');
+SQL
+          }
+
+          seed_endangered_sample() {
+            "${psql_base[@]}" << 'SQL'
+            insert into ash.sample (
+              sample_ts, datid, active_count, data, slot
+            )
+            values (
+              ash.ts_from_timestamptz(date_trunc('minute', now()) - interval '2 minutes'),
+              0::oid,
+              1::smallint,
+              array[-1,1,0]::integer[],
+              2::smallint
+            );
+            select count(*) as seeded_rows from ash.sample_2;
+SQL
+          }
+
+          assert_raw_sample_survived() {
+            local context="$1"
+            local raw_rows
+            local rollup_rows
+            local current_slot
+
+            raw_rows="$("${psql_base[@]}" -tAc "select count(*) from ash.sample_2")"
+            rollup_rows="$("${psql_base[@]}" -tAc "select count(*) from ash.rollup_1m")"
+            current_slot="$("${psql_base[@]}" -tAc "select current_slot from ash.config where singleton")"
+
+            if [ "$raw_rows" != "1" ]; then
+              echo "::error::$context: endangered raw sample was truncated (sample_2 rows=$raw_rows, rollup_1m rows=$rollup_rows, current_slot=$current_slot)" >&2
+              exit 1
+            fi
+            if [ "$current_slot" != "0" ]; then
+              echo "::error::$context: rotate advanced current_slot despite failed pre-rollup (current_slot=$current_slot)" >&2
+              exit 1
+            fi
+          }
+
+          wait_until_sql_true() {
+            local description="$1"
+            local sql="$2"
+            local value
+
+            for _ in {1..100}; do
+              value="$("${psql_base[@]}" -tAc "$sql")"
+              if [ "$value" = "t" ]; then
+                return 0
+              fi
+              sleep 0.1
+            done
+
+            echo "::error::timed out waiting for $description" >&2
+            exit 1
+          }
+
+          echo "=== #81A: rollup_minute must not advance while sampler lock is held ==="
+          reset_issue_81_state
+          "${psql_base[@]}" << 'SQL'
+          insert into ash.sample (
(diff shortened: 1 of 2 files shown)


r314tive:
Superseded by #86. Same fix; recreated only to use the repository-style branch name `fix-issue-81-rollup-rotation-data-loss` instead of the earlier account-prefixed branch name.

PR #105, opened 2026-06-10
ci: guard released SQL workflow

NikolayS:
## Summary

- add a CI guard that rejects ordinary PRs changing released `sql/*.sql`
- allow released SQL edits only from release-stamp PRs (`release/*`, `release-*`, or `release:` title)
- teach `ash_sql_chain.py` to report the version of the selected fresh installer so future `devel/sql/ash-install.sql` can carry the next development version without breaking CI assertions
- document the guard in the release-process and development SQL staging docs

## Testing

- `python3 devel/scripts/ash_sql_chain.py fresh-install-version`
- `python3 devel/scripts/ash_sql_chain.py latest-released-version`
- `python3 devel/scripts/ash_sql_chain.py full-upgrade-chain`
- temp-copy smoke: staged `devel/sql/ash-install.sql` with version `1.6`, confirmed `fresh-install-path` selects `devel/sql/ash-install.sql` and `fresh-install-version` returns `1.6`
- `bash -n` on the new CI guard shell block

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -20,6 +20,31 @@ jobs:
         with:
           fetch-depth: 0
 
+      - name: Guard SQL release workflow
+        if: github.event_name == 'pull_request'
+        env:
+          BASE_SHA: ${{ github.event.pull_request.base.sha }}
+          HEAD_REF: ${{ github.head_ref }}
+          PR_TITLE: ${{ github.event.pull_request.title }}
+        run: |
+          set -euo pipefail
+
+          changed="$(git diff --name-only "$BASE_SHA"...HEAD)"
+
+          if printf '%s\n' "$changed" | grep -E '^sql/.*\.sql$' >/dev/null; then
+            case "$HEAD_REF:$PR_TITLE" in
+              release/*:*|release-*:*|*:release:*)
+                ;;
+              *)
+                echo "Released SQL under sql/ is frozen between releases." >&2
+                echo "Feature/fix PRs must stage SQL in devel/sql/ and let a release-stamp PR promote it." >&2
+                printf 'Changed released SQL files:\n%s\n' \
+                  "$(printf '%s\n' "$changed" | grep -E '^sql/.*\.sql$')" >&2
+                exit 1
+                ;;
+            esac
+          fi
+
       - name: Guard docs against stale privilege and reader names
         run: |
           set -euo pipefail
@@ -246,7 +271,7 @@ jobs:
           PGPASSWORD: postgres
         run: |
           psql -h localhost -U postgres -d postgres -v ON_ERROR_STOP=1 \
-            -v expected_version="$(python3 devel/scripts/ash_sql_chain.py latest-released-version)" << 'EOF'
+            -v expected_version="$(python3 devel/scripts/ash_sql_chain.py fresh-install-version)" << 'EOF'
           select set_config('pg_ash.expected_version', :'expected_version', false);
 
           do $$
@@ -4290,7 +4315,7 @@ jobs:
           python3 devel/scripts/ash_sql_chain.py fresh-install-path | sed 's#^#\\i #' > /tmp/ash_fresh_install.sql
           python3 devel/scripts/ash_sql_chain.py reapply-chain > /tmp/ash_reapply_chain.sql
           psql -h localhost -U postgres -d postgres -v ON_ERROR_STOP=1 \
-            -v expected_version="$(python3 devel/scripts/ash_sql_chain.py latest-released-version)" << 'EOF'
+            -v expected_version="$(python3 devel/scripts/ash_sql_chain.py fresh-install-version)" << 'EOF'
           select set_config('pg_ash.expected_version', :'expected_version', false);
 
           \i /tmp/ash_fresh_install.sql
--- a/devel/scripts/ash_sql_chain.py
+++ b/devel/scripts/ash_sql_chain.py
@@ -12,6 +12,7 @@ ROOT = Path(__file__).resolve().parents[2]
 SQL_DIRS = (ROOT / "sql", ROOT / "devel" / "sql")
 INSTALL_RE = re.compile(r"ash-(\d+)\.(\d+)\.sql$")
 UPGRADE_RE = re.compile(r"ash-(\d+\.\d+)-to-(\d+\.\d+)\.sql$")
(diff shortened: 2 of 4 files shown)


NikolayS:
## REV Code Review Report

- **PR:** [NikolayS/pg_ash#105](https://github.com/NikolayS/pg_ash/pull/105) — ci: guard released SQL workflow
- **Author:** @NikolayS
- **Branch:** `leo/release-workflow-guard`
- **CI:** ✅ success (all 10 checks green: test 14–18, docs-lint, CodeQL, Analyze)

Reviewed by 5 parallel agents (security, bugs, tests, guidelines, docs). The change is well-scoped and sound: untrusted CI inputs are correctly env-isolated (no Actions injection), the `git diff base...HEAD` range and `grep '^sql/.*\.sql$'` anchoring are correct, and the `latest-released-version → fresh-install-version` assertion swap is a genuine **correctness improvement** — it makes the fresh-install regression and re-apply tests assert against the installer they actually run, and surfaces a dev-cycle inconsistency the old value silently masked. No blocking issues. Findings below are all low-severity hardening / doc-accuracy items.

> Note: this PR does not modify `sql/*.sql`, so the new guard does not fire on itself — self-consistent.

---

### NON-BLOCKING (5)

**LOW** `.github/workflows/test.yml:36` — Title match is broader than documented
> The arm `*:release:*` matches `"$HEAD_REF:$PR_TITLE"`, i.e. the substring `:release:` *anywhere*. Docs say the title must be **"starting with `release:`"**. For real titles the common case behaves as documented (the `:` separator anchors it), and branch names can't contain `:`, but a title like `chore: tweak :release: notes` would also bypass the freeze. Verified empirically: `ref=feat/x`, title `chore: tweak :release: notes` → ALLOW.
> **Fix:** match the fields separately instead of concatenating —
> `case "$HEAD_REF" in release/*|release-*) ok=1 ;; esac` and `case "$PR_TITLE" in release:*) ok=1 ;; esac`. This enforces "title starts with `release:`" exactly and matches the docs.

**LOW** `.github/workflows/test.yml:287,4329,4343` — Stale assertion messages
> The three assertion messages still read "version should match **latest released** version", but the compared value now comes from `fresh-install-version` (the fresh installer's own declared version). During a dev cycle `fresh-install-path` resolves to `devel/sql/ash-install.sql`, whose version is the *next, unreleased* version — so the message is misleading exactly when it would fire.
> **Fix:** reword to "version should match the fresh installer's declared version" across all three lines.

**LOW** `devel/scripts/ash_sql_chain.py:15` — `VERSION_DEFAULT_RE` relies on first-match ordering
> `version\s+text\s+not\s+null\s+default\s+'([^']+)'` matches two lines in `sql/ash-install.sql`: the `ash.config` table column (line 87, `'1.5'` — correct) and a later migration backfill (line 4769, `add column version text not null default '1.3'`). `.search()` returns the first, so today it correctly yields `1.5` — but only because the CREATE precedes the ALTER. A future reorder, or any `*version text … default '…'` column added above line 87, would silently return the wrong value with no error.
> **Fix:** anchor the regex, e.g. `re.compile(r"(?m)^\s*version\s+text\s+not\s+null\s+default\s+'([^']+)'")` (or scope to the `create table ash.config` block) so only the standalone column definition matches regardless of file order.

**LOW** `devel/scripts/ash_sql_chain.py:87` / `.github/workflows/test.yml:23` — New logic ships without automated tests
> CLAUDE.md mandates TDD for new features ("write tests … before or alongside the implementation"). The new `fresh_install_version()` function and the new CI guard `case` logic have no automated test — the repo has no Python test infra at all, and the PR description lists only manual smoke testing. The function is exercised only indirectly (its output feeds an existing `ash.config.version = expected_version` round-trip, which is somewhat self-referential), and the guard's branch/title logic only runs against real PRs.
> **Suggestion:** extract the guard into a small script (`HEAD_REF`/`PR_TITLE`/changed-files as args) and add a table-driven test — `sql/ash-install.sql + feat/x → fail`, `+ release/1.6 → pass`, `+ release: title → pass`, `devel/sql/x + feat/x → pass` — plus a tiny unit test pinning `fresh_install_version()` against a fixture containing both a config-default and a migration-default line. Neither needs Postgres.

**INFO** `docs/RELEASE_PROCESS.md:31` — Guard is a discipline control, not a security boundary
> The guard is trivially bypassable by a PR author (they control branch name and title), so it cannot stop a *determined* edit of frozen `sql/`. This is fine: the workflow runs on `pull_request` (not `pull_request_target`), holds only `permissions: contents: read`, uses no secrets, and gates no deploy/tag automation — a bypass grants nothing. It correctly catches *accidental* edits by trusted contributors.
> **Suggestion:** keep the docs from implying tamper-resistance; real integrity of `main`'s `sql/` should come from branch protection / CODEOWNERS on `sql/`, not this string match.

---

### Cleared (no finding)

- **Actions injection:** none — `github.head_ref`, `pull_request.title`, `base.sha` all passed via `env:` and referenced as quoted `"$VAR"`, never inlined as `${{ }}`. ✓ recommended pattern.
- **`git diff "$BASE_SHA"...HEAD`:** correct — on `pull_request`, HEAD is the merge ref and `base.sha` is its ancestor, so the 3-dot range equals exactly the PR's changes; `fetch-depth: 0` guarantees `base.sha` is present.
- **`grep '^sql/.*\.sql$'`:** correctly anchored — excludes `devel/sql/…`, catches nested + deletions.
- **`set -euo pipefail` + pipefail in the `if`:** safe (the `grep` exit is the `if` condition, exempt from `set -e`).
- **Regex ReDoS:** none (linear `[^']+`, no nested quantifiers); operates on repo-controlled files only.
- **Re-apply assertion change:** verified a correctness *improvement*, not a regression — under the established dev convention both helpers agree; in an off-convention state the new value correctly fails where the old one masked the mismatch.
- **Python style / docs placement / emojis / whitespace / naming:** clean and consistent with surrounding code.

---

### Summary

| CI | 0 (green) | — | — |
| Security | 1 (INFO) | 0 | injection, ReDoS, secret exposure |
| Bugs | 2 (LOW) | 0 | diff range, pipefail, assertion-swap (verified correct) |
| Tests | 1 (LOW) | 0 | — |
| Guidelines | 0 (overlaps title finding) | 0 | style, naming, emojis |
| Docs | 1 (LOW) | 0 | subcommand doc placement OK |

**Verdict:** Mergeable. No blocking issues — CI is green and the core change is correct (and a net improvement). The five items above are optional hardening: anchor the title match (#1) and the version regex (#3), refresh the stale assertion messages (#2), and ideally add a small guard/helper test (#4).

---

Issue #107, opened 2026-06-10
Bug hunt: existing pg_ash correctness issues


NikolayS:
Tracking issue for the existing-code bug hunt requested on 2026-06-10.

Scope: current `main` after release-workflow PR https://github.com/NikolayS/pg_ash/pull/105, excluding the AAS feature PR https://github.com/NikolayS/pg_ash/pull/106.

Confirmed bugs

1. pg_stat_statements joins can duplicate ASH reader rows

Affected readers: `ash.samples()`, `ash.samples_at()`, `ash.event_queries()`, `ash.event_queries_at()`.

Cause: these functions join directly to `pg_stat_statements`. The pgss view can contain more than one row for the same `queryid` because rows are keyed by `(userid, dbid, queryid, toplevel)`. When ASH has one sample for that `queryid`, direct joins can multiply it into multiple output rows.

Reproducer evidence on current `main` (`82ed28e2dbc66328d14337920531c54c9c8377d7`): in a PG18 container with `pg_stat_statements` preloaded, running the same normalized query as two roles produced two pgss rows for one `queryid`; after seeding one ASH sample for that query:

```text
pgss_duplicate_rows | -3688696628780506391 | 2
samples_rows        | 2
samples_detail      | -3688696628780506391 | 2
event_queries_rows  | 2
event_queries_detail| -3688696628780506391 | 1 | 100.00 | 2
event_queries_detail| -3688696628780506391 | 1 | 100.00 | 2
```

Expected: one ASH sample should produce one decoded sample row and one event-query row. Query text should be joined from a pre-aggregated pgss relation, not by multiplying ASH rows.

Planned PR split

- PR 1: red/green fix for pgss fan-out in `samples*` and `event_queries*`. First add a CI regression that fails on current `main`, then pre-aggregate pgss rows before joining.

Pending

A separate Fable-high audit is still running with subagents for core SQL correctness, reporting functions, upgrade-chain integrity, CI/scripts/release tooling, and docs/config behavior. Any additional confirmed bugs will be appended here, and each confirmed bug should get its own red/green PR.

PR #106, opened 2026-06-10
feat: add AAS drilldowns

NikolayS:
## Summary

- starts the next SQL development cycle in `devel/sql/` as v1.6
- adds AAS summary readers over minute rollups:
  - `ash.aas_summary()` / `ash.aas_summary_at()` for arbitrary periods
  - `ash.aas_periods()` for 1 minute, 5 minutes, 1 hour, 1 day, 1 week, 1 month (30 days)
- adds drill-down readers:
  - `ash.aas_wait_types()` / `_at()`
  - `ash.aas_wait_events()` / `_at()`
  - `ash.aas_queries()` / `_at()` with pg_stat_statements query text when available
- snaps trailing-period helpers to the current minute boundary so `1 minute` means the last completed minute rollup, not a partial wall-clock slice that may miss the latest rollup row
- documents a full copy-paste AAS usage list in README: standard windows, each standard trailing period, arbitrary trailing/absolute periods, standard periods anchored to an explicit timestamp, wait-type drill-down, wait-event drill-down, query drill-down, and exact incident-window drill-downs
- expands CI assertions with deterministic rollup fixtures that cover public function existence, standard periods, arbitrary ranges, avg/peak/p99 math, drill-down percentages, limits, empty windows, invalid inputs, and no-pg_stat_statements query behavior

AAS math: `avg_aas = backend_seconds / elapsed wall-clock seconds`; missed/empty minutes count as zero activity; `peak_aas = max(rollup_1m.peak_backends)`; `p99_1m_aas = percentile_cont(0.99)` over zero-filled per-minute AAS values.

Follow-up commit 8bb84f1 (REV review fixes, red/green TDD):

- fixes `aas_summary_at` returning NULL instead of 0 for `peak_aas`/`p99_1m_aas` on windows that never cross a minute boundary (zero-fill contract)
- new CI asserts: multi-datid minutes (peak = max across databases, not sum), non-minute-aligned start windows, relative wrappers' minute-boundary snap discrimination, negative interval validation, NULL limit pinning, ordering under limit, `CPU*` no-colon label branch, and the positive pg_stat_statements query-text path
- restores the standard pgss trust comments on `aas_queries`/`aas_queries_at` and the development-cycle header on the devel installer
- README: two-row table pattern for relative/absolute pairs, defaults documented, `1 month (30 days)` wording aligned with the `aas_periods` label

## Testing

- `git diff --check`
[...]

PR #108, opened 2026-06-10
fix: prevent pgss fanout in ASH readers

NikolayS:
Fixes the first confirmed item from https://github.com/NikolayS/pg_ash/issues/107.

## Problem

`pg_stat_statements` can have multiple rows for one `queryid` because rows are keyed by `(userid, dbid, queryid, toplevel)`. `ash.samples()`, `ash.samples_at()`, `ash.event_queries()`, and `ash.event_queries_at()` joined directly to pgss, so one ASH sample/query could fan out into duplicate output rows.

## Red/green evidence

Red first:

- Added `Regression: pg_stat_statements joins do not duplicate ASH rows` to CI.
- Before the fix, the same scenario failed locally on current `main` with:

```text
ERROR: samples() should return one ASH row for one sample; got 4 with 2 pgss rows
```

Green after fix:

- The affected readers now pre-aggregate pgss rows before joining.
- Local PG18 targeted regression passes:

```text
NOTICE: pg_stat_statements fanout regression test PASSED
```

## Local checks

- `git diff --check`
- `python3 devel/scripts/ash_sql_chain.py fresh-install-path` -> `devel/sql/ash-install.sql`
- `python3 devel/scripts/ash_sql_chain.py fresh-install-version` -> `1.6`
- `python3 devel/scripts/ash_sql_chain.py full-upgrade-chain` includes `devel/sql/ash-1.5-to-1.6.sql`

## Notes

This PR intentionally stages SQL in `devel/sql/` per https://github.com/NikolayS/pg_ash/pull/105 release workflow guard. It is independent from the AAS feature PR https://github.com/NikolayS/pg_ash/pull/106, though both currently start the 1.6 development SQL area and will need ordering/rebase before merge.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -1676,6 +1676,110 @@ jobs:
           $$;
           EOF
 
+      - name: "Regression: pg_stat_statements joins do not duplicate ASH rows"
+        env:
+          PGPASSWORD: postgres
+        run: |
+          psql -h localhost -U postgres -d postgres -v ON_ERROR_STOP=1 << 'EOF'
+          drop role if exists pgash_pgss_fanout_a;
+          drop role if exists pgash_pgss_fanout_b;
+          create role pgash_pgss_fanout_a;
+          create role pgash_pgss_fanout_b;
+          grant pgash_pgss_fanout_a, pgash_pgss_fanout_b to postgres;
+
+          select pg_stat_statements_reset();
+
+          set role pgash_pgss_fanout_a;
+          select 424242::int as pgash_fanout_probe;
+          reset role;
+
+          set role pgash_pgss_fanout_b;
+          select 424242::int as pgash_fanout_probe;
+          reset role;
+
+          do $$
+          declare
+            v_query_id bigint;
+            v_pgss_rows int;
+            v_wait_id smallint;
+            v_map_id int4;
+            v_sample_ts int4;
+            v_datid oid := (select oid from pg_database where datname = current_database());
+            v_count int;
+          begin
+            select queryid, count(*)
+            into v_query_id, v_pgss_rows
+            from pg_stat_statements
+            where query like '%pgash_fanout_probe%'
+            group by queryid
+            having count(*) >= 2
+            limit 1;
+
+            assert v_query_id is not null,
+              'pg_stat_statements did not record duplicate rows for the fanout probe';
+            assert v_pgss_rows >= 2,
+              'expected at least two pgss rows for same queryid, got ' || coalesce(v_pgss_rows, 0);
+
+            insert into ash.wait_event_map (state, type, event)
+            values ('active', 'CPU', 'CPU')
+            on conflict (state, type, event) do update set state = excluded.state
+            returning id into v_wait_id;
+
+            insert into ash.query_map_0 (query_id)
+            values (v_query_id)
+            on conflict (query_id) do update set query_id = excluded.query_id
+            returning id into v_map_id;
+
+            v_sample_ts := extract(epoch from now() - ash.epoch())::int4;
+            insert into ash.sample (sample_ts, datid, active_count, data, slot)
+            values (v_sample_ts, v_datid, 1, array[-v_wait_id::int, 1, v_map_id]::int[], 0);
+
+            select count(*)
+            into v_count
+            from ash.samples('1 hour', 1000)
+            where query_id = v_query_id;
+            assert v_count = 1,
+              format('samples() should return one ASH row for one sample; got %s with %s pgss rows',
+                     v_count, v_pgss_rows);
+
+            select count(*)
+            into v_count
+            from ash.samples_at(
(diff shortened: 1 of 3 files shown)


NikolayS:
## REV Code Review Report — PR #108: `fix: prevent pgss fanout in ASH readers`

**Verdict: APPROVE — no blockers.** Red/green bugfix is correct, well-tested, and behaviorally verified against real Postgres.

**Method:** Opus-level REV review across 5 dimensions (security, runtime correctness, test/CI, project guidelines/style, docs/operator UX). Findings behaviorally verified against real Postgres 17 in Docker (fresh install, upgrade chain, idempotent re-apply, red/green reproduction). SOC2-specific checks intentionally skipped per scope.

### Summary
`pg_stat_statements` is keyed by `(userid, dbid, queryid, toplevel)`, so one `queryid` can have many rows. `ash.samples()`, `ash.samples_at()`, `ash.event_queries()`, and `ash.event_queries_at()` joined pgss directly on `queryid`, multiplying each ASH row by the number of pgss rows. The fix pre-aggregates pgss to one row per key before joining.

### Behavioral verification (Docker, PG17, pg_stat_statements loaded)
Two roles run the same probe query → 2 pgss rows for one queryid; a synthetic ASH sample referencing that queryid is injected.

- **RED (baseline `main` installer):** `ERROR: samples() got 2 with 2 pgss rows` — bug reproduced.
- **GREEN (PR `devel/sql/ash-install.sql`):** `pg_stat_statements fanout regression test PASSED` — all four readers return exactly 1 row.
- **Full upgrade chain** `1.0 → … → 1.5 → devel 1.6`: OK, `ash.config.version = 1.6`, and the regression test passes on the upgraded schema too.
- **Idempotent re-apply** of the devel installer: OK.
- **CI:** all 10 checks green (test matrix PG 14–18 incl. pgss-off, CodeQL, docs-lint, analyzers).

### Correctness review of the fix (SQL semantics)
- **No fan-out:** `samples()`/`samples_at()` pre-aggregate `group by p.queryid, p.dbid` and join on both `queryid` and `dbid` → at most one pgss match per decoded row. `event_queries()`/`event_queries_at()` are query_id-level and pre-aggregate `group by p.queryid` → one match per queryid. ✔
- **Missing pgss rows do not drop ASH rows:** all joins remain `LEFT JOIN`; a non-match yields `NULL`/truncated text while preserving the ASH row. ✔
- **query_text stays best-effort:** `min(p.query)` collapses across `userid`/`toplevel`; normalized text is identical across those keys for a given `queryid`, so `min()` is a safe representative. ✔
- **dbid/user/toplevel behavior is defensible:** the per-database readers (`samples*`) now scope text by `dbid = d.datid`, correctly attributing text to the sample's database; the query_id-level readers (`event_queries*`) intentionally ignore `dbid`, matching their output shape. ✔
- **Convention-consistent:** mirrors the existing pre-aggregation pattern already used by `top_queries`/`samples_by_database` in the released baseline, extending it with `dbid` precision where the reader exposes a database dimension. ✔
- **Degraded mode unchanged:** the no-pgss branches are untouched (NULL text + the existing warning/exception). ✔

### Security
No new injected SQL, no dynamic SQL, no privilege change. Functions keep `set search_path = pg_catalog, ash, public` and `set jit = off`. pgss access stays guarded by `_pgss_schema()` + a probe inside a sub-block. No concerns.

### Required checks (all pass)
- `git diff --check` → clean
- `ash_sql_chain.py fresh-install-path` → `devel/sql/ash-install.sql`
- `ash_sql_chain.py fresh-install-version` → `1.6`
- `ash_sql_chain.py full-upgrade-chain` → ends with `\i devel/sql/ash-1.5-to-1.6.sql`
- `gh pr checks 108` → all green

### Non-blocking notes
1. **Merge ordering vs PR #106 (acknowledged in PR body):** both PRs initialize the 1.6 `devel/sql/` area, so a rebase/conflict-resolution is required before whichever merges second. Coordination item, not a code defect.
2. **Dev upgrade wrapper:** `devel/sql/ash-1.5-to-1.6.sql` replays the full installer via `\ir ash-install.sql`. Correct and schema-equivalent for the dev cycle (verified), but at release-stamp time it must be promoted to a real incremental `1.5 → 1.6` delta per `docs/RELEASE_PROCESS.md`.
3. **Test hygiene (minor):** the regression step leaves the two `pgash_pgss_fanout_*` roles and the synthetic `ash.sample`/`wait_event_map`/`query_map_0` rows in place. Consistent with other steps in `test.yml` and harmless (CI green across all versions), but a trailing cleanup would make the step self-contained.
4. **Behavioral subtlety (improvement, not a regression):** in a multi-database deployment, if a `queryid` exists in pgss only under a different `dbid` than the sample's `datid`, `samples*` now returns `NULL` text instead of borrowing another database's text. This is more correct and also what eliminates the fan-out.

**Recommendation:** Mergeable on its own merits after resolving #106 ordering. 🤖 Opus REV review.

PR #109, opened 2026-06-10
docs: clarify pgss query text privileges

NikolayS:
## Summary

- add a terse README note: `ash.grant_reader()` does not grant `pg_read_all_stats`
- show the symptom and fix for monitoring roles: `query_id` present, `query_text` NULL
- add the pg_stat_statements text-lifetime caveat: reset/eviction can lose text even when pg_ash still has `query_id`

## Testing

- `git diff --check`
- docs-lint guard subset from `.github/workflows/test.yml`

### DIFF
--- a/README.md
+++ b/README.md
@@ -929,6 +929,13 @@ quote the role name, and emit a `RAISE NOTICE` summarizing what changed.
 
 **Note on pg_cron visibility:** `ash.grant_reader()` does **not** grant `USAGE ON SCHEMA cron`, since pg_cron is not an `ash` object. When pg_cron is loaded but the monitoring role lacks USAGE on schema `cron`, `ash.status()` emits a single fallback row of the form `cron_jobs = '<no cron.job access; grant USAGE ON SCHEMA cron TO <role>>'` instead of per-job `cron_job_*` rows. To surface real cron job details, either run `grant usage on schema cron to <role>` (and `grant select on cron.job to <role>`) once, or simply ignore the row.
 
+**Query text visibility:** `ash.grant_reader()` does not grant `pg_read_all_stats`. Monitoring roles need it to resolve other users' SQL text from `pg_stat_statements`; otherwise pg_ash may show `query_id` with NULL `query_text`.
+
+```sql
+select pg_has_role(current_user, 'pg_read_all_stats', 'member');
+grant pg_read_all_stats to my_monitor_role;
+```
+
 Prefer `ash.grant_reader()` for full monitoring access. If you need manual control, grant only the specific readers and tables the role actually needs — do **not** use blanket `EXECUTE` grants on the whole `ash` schema, because that can include owner-only maintenance helpers now or after future upgrades.
 
 Minimal example for a dashboard that only calls `ash.status()` and `ash.top_waits()`:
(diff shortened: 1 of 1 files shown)


NikolayS:
## REV Code Review Report — PR #109

**Method:** Opus REV agents (security, runtime correctness, test/CI, project guidelines, docs/operator UX). SOC2-only checks skipped. Claims verified empirically against PostgreSQL 17 + pg_stat_statements in Docker.

**Verdict:** ✅ Approve (one minor correctness fix applied during review — see below).

### Scope
Docs-only: README note that `ash.grant_reader()` does not grant `pg_read_all_stats`, the `query_id`-present/`query_text`-NULL symptom, and the pgss text-lifetime (reset/eviction) caveat.

### Findings

**Accuracy confirmed (empirical).** For a role lacking `pg_read_all_stats`, PG masks **both** `query` (→ `<insufficient privilege>`) **and `queryid` (→ NULL)** for other users' statements. pg_ash resolves text via `left join pg_stat_statements pss on pss.queryid = r.query_id`; a NULL `queryid` never matches, so the join yields a genuinely **NULL** `query_text` while pg_ash's own `query_id` (from `ash.query_map_all`) is still present. The documented symptom is correct. Granting `pg_read_all_stats` restores visibility — verified.

**Fix applied (minor, by reviewer):** the diagnostic used `pg_has_role(current_user, 'pg_read_all_stats', 'member')`. pgss gates visibility on `has_privs_of_role` (inheritance-sensitive), which corresponds to `pg_has_role(..., 'usage')`, not `'member'`. A **NOINHERIT** role member of `pg_read_all_stats` returns `member=true` but `usage=false` and still sees NULL `queryid`/text — so `'member'` reports "privileges correct" and steers the operator to a false "text was evicted" conclusion. Changed both occurrences to `'usage'` (commit `349d6a8`). Verified: NOINHERIT → `member=t, usage=f, can_see_qid=NULL`; default INHERIT and superuser → `usage=t`.

- Security/privacy: no regression — guidance is opt-in operator advice; `grant_reader()` deliberately does not auto-grant a cluster-wide stats role. ✓
- Bug/runtime: troubleshooting `where queryid = ...` lookup is sound *given* the privilege check above (which now correctly precedes it). ✓
- Style/guidelines: docs-only, no SQL touched; complies with the frozen-`sql/` release guard. ✓
- Operator UX: symptom→check→fix flow is clear and actionable. ✓

### Checks
- `git diff --check`: clean
- docs-lint guard subset (blanket-EXECUTE, stale `waits_by_type`): pass
- CI: docs-lint, `test (14–18, on/off)`, Analyze (actions) — **green**. `Analyze (python)` failed on a transient GitHub infra/auth error (`Requires authentication` during CodeQL init/feature-enablement, no repo content analyzed) — unrelated to this diff; recommend re-run.

No blockers.

NikolayS:
CI follow-up after Opus REV:

The first CodeQL Python run failed in GitHub init with `Requires authentication` and could not be retried. I force-pushed the same docs change as fresh head `6239361` to trigger a new run.

Current status on `6239361`: all checks green, including CodeQL, docs-lint, and the PG 14-18 SQL matrix.

PR #110, opened 2026-06-10
fix: preserve reader EXECUTE grants across installer drop/recreate (#107)

NikolayS:
Fixes #107 (bug-hunt B1).

## Problem

The installer drops every `ash.*` function whose signature changed across versions before recreating it (top-of-file DO block plus explicit `drop function if exists` for `top_queries_with_text` / `event_queries` / `event_queries_at`). `DROP FUNCTION` destroys ACLs that `CREATE OR REPLACE` would have preserved, so the documented upgrade path (`v1.4 -> v1.5` replays the installer) and **any** `ash-install.sql` re-apply silently revoked reader EXECUTE grants handed out via `ash.grant_reader()`. A Grafana-style monitoring role loses access to `top_waits/_at`, `timeline_chart/_at`, `query_waits/_at`, `top_by_type/_at`, `event_queries/_at`, `top_queries_with_text` and gets `permission denied` on its next dashboard refresh.

## Fix (staged in `devel/sql/`, `sql/` untouched per the release freeze)

- Snapshot every explicit non-owner EXECUTE grant on `ash.*` functions into a temp table **before** the drops; re-apply them (including grant option) by function name to every recreated overload at the **end** of the installer.
- Faithful-restore semantics, mirroring what `CREATE OR REPLACE` would have kept: PUBLIC stays excluded (the REVOKE-from-PUBLIC hardening posture is unchanged), removed/renamed function names and dropped roles are skipped, and a role's reach is never widened — admin functions stay admin-only, and functions newly introduced by a release still require a fresh `ash.grant_reader()` call.
- This PR also bootstraps the 1.6 development cycle (`devel/sql/ash-install.sql` + `devel/sql/ash-1.5-to-1.6.sql`); the bootstrap is byte-identical to the one in PR #106 to keep the eventual merge clean.

## CI hardening (B4-style)

- New step **"Issue #107: reader grants survive upgrade and installer re-apply"**: installs the discovered *released* chain, configures a reader role via `ash.grant_reader()`, applies the in-development upgrade, then asserts (a) a non-enumerated catalog sweep — every function name the role could execute before the upgrade and that still exists must remain executable on every current overload, (b) signature-pinned `has_function_privilege` checks for all eleven functions from the bug-hunt findings, (c) no-escalation checks (admin functions, `grant_reader`/`revoke_reader`, PUBLIC), (d) behavioral reader calls under `set role`, and (e) the same sweep + behavioral call again after an installer re-apply.
[...]

PR #111, opened 2026-06-13
fix: keep installer grant restore least-privilege (#107 follow-up)

NikolayS:
## Summary

Follow-up hardening for the #107 grant-preservation fix shipped in PR #110. During pre-1.6 release-gate validation (manual red/green on real Postgres + a REV-style multi-agent review of #108/#110), the #110 grant snapshot/restore was found to **widen privileges** on every installer re-apply:

1. **Admin-function grants re-established after hardening.** The restore had no `ash._admin_funcs()` filter and runs as the *last* block in the installer — after the admin-hardening block. A stray non-owner `EXECUTE` grant on an admin function in the drop list (e.g. `ash.rebuild_partitions`, `ash.uninstall`) was snapshotted, scrubbed by `DROP FUNCTION`, then **re-granted** afterward, silently undoing the "admin functions stay admin-only on every install" posture.
2. **Overload widening.** The restore re-granted by *name* to every overload. A role holding a single overload of `ash.decode_sample` was widened to all overloads — contradicting the restore's own documented "a role's reach is never widened" contract.

Neither is exploitable for a properly `ash.grant_reader()`-configured role (it never hands out admin functions and already grants every reader overload), but both weaken the documented least-privilege posture and are **latent escalation paths** for any future reader→admin migration. This is a regression vs. pre-#110 behaviour, where `DROP FUNCTION` scrubbed such grants.

## Reproduction (real Postgres, PG17)

| Scenario | pre-#110 (released `sql/`) | post-#110 (`main` `devel/`) | this PR |
|---|---|---|---|
| Stray grant on `rebuild_partitions` survives re-apply | `f` (scrubbed) | **`t` (regression)** | `f` (scrubbed) |
| Single `decode_sample` overload widens to sibling | n/a | **`t` (widened)** | `f` (not widened) |
| `grant_reader` reader grants survive re-apply | `t` | `t` | `t` |

## Fix

Mirrors `grant_reader`'s own filter:
- exclude `ash._admin_funcs()` from the restore (placed in the end-of-script block where that function is defined; the top-of-script snapshot cannot call it);
- capture the argument signature in the snapshot and restore the **exact** overload when it still exists, falling back to by-name for every current overload **only** when the snapshotted signature is gone — the genuine cross-version signature change the by-name restore exists to cover. So #107 reader-grant preservation across real upgrade chains is unaffected (verified: 1.0→1.5→1.6 + re-apply keeps all reader grants).

Staged in `devel/sql/` per the release freeze; `sql/` untouched.

## Tests

Adds a CI regression (`Issue #107 follow-up: grant restore stays least-privilege`) asserting, across an installer re-apply: legitimate reader grants survive, a stray admin grant is scrubbed, a single-overload grant does not widen, reader cannot reach admin functions, and PUBLIC stays revoked. Red on `main` (`admin EXECUTE grant re-established after hardening`), green with this fix. Validated locally on PG 16, 17, 18.

> Do not merge without project-owner approval and a full release-gate re-run.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -4610,6 +4610,84 @@ jobs:
           drop role ash_b1_reader;
           EOF
 
+      - name: "Issue #107 follow-up: grant restore stays least-privilege"
+        env:
+          PGPASSWORD: postgres
+        run: |
+          # The #107 grant snapshot/restore must not become a privilege-widening
+          # vector. On every installer re-apply it must:
+          #   (a) NOT re-establish a stray non-owner EXECUTE grant on an admin
+          #       function after the hardening block revoked it (before #107,
+          #       DROP FUNCTION scrubbed such grants);
+          #   (b) NOT widen a role that held a single overload of a function to
+          #       its sibling overloads (restore by exact signature);
+          #   while still preserving legitimate ash.grant_reader() grants.
+          python3 devel/scripts/ash_sql_chain.py fresh-install-path | sed 's#^#\\i #' > /tmp/ash_fresh_install.sql
+          psql -h localhost -U postgres -d postgres -v ON_ERROR_STOP=1 << 'EOF'
+          \i /tmp/ash_fresh_install.sql
+
+          do $$ begin
+            if not exists (select from pg_roles where rolname = 'lp_reader')   then create role lp_reader   login password 'x'; end if;
+            if not exists (select from pg_roles where rolname = 'lp_stray')    then create role lp_stray    login password 'x'; end if;
+            if not exists (select from pg_roles where rolname = 'lp_overload') then create role lp_overload login password 'x'; end if;
+          end $$;
+
+          -- legitimate reader: must SURVIVE re-apply (guards the #107 fix)
+          select ash.grant_reader('lp_reader');
+          -- stray manual grant on an admin function in the drop list: must be SCRUBBED
+          grant execute on function ash.rebuild_partitions(int, text) to lp_stray;
+          -- single-overload grant: must NOT widen to the sibling overload
+          grant execute on function ash.decode_sample(integer[], smallint) to lp_overload;
+
+          do $$ begin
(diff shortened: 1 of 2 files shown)


NikolayS:
## REV-assisted review — PRs #108 & #110 (merged bugfixes), pre-1.6 validation

Multi-agent REV review (5 dimensions × 2 PRs → adversarial verification): **10 raw findings, 9 confirmed, 1 refuted.** Every finding was independently reproduced on real Postgres before being treated as actionable.

### PR #108 (pgss fanout) — clean
All findings INFO/LOW and non-blocking: test exercises only the userid fanout dimension (fix's `GROUP BY` collapses all three, so correct), a fixture insert without `begin/rollback` (no actual flakiness — verified), and cosmetic alias/comment style nits. **Manually validated red→green:** pre-fix every reader returned 2 rows for 1 sample with 2 pgss rows; fixed returns exactly 1 (PG 16/17/18). No reader path was missed — all 7 pgss readers pre-aggregate.

### PR #110 (reader grant preservation) — one substantive regression → fixed in this PR
- **MEDIUM (confirmed, reproduced):** restore re-establishes a stray non-owner grant on an **admin** function after the hardening block → defeats admin-only posture. Regression vs pre-#110 (DROP used to scrub it).
- **LOW (confirmed, reproduced):** restore-by-name **widens** a single-overload grant (`decode_sample`) to all overloads.
- LOW/INFO coverage notes: WITH GRANT OPTION survival and degraded-mode grant survival are untested in CI (both manually verified working here).
- Refuted: "missing RELEASE_NOTES entry" — correct mid-cycle state per `docs/RELEASE_PROCESS.md` (notes written at release-stamp time).

This PR fixes the MEDIUM + LOW widening findings and adds the regression test. #107 reader-grant preservation across real upgrade chains is unaffected (verified 1.0→1.5→1.6 + re-apply).

*REV-assisted review (AI analysis adapted from [postgres-ai/rev](https://gitlab.com/postgres-ai/rev)). Findings reproduced on Postgres 17 in Docker.*

PR #123, opened 2026-07-10
fix: avoid stale rollup reads

### LINKED ISSUE #122: Bug: 2.0 readers prefer a stale rollup over complete raw data
## Summary

## Reproduction

The relevant results are:

```text

```

## Expected

## Impact

## Proposed regression

NikolayS:
## What changed

- adds a behavioral regression for a greater-than-one-hour window where raw samples cover the requested window but `rollup_1m` stops after its first minute
- makes `ash._pick_source_agg()` require the rollup watermark to reach `until` when raw covers `since`
- preserves the existing partial-rollup fallback when raw cannot answer the window start

## Why

The 2.0 aggregate readers selected `rollup_1m` from its retention start alone. If the rollup worker lagged or stopped, the readers returned plausible but incomplete values and omitted newer raw load.

This is a release-candidate SQL correction on a `release/*` branch because the beta installer is already stamped under `sql/`.

## Impact

The correction applies to the new aggregate surface that delegates to `_pick_source_agg()`: `aas`, `timeline`, `periods`, non-tie `top`, and `chart`.

## Validation

- RED on Postgres 18 at `54362ff`: `stale rollup must fall back to raw, got rollup_1m`
- GREEN on Postgres 18 at `6123780`: exact source, covered-minute, and backend-second assertions pass
- fresh install and full installer re-apply with `ON_ERROR_STOP=1` pass
- `git diff --check main...HEAD` passes

Closes #122

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -3712,6 +3712,69 @@ jobs:
           end $$;
           EOF
 
+      - name: "Test 2.0: stale rollup falls back to raw"
+        env:
+          PGPASSWORD: postgres
+        run: |
+          psql -h localhost -U postgres -d postgres -v ON_ERROR_STOP=1 << 'EOF'
+          -- A wide aggregate read may prefer rollup_1m only when that rollup
+          -- covers the requested end. If raw covers the whole window while the
+          -- rollup worker lags, choosing rollup silently drops the recent load.
+          do $$
+          declare
+            v_wait smallint;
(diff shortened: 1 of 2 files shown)


NikolayS:
Superseded by #199, which keeps stale wide rollup reads honest without losing the healthy rounded-up dashboard fast path and landed on current main as 9efad9ccb0eb9a6b8125eeb17ff9ff263c8606ea. Closing this earlier attempt in favor of that successor; issue #122 remains open for the explicitly unclaimed older-than-physical-raw completeness case.

PR #125, opened 2026-07-10
fix: make SQL entrypoints atomic

### LINKED ISSUE #124: Bug: documented SQL entrypoints can mask partial install failures
## Summary

The documented fresh-install and canonical migration entrypoints do not consistently fail fast, and the fresh installer is not transactional. A SQL error can therefore leave a partial pg_ash schema while `psql` exits with status 0.

This is a release blocker because automation and interactive users can treat a broken install or upgrade as successful.

## Fresh-install reproduction

On `origin/main` at `5aac2e3`, `sql/ash-install.sql` has neither `\set ON_ERROR_STOP on` nor a transaction. README Quick Start tells users to run it with plain `\i sql/ash-install.sql`.

Two Postgres 18 reproductions:

1. Running the installer as a role that cannot create the `ash` schema emitted 113 errors, continued to the end, printed a misleading final warning, and exited 0.
2. Pre-creating an incompatible `ash.config` view, then running `psql -f sql/ash-install.sql`, emitted 24 errors and exited 0. The database was left with 53 `ash` functions but no `ash.sample` table.

Adding `-v ON_ERROR_STOP=1` makes the command stop at the first error, but prior statements remain committed because the installer is not atomic.

## Canonical migration reproduction

README and `sql/migrations/README.md` advertise `sql/migrations/*` as the canonical upgrade path. Several canonical migrations do not set `ON_ERROR_STOP`; only the root compatibility wrappers do.

On Postgres 18, applying `sql/migrations/ash-1.1-to-1.2.sql` to an invalid precondition produced 19 errors and exited 0. Applying the root `sql/ash-1.1-to-1.2.sql` wrapper exited 3 after the first error.

## Expected

Every documented SQL entrypoint should:

- stop on the first SQL error without requiring caller-specific psql configuration;
- return a nonzero exit status to automation; and
- execute atomically so a failure does not leave a partial install or migration.

Migration wrappers that include the fresh installer must avoid nested transaction ownership.

## Proposed regression coverage

- Induce a deterministic mid-install failure and assert a nonzero exit plus rollback of all installer-created objects.
- Exercise canonical migration paths without external `ON_ERROR_STOP` and assert nonzero failure on an invalid precondition.
- Keep successful fresh install, full chain, and installer/migration re-apply coverage green.

NikolayS:
## What changed

- enables `ON_ERROR_STOP` inside the documented fresh installer and canonical migration entrypoints
- wraps the fresh installer and legacy inline migrations in a transaction
- lets migrations that replay `ash-install.sql` delegate transaction ownership to the installer, avoiding nested `BEGIN` / premature `COMMIT`
- adds negative CI coverage for a deterministic mid-install failure and an invalid canonical migration precondition

## Why

The README-supported paths could emit many SQL errors, continue, and exit 0. A failed fresh install could leave dozens of functions and several tables behind, so automation and interactive operators had no trustworthy success signal.

## Impact

Failed installs and migrations now stop at the first error, return a nonzero psql status, and roll back installer-created objects. Successful fresh installs, re-applies, and the historical upgrade chain retain their existing behavior.

## Validation

Postgres 18:

- RED at `9274ab9`: fresh installer and canonical migration both exited 0; the installer left 53 functions and three tables
- GREEN at `3a46c93`: both failure paths exit 3
- failed fresh install leaves zero functions and zero tables; the pre-existing failure fixture survives
- failed canonical migration leaves no `ash` schema
- fresh install and full installer re-apply pass at `2.0-beta1`
- discovered 1.0 → 2.0 upgrade chain passes and exposes all nine new reader names
- `git diff --check main...HEAD` passes

Closes #124

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -270,6 +270,81 @@ jobs:
         run: |
           psql -h localhost -U postgres -d postgres -v ON_ERROR_STOP=1 -f "$(python3 devel/scripts/ash_sql_chain.py fresh-install-path)"
 
+      - name: Test SQL entrypoints fail atomically
+        env:
+          PGPASSWORD: postgres
+        run: |
+          psql_base=(
+            psql
+            --no-psqlrc
+            --host=localhost
+            --username=postgres
+          )
+
+          "${psql_base[@]}" --dbname=postgres --set=ON_ERROR_STOP=1 << 'EOF'
+          drop database if exists ash_install_failure_test with (force);
+          create database ash_install_failure_test;
+          drop database if exists ash_migration_failure_test with (force);
+          create database ash_migration_failure_test;
+          EOF
+          "${psql_base[@]}" \
+            --dbname=ash_install_failure_test \
+            --set=ON_ERROR_STOP=1 << 'EOF'
+          create schema ash;
+          create view ash.config as select true as singleton;
+          EOF
+
+          set +e
+          "${psql_base[@]}" \
+            --dbname=ash_install_failure_test \
+            --file=sql/ash-install.sql \
+            >/tmp/ash_install_failure.log 2>&1
+          install_rc=$?
+          set -e
+          if (( install_rc == 0 )); then
+            cat /tmp/ash_install_failure.log >&2
+            echo "fresh installer masked an expected SQL failure" >&2
+            exit 1
+          fi
+
+          "${psql_base[@]}" \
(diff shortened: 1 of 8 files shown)


NikolayS:
Superseded by #184, which implements the rescoped atomic installer contract, retains the discriminating rollback coverage, and is now merged on main as 8eff377.

Issue #128, opened 2026-07-10
Bug: raw retention start includes the excluded ring slot


NikolayS:
## Summary

`ash._raw_retention_start()` computes `min(sample_ts)` across every physical ring partition, including the oldest slot that raw readers deliberately exclude.

The usable raw slot set is current plus `num_partitions - 2` completed slots. The remaining physical slot is retained only until its next truncate and is not queryable through `_active_slots_for*()`.

Verified on `origin/main` at `5aac2e3` with Postgres 18.3.

## Reproduction

With the default three slots and `current_slot = 0`:

- readable slots are `{0,2}`;
- slot 1 is excluded;
- seed a row from two days ago in slot 1;
- seed a row from ten minutes ago in slot 0.

Observed:

- `ash._raw_retention_start()` reports the two-day-old excluded row;
- `ash.status()` and report coverage expose that false boundary;
- a tie drill for the old event/query passes `_raise_tie_retention()`, but `_active_slots_for_at()` returns no slots, so the reader silently returns `source = raw` with zero data.

## Expected

The raw retention start must be the earliest row in the same retained slot set used by raw readers. The old tie drill should then raise the documented outside-retention error.

## Root cause

`_raw_retention_start()` scans all of `ash.sample`. The zero-argument `_active_slots()` helper still returns only current plus previous instead of enumerating the configured retained set, so it cannot safely be reused for the boundary.

## Proposed fix and regression

- Make `_active_slots()` enumerate offsets `0 .. num_partitions - 2`, matching `_active_slots_for*()`.
- Filter `_raw_retention_start()` with that slot set.
- Seed the excluded ring slot in CI and assert the helper/status boundary ignores it and an old wait/query tie drill raises.

PR #129, opened 2026-07-10
fix: ignore excluded raw ring slot

NikolayS:
## What changed

- makes `ash._active_slots()` enumerate the configured retained ring offsets `0 .. num_partitions - 2`
- filters `ash._raw_retention_start()` to that exact readable slot set
- adds a behavioral regression with a populated excluded oldest slot
- adds a discriminating six-partition assertion for the generalized slot enumeration

## Why

The physical oldest ring slot remains populated until its next truncate but is deliberately excluded from raw readers. Including it in the retention minimum made `status()` and report coverage advertise unreadable history and allowed old event/query tie drills to pass the guard before returning a silent zero.

## Impact

Raw-retention metadata now describes data the reader helpers can actually access. An event/query tie drill over excluded history raises the documented outside-retention error.

This PR intentionally does not fix the time-bound path from #128. It also moves the boundary newer, making the partial-first-minute defect in #163 strictly worse; #163 must be fixed with or immediately after this change. Frozen 1.x installers still contain the old two-slot `_active_slots()` body and must not be reapplied as a terminal installer over 2.0.

## Validation

Postgres 17.9 after rebasing onto `origin/main` at `ce541ac`:

- RED: the new N=6/current=2 assertion exits 3 on the base with `got {2,1}`
- GREEN at `2006159`: `_active_slots()` returns exact `{2,1,0,5,4}` and the complete named retention regression exits 0
- existing N=3 fixture still verifies `{0,2}`, helper/status agreement, and the outside-retention error
- actionlint and `git diff --check` pass
- `codex review --base origin/main` reports no introduced correctness issue

Refs #128

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -4693,6 +4693,30 @@ jobs:
           PGPASSWORD: postgres
         run: |
           psql -h localhost -U postgres -d postgres -v ON_ERROR_STOP=1 << 'EOF'
+          begin;
+
+          update ash.config
+          set num_partitions = 6,
+              current_slot = 2
+          where singleton;
+
+          do $$
+          declare
+            v_slots smallint[];
+          begin
+            select ash._active_slots() into v_slots;
+
+            assert v_slots = array[2, 1, 0, 5, 4]::smallint[],
+              format(
+                '_active_slots() for N=6/current=2: '
+                'expected {2,1,0,5,4}, got %s',
+                v_slots
+              );
+          end
+          $$;
+
+          rollback;
+
           do $$
           declare
             v_wait smallint;
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -4688,6 +4688,70 @@ jobs:
           end $$;
           EOF
 
+      - name: "Test 2.0: raw retention ignores the excluded ring slot"
+        env:
+          PGPASSWORD: postgres
+        run: |
+          psql -h localhost -U postgres -d postgres -v ON_ERROR_STOP=1 << 'EOF'
+          do $$
+          declare
+            v_wait smallint;
+            v_old_ts int4 := ash.ts_from_timestamptz(
+              date_trunc('minute', now()) - interval '2 days');
+            v_recent_ts int4 := ash.ts_from_timestamptz(
+              date_trunc('minute', now()) - interval '10 minutes');
+            v_start timestamptz;
+            v_status_start timestamptz;
+            v_slots smallint[];
+            v_raised boolean := false;
+          begin
+            truncate ash.sample_0, ash.sample_1, ash.sample_2;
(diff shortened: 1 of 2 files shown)


NikolayS:
> **Ready, with a caveat** — Both halves RED on base. `_active_slots()` rewrite is required, not gratuitous (base returns `{2,1}` for N=6 while both `_for` variants return 5 slots). No perf regression (InitPlan Limit over MergeAppend). **Do not close #128** — the *time*-bound path to the same fabricated zero is untouched. **Fragility:** frozen 1.x scripts re-create `_active_slots()` with the old two-slot body, now load-bearing. **And see B3 — this makes B3 worse.**

Rebased by hand onto `origin/main` at `ce541ac` with no conflict.

RED on the rebased base: `Test 2.0: raw retention ignores the excluded ring slot` exits 3 at the new discriminator with `_active_slots() for N=6/current=2: expected {2,1,0,5,4}, got {2,1}`.

GREEN on `2006159`: the exact same complete step exits 0; the N=6 helper returns `{2,1,0,5,4}`, then the existing excluded-slot helper/status/error-path assertions pass.

`codex review --base origin/main` is standing in for REV and found no introduced correctness issue: retained-slot enumeration matches rotation semantics, and the retention boundary uses the same readable slot set. Fresh-install SQL checks and workflow syntax validation passed.

Caveats accepted explicitly: this does **not** fix #128's time-bound path, so #128 stays open. Moving the boundary newer makes #163 strictly worse; #163 must be fixed with or immediately after this PR. Frozen 1.x installers still carry the old two-slot helper body and are now load-bearing migration fragility.

NikolayS:
**Heads-up — merging this will require a one-line test change once #219 lands.**

#219 (fixes #212) redefines `ash.status().raw_retention_start` from *the exact oldest retained sample* to *the configured ring boundary* (`rotated_at - (num_partitions - 2) * rotation_period`, floored to the minute). That split is the fix for #212: the field is documented as a window-planning boundary, but returning `min(sample_ts)` made the README's flagship drill hard-error on any young install.

This PR's assertion that `raw_retention_start` equals the exact oldest retained sample is correct today and becomes wrong afterwards.

Composition was tested against this PR's head `2006159` rather than assumed. The **code** cherry-picks with no textual conflict; only that test expectation needs updating, to compute the boundary from `rotated_at`, `num_partitions` and `rotation_period`. With exactly that change, both suites pass together:

```
NOTICE: raw retention excludes the oldest ring slot PASSED
NOTICE: issue #212 raw retention boundary tests PASSED
NOTICE: retention error split PASSED (boundary hint is followable, echoes passed start)
```

Recommended order: **this PR first, then #219.** Note also that this PR moves the boundary *newer*, which makes #212 strictly worse in the interim — so they should land close together.

Context: #160.

PR #133, opened 2026-07-10
fix: defer hourly rollup until minute coverage

### LINKED ISSUE #132: Bug: rollup_hour can permanently seal an incomplete hour
## Summary

At the top of an hour, `rollup_minute()` and `rollup_hour()` compete for the same rollup advisory lock. If the hourly worker wins before the just-completed final minute is rolled up, it materializes an incomplete hourly row and advances `last_rollup_1h_ts`. The late minute is then never revisited.

This directly corrupts long-window 2.0 AAS readers.

Verified on `origin/main` at `5aac2e3` with Postgres 18.3.

## Reproduction

1. Seed minute rollups 0 through 58 for a completed hour.
2. Set `last_rollup_1m_ts = hour + 3540` and `last_rollup_1h_ts = hour`.
3. Call `ash.rollup_hour()`.
4. Insert minute 59 and advance `last_rollup_1m_ts = hour + 3600`.
5. Call `ash.rollup_hour()` again.

Observed before the fix:

```text
first call:
samples = 3540
wait_counts = {cpu_id,3540}
minute_counts[60] = NULL

second call:
rollup_hour() = 0
row remains incomplete
```

## Expected

The hourly worker must not insert or advance past an hour until the minute-rollup watermark reaches that hour's end. Once minute 59 is available, a later call should aggregate the exact complete hour and advance normally.

## Proposed fix

Read both watermarks under the existing shared rollup lock. Before each hourly upsert, require `last_rollup_1m_ts >= hour_end`; otherwise stop without changing the hourly row or watermark.

The next scheduled hourly call can safely catch up.

NikolayS:
## What changed

- reads `last_rollup_1m_ts` alongside the hourly watermark and refuses to seal an incomplete hour
- makes `rollup_hour()` wait behind an in-flight minute rollup instead of losing the shared-lock race
- schedules hourly rollup at minute 1 for pg_cron and documented external schedulers
- migrates an existing minute-0 pg_cron job in place during installer reapply, using the ordinary job-owner-capable named `cron.schedule()` path
- adds deterministic barrier, lock-contention, schedule, and installer-reapply regressions

## Why

If the hourly worker ran before minute 59 was rolled up, it could permanently seal a 59-minute row. A try-lock at minute 1 could also starve hourly aggregation, and merely redefining `ash.start()` did not migrate existing jobs.

## Impact

Incomplete hours are deferred without inserting or advancing. The hourly worker waits for an in-flight minute worker, and existing/default schedulers use minute 1. Installer reapply preserves the legacy job ID, command, and active state while changing only its exact old default schedule.

## Validation

- RED on rebased `origin/main`: incomplete-hour regression exits 3 after `rollup_hour()` returns 1 and seals the partial hour
- GREEN: incomplete hour returns 0; no row or watermark advance; retry after minute 59 creates exactly 3,600 samples/counts and 60 minute slots
- lock contention RED before serialization: `rollup_hour_result=0`, `hourly_rows=0`; GREEN after fix: result 1 and one hourly row after the minute worker commits
- schedule RED/GREEN: legacy/default `0 * * * *` becomes `1 * * * *`
- ordinary non-superuser pg_cron job owner, without `cron.alter_job` EXECUTE: installer reapply succeeds and migrates the job
- fresh PG17 focused tests and disposable pg_cron PG17 tests pass; workflow YAML and `git diff --check` pass

Closes #132

## Audit side effects

- The former minute-0 collision could leave `rollup_1h` about one hour behind. The branch now uses minute 1 in pg_cron, installer migration, no-pg_cron guidance, and README examples; `rollup_hour()` also blocks behind any in-flight minute rollup.
- `ash.rollup_hour()` remains a silent no-op when `last_rollup_1m_ts` is NULL even if `ash.rollup_1m` already contains rows. This is an explicit operational side effect of using the minute watermark as the completion boundary.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -1424,6 +1424,49 @@ jobs:
           PGPASSWORD: postgres
         run: |
           psql -h localhost -U postgres -d postgres -v ON_ERROR_STOP=1 << 'EOF'
+          select case
+                   when exists (
+                     select from pg_extension where extname = 'pg_cron'
+                   ) then 'true'
+                   else 'false'
+                 end as has_pgcron \gset
+          \if :has_pgcron
+            -- Simulate an existing installation whose job still has the old
+            -- unsafe schedule, then prove installer re-apply migrates it.
+            create temp table _legacy_hourly_job (jobid bigint primary key);
+            insert into _legacy_hourly_job
+            select cron.schedule(
+              'ash_rollup_1h', '0 * * * *', 'select ash.rollup_hour()'
+            );
+            select cron.alter_job(
+              job_id := (select jobid from _legacy_hourly_job),
+              active := false
+            );
+            \i sql/ash-install.sql
+            do $migration$
+            declare
+              v_job_id bigint;
+              v_schedule text;
+              v_command text;
+              v_active bool;
+            begin
+              select jobid, schedule, command, active
+              into v_job_id, v_schedule, v_command, v_active
+              from cron.job
+              where jobname = 'ash_rollup_1h';
+              assert v_job_id = (select jobid from _legacy_hourly_job),
+                'installer re-apply replaced the existing hourly job';
+              assert v_schedule = '1 * * * *',
+                'installer re-apply must migrate hourly rollup to minute 1, '
+                'got: ' || v_schedule;
+              assert v_command = 'select ash.rollup_hour()',
+                'installer re-apply changed the hourly command';
+              assert not v_active,
+                'installer re-apply reactivated the disabled hourly job';
+            end
+            $migration$;
+          \endif
+
           do $$
           declare
             v_job_id bigint;
@@ -3506,6 +3549,8 @@ jobs:
         env:
           PGPASSWORD: postgres
         run: |
+          set -euo pipefail
+
           psql -h localhost -U postgres -d postgres -v ON_ERROR_STOP=1 << 'EOF'
           do $$
           declare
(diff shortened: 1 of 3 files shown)


NikolayS:
## Pre-merge verification (standing in for REV)

> **Ready** — Minimal, correct. Note two side effects in the PR: with the shipped cron (`* * * * *` + `0 * * * *` both firing at minute 0) `rollup_1h` will persistently lag ~1 hour — schedule the hourly job at minute 1; and `rollup_hour()` becomes a silent no-op if `last_rollup_1m_ts` is NULL while `rollup_1m` has rows.

Codex review findings and resolutions:

- P1: minute-1 still collided with the every-minute try-lock. Fixed by making `rollup_hour()` block behind the in-flight minute worker; deterministic contention is RED `result=0/rows=0`, GREEN `result=1/rows=1`.
- P1: existing minute-0 jobs were not migrated by installer reapply. Fixed with an exact-old-default migration that preserves job ID, command, and active state.
- P1: `cron.alter_job` is unavailable to ordinary job owners. Replaced it with named `cron.schedule()`, verified with a non-superuser owner lacking `alter_job` EXECUTE.
- P2: the external-scheduler example still used minute 0. README and no-pg_cron guidance now use minute 1.

RED/GREEN:

- Barrier RED on rebased main: the exact regression exits 3 after sealing the 59-minute hour. GREEN: defers with no row/watermark change, then produces exactly 3,600 samples/counts and 60 slots after minute 59.
- Schedule RED/GREEN: `0 * * * *` → `1 * * * *`.
- Focused PG17 barrier/lock tests, disposable pg_cron PG17 schedule/reapply tests, ordinary-owner reapply, YAML parse, and `git diff --check` all pass.

Additional pre-merge edits: minute-1 pg_cron schedule, blocking hourly lock, upgrade migration, external-scheduler docs, and their focused regressions. The NULL-watermark/no-op caveat remains explicitly documented in the PR description.

Issue #91, opened 2026-05-28
Bug: small reader/input-validation edge cases


r314tive:
## Summary

Batching several small but reproducible edge-case bugs into one issue so they do not pollute the tracker as separate one-line reports.

## 1. `top_waits(..., p_limit => 10)` can return 11 rows

Repro:

```sql
select ash._register_wait('active', 'WaitType', 'WaitEvent' || i)
from generate_series(1, 12) as gs(i);

insert into ash.sample (sample_ts, datid, active_count, data, slot)
values (
  ash.ts_from_timestamptz(now()),
  0::oid,
  12::smallint,
  array[
    -1,1,0,-2,1,0,-3,1,0,-4,1,0,
    -5,1,0,-6,1,0,-7,1,0,-8,1,0,
    -9,1,0,-10,1,0,-11,1,0,-12,1,0
  ]::int4[],
  0::smallint
);

select count(*) as top_waits_rows
from ash.top_waits('1 hour', 10, 10);
```

Observed:

```text
top_waits_rows = 11
```

The result includes 10 individual wait rows plus an `Other` row. If `p_limit=10` means a hard row limit, this should be 9 rows plus `Other`, or the docs/signature should clearly say it can return `p_limit + 1`.

## 2. Zero timeline bucket crashes with division by zero

Repro:

```sql
select * from ash.wait_timeline('1 hour', '0 seconds');
select * from ash.timeline_chart('1 hour', '0 seconds', 4, 20);
```

Observed:

```text
ERROR: division by zero
CONTEXT: SQL function "wait_timeline" statement 1

ERROR: division by zero
CONTEXT: PL/pgSQL function timeline_chart(interval,interval,integer,integer,boolean) line 29 at SQL statement
```

Expected: explicit validation error, e.g. bucket must be at least 1 second and a whole number of seconds.

## 3. `ash.start(NULL::interval)` fails through a raw NOT NULL constraint

Repro:

```sql
select * from ash.start(null::interval);
```

Observed:

```text
ERROR: null value in column "sample_interval" of relation "config" violates not-null constraint
CONTEXT: SQL statement "update ash.config
    set sample_interval = p_interval,
      sampling_enabled = true,
      skipped_samples = 0
    where singleton"
PL/pgSQL function start(interval) line 100 at SQL statement
```

Expected: either treat NULL as default interval or reject it before touching `ash.config` with a clear function-level validation error.

[...]

PR #142, opened 2026-07-10
fix: validate ash.start intervals exactly

### LINKED ISSUE #141: Bug: ash.start rounds unsupported intervals before validation
## Summary

`ash.start(every interval)` casts `extract(epoch from every)` to `int` before validating the interval. PostgreSQL rounds the value during that cast, so fractional-second cadences are accepted even though pg_cron is scheduled at a whole-second cadence. Very large intervals raise an internal `integer out of range` error instead of returning the documented validation row.

Verified on `origin/main` at `5aac2e3`, using Postgres 18 without pg_cron. The validation is intentionally shared by pg_cron and degraded mode, so the same cast occurs in both paths.

## Reproduction

```sql
select job_type, status from ash.start(interval '0.6 seconds');
select sample_interval from ash.config where singleton;

select job_type, status from ash.start(interval '59.6 seconds');
select sample_interval from ash.config where singleton;

select * from ash.start(interval '100000000 days');
```

Observed:

```text
sampler | interval set to 00:00:00.6 ...
sample_interval = 00:00:00.6

sampler | interval set to 00:00:59.6 ...
sample_interval = 00:00:59.6

ERROR: integer out of range
CONTEXT: PL/pgSQL function start(interval) ... at assignment
```

With pg_cron available, the first two inputs become schedules such as `1 seconds` and `60 seconds`, while `ash.config.sample_interval` retains the fractional value. The 2.0 AAS readers multiply sample counts by the configured interval, so that mismatch can mis-scale historical load.

## Expected

- Reject cadences that are not an exact whole number of seconds before any integer cast.
- Reject intervals below 1 second and above 23 hours with a normal `job_type = 'error'` result.
- Preserve the existing exact-second, exact-minute, and exact-hour schedule rules.
- Exercise the same behavior with and without pg_cron.

NikolayS:
Closes #141

## Summary

- validate the exact numeric epoch before converting an interval to `int`
- reject fractional seconds before scheduling can diverge from `ash.config.sample_interval`
- reject intervals above 23 hours before an integer-overflow failure
- preserve the accepted shapes: 1–59 whole seconds, whole minutes, and whole hours through 23h
- document the public cadence contract and add a matrix regression gate
- credit the #141 fix to PR #142 in the release notes

## Validation

Postgres 17.9 degraded/no-pg_cron path after rebasing onto `origin/main` at `6426500`:

- RED: the named gate exits 3 because base accepts `0.6 seconds` and returns zero error rows
- GREEN at `f7e64c4`: 0.6s, 1.4s, 59.6s, and 100000000 days all return validation rows and leave `sample_interval` unchanged
- valid 1s, 60s, and 23h shapes remain accepted; 90s remains rejected by the exact-minutes rule
- actionlint and `git diff --check` pass
- `codex review --base origin/main` reports no actionable issue

The accepted message drift is explicit: a value such as 23.5 hours now reports that it exceeds the 23-hour maximum before reaching the exact-hours shape check.


NikolayS:
> **Ready** — RED/GREEN verified on the degraded path; 20-case matrix diffed vs main. Fix `RELEASE_NOTES.md` first — the bullet credits **PR #99** for the #141 fix instead of #142. Message drift: `'23.5 hours'` now says "exceeds maximum 23 hours" rather than "must be exact hours."

Rebased by hand onto `origin/main` at `6426500` with no conflict.

RED on the rebased base: `Release-gate: ash.start rejects fractional and oversized intervals` exits 3 because `ash.start(interval '0.6 seconds')` returns zero error rows and proceeds down the no-pg_cron scheduling path.

GREEN on `f7e64c4`: the same complete step exits 0; fractional and huge intervals return validation rows and rejected calls leave `sample_interval` unchanged. The audit's valid-shape matrix was also rechecked during review.

Pre-merge edit applied: release notes now attribute issue #141 to PR #142 while retaining PR #99's separate issue #91 credit. The 23.5-hour message-order drift is consciously accepted and documented in the PR body.

`codex review --base origin/main` is standing in for REV and found no actionable issue: exact numeric validation precedes the integer cast, valid schedule shapes remain intact, and rejected inputs do not mutate configuration.

PR #150, opened 2026-07-10
fix: keep compacted query attribution honest

### LINKED ISSUE #136: Bug: compacted query rollups are presented as exact AAS
## Summary

The rollup pipeline intentionally compacts query attribution, but the 2.0 `ash.top('query_id')` surface presents the retained subset as an exact total.

- `rollup_minute()` drops query IDs below `rollup_min_backend_seconds`.
- `rollup_hour()` keeps only the top 100 query IDs.
- `_grain_by(..., 'query_id')` emits only retained IDs.
- `ash.top()` computes its denominator from those retained keys.

This can make a small retained query appear to represent 100% of load while most backend time disappears.

Verified on `origin/main` at `5aac2e3` with Postgres 18.3 through the real `rollup_minute()` path.

## Reproduction

One raw minute contains:

```text
total backend appearances = 26
query 111 = 3
query 222 = 2
query 333 = 1
no query_id = 20
```

With the default minimum of 3, the produced `query_counts` is only `{111,3}`. After raw data is removed:

```text
ash.aas:
  avg_aas = 0.43
  backend_seconds = 26

ash.top('query_id'):
  key = 111
  avg_aas = 0.05
  backend_seconds = 3
  pct = 100.00
```

The correct retained-query share is `3 / 26 = 11.54%`, and 23 seconds require an explicit unattributed/not-preserved residual.

## Expected

For rollup-backed query breakdowns:

- compute the exact per-grain residual `sum(wait_counts) - sum(query_counts)`;
- expose that residual through the NULL/unattributed bucket;
- use the full backend total as the percentage denominator; and
- document that named query IDs in rollups are compacted top attribution.

A specific `query_id` filter cannot distinguish true zero from compacted-away attribution. It should use raw data and raise outside raw retention, or otherwise expose an explicit unavailable state.

## Regression requirement

Generate the fixture through `rollup_minute()`, not by hand. After removing raw data, assert:

- sum of `top('query_id').backend_seconds` equals `ash.aas().backend_seconds`;
- query 111 is 3 seconds and 11.54%;
- the NULL residual is 23 seconds; and
- a compacted-away explicit query filter is not reported as an exact rollup zero.

NikolayS:
Closes #136

## Summary

- keep preserved named query IDs useful on rollup-backed `top('query_id')` reads
- add a NULL residual for uncaptured and compacted-away attribution so `backend_seconds` and `pct` reconcile to total AAS
- force raw samples for every explicit `query_id` filter in `aas`, `timeline`, `top`, and inherited `compare` paths
- continue forcing raw for query breakdowns combined with wait filters
- document minute-threshold/hourly-top-set compaction and exact-vs-approximate query semantics throughout the 2.0 API docs

## TDD evidence

A real packed sample contains 26 backend appearances: q111=3, q222=2, q333=1, and NULL=20. With `rollup_min_backend_seconds = 3`, the stored minute `query_counts` is only `[111,3]`.

Before the fix, rollup-backed `top('query_id')` reported q111 as 3 seconds and 100% of query load. After the fix it reports:

- q111: 3 seconds, 11.54%
- NULL residual: 23 seconds, 88.46%
- total: 26 seconds, 100%

An independent hourly-rollup fixture produced the same reconciliation after hourly top-set compaction.

## Validation

Postgres 18:

- fresh install
- dedicated #136 regression
- full 2.0 behavioral reader suite
- existing rollup-source reader suite
- independent `rollup_1h` residual and explicit-filter check
- full historical upgrade chain followed by the #136 regression
- documentation lint
- `git diff --check`

The hosted PG14–PG19 beta and no-cron matrix is left to this draft PR.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -3712,6 +3712,159 @@ jobs:
           end $$;
           EOF
 
+      - name: "Issue #136: compacted query attribution stays honest"
+        env:
+          PGPASSWORD: postgres
+        run: |
+          psql -h localhost -U postgres -d postgres -v ON_ERROR_STOP=1 << 'EOF'
+          do $$
+          declare
+            v_ts int4 := (
+              ash.ts_from_timestamptz(
+                date_trunc('minute', now() - interval '20 minutes')
+              ) / 60
+            ) * 60;
+            v_from timestamptz;
+            v_to timestamptz;
+            v_datid oid;
+            v_wait_id smallint;
+            v_q1 int4;
+            v_q2 int4;
+            v_q3 int4;
+            v_result int;
+            v_aas record;
+            v_named record;
+            v_residual record;
+            v_rows int;
+            v_total numeric;
+            v_pct numeric;
+          begin
+            truncate ash.sample_0, ash.sample_1, ash.sample_2;
+            truncate ash.query_map_0, ash.query_map_1, ash.query_map_2
+              restart identity;
+            truncate ash.rollup_1m, ash.rollup_1h;
+            update ash.config
+            set current_slot = 0,
+                sample_interval = interval '1 second',
+                rollup_min_backend_seconds = 3,
+                last_rollup_1m_ts = null,
+                last_rollup_1h_ts = null
+            where singleton;
+
+            select oid into v_datid
+            from pg_database where datname = current_database();
+            select ash._register_wait('active', 'CPU*', 'CPU*')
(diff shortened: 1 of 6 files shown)


NikolayS:
Superseded by #198, which recovered the #136 compacted-query-attribution rework on current main and landed it as 6d7258335e64dca65602e3a501e545f6d8f4ffe3. Closing this older attempt in favor of that successor.

PR #181, opened 2026-07-27
release: fix reader privilege blockers and sample slot race

### LINKED ISSUE #164: Release blocker: rebuild_partitions strips all reader grants
## Summary

`ash.rebuild_partitions()` drops and recreates the query-map view and every raw partition without restoring reader ACLs. This removes custom `grant_reader()` access and the `pg_monitor` grant installed by default. The complete 2.0 reader surface then hard-errors once new samples arrive.

Audit finding: B4 in #160.

## Location

- Drops `ash.query_map_all`: `sql/ash-install.sql:1575`.
- Creates replacement sample/query-map objects: `sql/ash-install.sql:1588-1610`.
- Rebuilds `query_map_all`: `sql/ash-install.sql:1613`.
- Existing installer role snapshot: `sql/ash-install.sql:14-94`.
- Existing installer replay through `grant_reader()`: `sql/ash-install.sql:6283-6360`.

## Reproduction

Verify a configured reader role can use the reader API, then rebuild:

```sql
select ash.rebuild_partitions(6, 'yes');
```

After sampling resumes, calls including these fail:

```sql
select * from ash.aas();
select * from ash.summary();
select * from ash.samples();
select * from ash.top('wait_event');
select * from ash.chart();
select * from ash.timeline();
select * from ash.status();
```

Observed:

```text
ERROR: permission denied for view query_map_all
```

The same failure occurs for a plain `pg_monitor` member that never called `grant_reader()`. ACL inspection shows `relacl = NULL` on `query_map_all`, `query_map_0`, and `sample_0`, while the retained parent `sample` and rollup tables keep their ACLs.

## Observed vs expected

Observed: a documented admin operation destroys every configured reader's access. Immediately after rebuild the empty raw storage lets readers appear healthy; the outage begins only after the first new samples arrive.

Expected: rebuilding storage must preserve the complete existing reader-role bundle, including the default `pg_monitor` grant, across every dropped/recreated table and view.

## Impact

Every configured dashboard can switch to hard errors and blank panels. The delayed onset defeats the obvious post-rebuild smoke test. `README.md:376-378` tells operators to re-grant monitoring roles because new partitions need grants, but the first failure is the recreated view, and `README.md:403` does not warn that the installer-created `pg_monitor` grant is also destroyed.

## Suggested fix

Use the existing #107 machinery around `rebuild_partitions()`: snapshot reader-role bundles before the drop/recreate operation and replay `ash.grant_reader()` for each role after all partitions and `query_map_all` have been rebuilt.

## Interaction

No open issue or in-flight PR addresses this rebuild path. The installer already solves the same class of drop/recreate ACL loss at `sql/ash-install.sql:14-94` and `sql/ash-install.sql:6283-6360`; reuse that pattern rather than creating a second privilege model.

### LINKED ISSUE #166: Release blocker: revoke_reader can brick a non-superuser-owned installation
## Summary

`ash.revoke_reader()` accepts the non-superuser role that owns schema `ash`. It then revokes schema USAGE and reader/helper EXECUTE privileges from that owner. Sampling, every reader, and `ash.grant_reader()` itself stop working, while the destructive call returns cleanly with no warning.

This affects the managed-Postgres deployment model pg_ash targets; a superuser owner is immune through ACL bypass.

Audit finding: B6 in #160.

## Location

- `ash.revoke_reader()` body: `sql/ash-install.sql:6195-6263`.
- Existing validation checks only empty/nonexistent roles: `sql/ash-install.sql:6208-6215`.
- Non-admin function EXECUTE revocation: `sql/ash-install.sql:6219-6231`.
- Schema USAGE revocation: `sql/ash-install.sql:6256`.
- `_admin_funcs()` definition: `sql/ash-install.sql:5930-5958`; it does not include `_admin_funcs` itself.

## Reproduction

As a non-superuser schema owner, run:

```sql
select ash.revoke_reader(current_user);
```

The call completes. Afterward, these and the rest of the reader surface fail:

```sql
select ash.take_sample();
select * from ash.status();
select ash.grant_reader(current_user);
```

Observed:

```text
ERROR: permission denied for schema ash
```

Manually restoring only schema USAGE does not repair the documented inverse:

```text
ERROR: permission denied for function _admin_funcs
```

`revoke_reader()` treats `_admin_funcs()` as a non-admin reader helper and strips EXECUTE on the function that `grant_reader()` initializes from.

## Observed vs expected

Observed: a valid-looking helper call disables sampling and the entire reader/admin repair path for the installation owner without warning.

Expected: the helper must reject a target that owns schema `ash`, is `current_user`, or is a superuser before revoking anything. The documented inverse must remain callable after every permitted revocation.

## Impact

Sampling and every reader fail for the database. Both documented repair paths are broken by the same call. Recovery requires hand-written grants across all functions/tables or a full installer re-apply.

## Suggested fix

Add an early guard that refuses schema-owner, current-user, and superuser targets before the revocation loop.

## Interaction

No in-flight PR or existing open issue addresses this owner-target path. B4 concerns ACL loss during partition rebuild, and the `pg_monitor` opt-out finding concerns installer re-apply; neither duplicates this installation-bricking call.

### LINKED ISSUE #169: Bug: take_sample can stamp query-map IDs with the wrong slot during rotation
## Summary

`ash.take_sample()` captures `v_current_slot` for query-map ID resolution, but its final sample INSERT omits the `slot` column. The table default calls `ash.current_slot()` again under the INSERT statement's later snapshot. If rotation lands between those points, packed IDs from the old dictionary are stamped with the new slot.

Audit finding: N2 in #160. Recommendation: fix now.

## Location

- Captured slot: `sql/ash-install.sql:836`.
- INSERT that omits `slot`: `sql/ash-install.sql:1017-1018`.
- Table default: `sql/ash-install.sql:537`.
- Sampler lock ID: `sql/ash-install.sql:823-825`.
- Rotation lock ID: `sql/ash-install.sql:1340-1342`.

## Reproduction

Allow rotation to occur after `take_sample()` reads `v_current_slot` and resolves query-map IDs, but before its final INSERT. Sampling and rotation use different advisory-lock object IDs, so they do not serialize this transition.

## Observed vs expected

Observed: packed map IDs from the old slot are stored with the new slot. Attribution is initially NULL, then can resolve to unrelated query IDs when the new slot's identity sequence reissues the same integers. The row passes `_sample_data_is_valid()`; no warning is emitted and `insert_errors` is not incremented.

Expected: the sample row must carry the same captured slot used for every dictionary lookup in that sampling tick.

## Impact

The blast radius is one tick per collision and grows with shorter rotation periods. The resulting attribution corruption is not detectable after the fact.

## Suggested fix

Add `slot` to the INSERT column list and `v_current_slot` to its values so this path never re-evaluates the table default.

## Interaction

This is not #128, which concerns retained/readable slot selection, or #143, which concerns malformed packed arrays. The sub-day rotation decision in #167 affects collision frequency but not the correctness or necessity of this fix.

NikolayS:
Closes #164.
Closes #166.
Closes #169.

## What changed

- Preserve complete direct reader bundles, including the installer-default `pg_monitor` bundle, across `ash.rebuild_partitions()` by reusing the installer's `_ash_install_reader_roles` snapshot/replay mechanism.
- Refuse `ash.revoke_reader()` when its target owns schema `ash`, is `current_user`, or is a superuser. Keep `_admin_funcs()` out of reader grants and scrub its legacy EXECUTE grant during installer re-apply.
- Stamp each `ash.sample` row with the already captured `v_current_slot`, so packed query-map IDs and their slot cannot diverge during rotation.
- Correct README/RELEASE_NOTES guidance around partition rebuilds and reader ACL preservation.

## Regression coverage

- #164: exact direct ACL checks before/after `rebuild_partitions(6, 'yes')`; immediate empty-reader smoke test; delayed sample insertion; custom reader plus a plain `pg_monitor` member.
- #166: isolated database owned by a LOGIN/NOSUPERUSER role; exact refusal message and no privilege loss; independent owner/current-user/superuser branches; fresh reader and legacy `_admin_funcs()` ACL re-apply coverage.
- #169: forced two-value `ash.current_slot()` transition proves the sample uses the captured slot and reads it once; exact query attribution and `insert_errors = 0`; normalized function-definition assertion proves there is exactly one `ash.sample` INSERT path and that it names `slot` explicitly.

## Local validation

- PostgreSQL 17.9 fresh install, installer re-apply, and full upgrade chain
- Verbatim RED reproduction and GREEN regression runs for #164, #166, and #169
- Workflow YAML parse, targeted embedded-shell syntax, and `git diff --check`

The N2 behavioral probe deterministically forces the slot transition. Its tagged one-active-backend fixture is controlled in CI, but—as with any `pg_stat_activity` fixture—could be perturbed by an unexpected concurrent client in a shared local database; the structural assertion remains fully deterministic.

This draft was rebased onto current `origin/main` immediately before push. It still needs a fresh rebase onto `origin/main` before merge.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -954,6 +954,432 @@ jobs:
           drop role ash_pgmon_outsider;
           EOF
 
+      - name: "Issue #164: rebuild_partitions preserves reader ACLs"
+        env:
+          PGPASSWORD: postgres
+        run: |
+          psql -h localhost -U postgres -d postgres -v ON_ERROR_STOP=1 << 'EOF'
+          do $$
+          begin
+            if not exists (
+              select from pg_roles where rolname = 'ash_b4_reader'
+            ) then
+              create role ash_b4_reader login password 'x';
+            end if;
+            if not exists (
+              select from pg_roles where rolname = 'ash_b4_pgmon_member'
+            ) then
+              create role ash_b4_pgmon_member login password 'x';
+            end if;
+          end $$;
+
+          grant pg_monitor to ash_b4_pgmon_member;
+          select ash.grant_reader('ash_b4_reader');
+
+          -- Seed one exact query-map-backed row before the rebuild. This proves
+          -- both a configured reader and a plain pg_monitor member can traverse
+          -- query_map_all; the member itself has no direct ash ACL.
+          do $$
+          declare
+            v_wait_id smallint;
+            v_map_id int4;
+            v_datid oid;
+          begin
+            truncate ash.sample, ash.query_map_0, ash.query_map_1, ash.query_map_2
+              restart identity;
+            update ash.config
+            set current_slot = 0,
+              sampling_enabled = true
+            where singleton;
+
+            select ash._register_wait('active', 'B4Test', 'B4Wait')
+              into v_wait_id;
+            insert into ash.query_map_0 (query_id)
+            values (164000000001)
+            returning id into v_map_id;
+            select oid into v_datid
+            from pg_database
+            where datname = current_database();
+            insert into ash.sample (
+              sample_ts,
+              datid,
+              active_count,
+              data,
+              slot
+            )
+            values (
+              ash.ts_from_timestamptz(
+                date_trunc('minute', now()) - interval '2 minutes'
+              ),
+              v_datid,
+              1,
+              array[-v_wait_id, 1, v_map_id]::integer[],
+              0
+            );
+          end $$;
+
+          select
+            'before rebuild' as phase,
+            rel.relname,
+            rel.relacl,
+            aclcontains(
+              rel.relacl,
+              makeaclitem(reader_role.oid, rel.relowner, 'SELECT', false)
+            ) as custom_direct,
+            aclcontains(
+              rel.relacl,
+              makeaclitem(pg_monitor.oid, rel.relowner, 'SELECT', false)
+            ) as pg_monitor_direct,
+            coalesce(
+              aclcontains(
+                rel.relacl,
+                makeaclitem(member_role.oid, rel.relowner, 'SELECT', false)
+              ),
+              false
+            ) as member_direct
+          from pg_class as rel
+          join pg_namespace as nsp on nsp.oid = rel.relnamespace
+          cross join pg_roles as reader_role
+          cross join pg_roles as pg_monitor
+          cross join pg_roles as member_role
+          where nsp.nspname = 'ash'
+            and rel.relname in ('query_map_all', 'query_map_0', 'sample_0')
+            and reader_role.rolname = 'ash_b4_reader'
+            and pg_monitor.rolname = 'pg_monitor'
+            and member_role.rolname = 'ash_b4_pgmon_member'
+          order by rel.relname;
+
+          do $$
+          declare
+            v_direct_acl_count int;
+            v_member_acl_count int;
+          begin
+            select count(*) into v_direct_acl_count
+            from pg_class as rel
+            join pg_namespace as nsp on nsp.oid = rel.relnamespace
+            cross join pg_roles as grantee_role
+            where nsp.nspname = 'ash'
+              and rel.relname in ('query_map_all', 'query_map_0', 'sample_0')
+              and grantee_role.rolname in ('ash_b4_reader', 'pg_monitor')
+              and coalesce(
+                aclcontains(
+                  rel.relacl,
+                  makeaclitem(
+                    grantee_role.oid,
+                    rel.relowner,
+                    'SELECT',
+                    false
+                  )
+                ),
+                false
+              );
+            assert v_direct_acl_count = 6,
+              format(
+                '#164: expected 6 direct reader ACLs before rebuild, got %s',
+                v_direct_acl_count
+              );
+
+            select count(*) into v_member_acl_count
+            from pg_class as rel
+            join pg_namespace as nsp on nsp.oid = rel.relnamespace
+            cross join pg_roles as member_role
+            where nsp.nspname = 'ash'
+              and rel.relname in ('query_map_all', 'query_map_0', 'sample_0')
+              and member_role.rolname = 'ash_b4_pgmon_member'
+              and coalesce(
+                aclcontains(
+                  rel.relacl,
+                  makeaclitem(
+                    member_role.oid,
+                    rel.relowner,
+                    'SELECT',
+                    false
+                  )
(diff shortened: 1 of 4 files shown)


NikolayS:
Pre-merge verification complete after rebasing onto `main` (`8eff377`). `codex review --base origin/main` stood in for REV because REV is unavailable; its first pass found two real least-privilege gaps, both now fixed, regression-covered, and clean on a second independent review.

Conflict resolution:
- In `sql/ash-install.sql`, retained the corrected #129 catalog formula `(num_partitions - 2) * rotation_period` while preserving #181's automatic complete reader-bundle replay wording and behavior.

RED / GREEN:
- #164 RED on current `main`: after `rebuild_partitions`, both a configured reader and a `pg_monitor` member failed on `query_map_all`; direct ACL count was zero. GREEN: schema/function/relation bundle is restored and both readers see the seeded key. Added a function-only-role regression so partial grants cannot be widened into a full bundle.
- #166 RED on current `main`: revoking protected owner/current roles succeeded or failed only after partial ACL mutation. GREEN: protected targets are rejected before mutation; legacy `_admin_funcs()` reader grants are scrubbed. Added delegated-GRANT-OPTION coverage and changed the legacy scrub to `CASCADE` so stale delegated access cannot survive.
- #169 RED on current `main`: `take_sample()` resolved slot 0 but stamped slot 1 after a probe changed `current_slot` (two reads). GREEN: it captures once, resolves and stamps slot 0, and records one `current_slot` read.

The rebuilt ACL logic remains dynamic: it derives the non-admin helper closure rather than freezing a function list. This is important for the upcoming #189 helper; after #189 lands I will explicitly verify `ash._raw_oldest_sample()` receives reader EXECUTE through this bundle.

Targeted PostgreSQL verification passed for all three scenarios, YAML parses, `git diff --check` is clean, and the second Codex review returned no findings.

NikolayS:
CI follow-up: the first rebased run exposed a stale v1.5 upgrade-test oracle, not a product regression. The upgrade correctly scrubbed the legacy reader grant on newly-admin `_admin_funcs()`, while the old equivalence sweep still required every historically reader-visible name to retain EXECUTE.

Fixed in `34671bb` by filtering both upgrade/re-apply sweeps through the current dynamic `ash._admin_funcs()` exclusion and adding explicit assertions that `_admin_funcs()` is denied after both transitions. The full affected v1.5 → current → wrapper re-apply step now passes locally on PostgreSQL 17 with cron absent (matching the first failing matrix variant), including state preservation and reader-surface behavior.

Issue #167, opened 2026-07-27
Decision request: define ash.report coverage and sub-day rotation contracts


samo-agent:
> **Owner decision required.** These are scope and contract choices, not unresolved empirical questions. Leave this issue unassigned until the owner chooses both.

Audit basis: `origin/main` at `37a0e2a`. Decision findings C1 and C3 in #160.

### Agreed facts

`ash.report()` counts only `ash.rollup_1m` coverage and returns NULL when none exists at `sql/ash-install.sql:5109-5115`. Its payload hardcodes `coverage.source = 'rollup_1m'` at `sql/ash-install.sql:5414-5418`.

The audit reproduced a window where `ash.aas()` reported `3.10 AAS` from raw and `ash.top()` returned two wait events while `ash.report()` returned SQL NULL with no NOTICE. The internal design document states the rollup-only contract, but the shipped README Machine report section and catalog comment at `sql/ash-install.sql:5769-5770` do not state the dependency.

A behavioral fallback to `rollup_1h` is not derivable. The report needs per-class, per-minute data; `rollup_1h.minute_counts` stores per-minute totals without the class breakdown. Synthesizing the payload would recreate the false precision tracked by #130.

### Option A — yes, this dependency is a user-facing contract

Apply N5's cheap half now: document the `rollup_1m`-only dependency in shipped user-facing text and emit a NOTICE naming the source that does hold the requested window when report returns NULL for lack of minute-rollup coverage. Defer any behavior/source change to 2.1.

### Option B — no, the internal contract is sufficient

Accept bare NULL with no NOTICE when only raw or `rollup_1h` holds the window, and explicitly decline N5's documentation/NOTICE change.

### Decision test

If a scraper author reading only shipped user documentation could interpret NULL as “pg_ash has no data” when pg_ash holds the window in another source, the documentation gap is real and Option A applies. Both audit lenses reproduced and agreed on the behavior; more empirical testing will not settle this scope choice.

If Option A is chosen, regress raw-only and `rollup_1h`-only coverage: report remains NULL, the NOTICE identifies the available source, and shipped docs state the dependency.

### Agreed facts

`ash.config.rotation_period` is a writable interval with a one-day default at `sql/ash-install.sql:254`, but there is no documented setter. `ash.start()` schedules `ash_rotation` with a hardcoded daily expression at `sql/ash-install.sql:1945-1950` and never derives it from `rotation_period`.

[...]

PR #184, opened 2026-07-27
release: make SQL installer atomic

samo-agent:
## What changed

- moves `ON_ERROR_STOP` and transaction ownership into `sql/ash-install.sql`
- lets `sql/migrations/ash-1.5-to-2.0.sql` delegate transaction ownership to the installer
- replaces #125's non-discriminating namespace check with a forced mid-install failure that asserts the `ash` schema is absent
- retains the three discriminating incompatible-config rollback assertions from #125
- documents that `\i` inside an existing transaction commits the caller's outer work

## Scope

This supersedes and rescopes draft PR #125. It intentionally does not modify the five finalized migration scripts protected by `docs/RELEASE_PROCESS.md:65`.

The rescope protects the fresh installer and the three canonical migration legs that replay it (1.3→1.4, 1.4→1.5, and 1.5→2.0): 4 of 7 primary documented entrypoints, or 3 of 6 canonical migration legs plus fresh install. Any path that reaches the installer no longer leaves the audited 23-of-62 partial schema after a mid-installer failure.

It does not make the early 1.0→1.1, 1.1→1.2, or 1.2→1.3 canonical legs atomic. Upgrade starts at 1.0, 1.1, or 1.2 therefore retain unsafe early legs. The frozen 1.4→1.5 shim also retains its own transaction bookends, so the full chain succeeds with PostgreSQL's expected nested/no-transaction warnings.

Relates to #124.

## RED on current `origin/main`

```text
complete installer exit code: 0
complete installer function count: 62
forced installer exit code: 3
psql:sql/ash-install.sql:2643: ERROR:  forced mid-install failure at ash.rollup_1m
1 schema, 23 of 62 functions survive
ERROR:  failed installer left a partial ash schema (23 of 62 functions survive)
CONTEXT:  PL/pgSQL function inline_code_block line 11 at ASSERT
RED assertion exit code: 3
ERROR:  function ash._color_on() does not exist
broken ash.status() exit code: 1
```

The retained incompatible-config fixture is also RED on main:

```text
fresh installer exit code: expected 3, got 0
53 functions, 3 tables remain
```

## GREEN on PostgreSQL 17.9

```text
complete_installer_function_count=62
incompatible_psql_rc=3
incompatible_three_assertions=GREEN
event_psql_rc=3
event_schema_count=0
event_surviving_function_count=0
DO
event_atomicity_assertion=GREEN
caller_marker_rows_after_caller_rollback=1
caller_transaction_persistence_assertions=GREEN
```

Additional validation:

[...]

PR #152, opened 2026-07-10
fix: correct admin documentation contracts

NikolayS:
## What changed

- limits the README's exact `'yes'` confirmation contract to `rebuild_partitions` and `uninstall`
- corrects `rotate()` and `rebuild_partitions()` catalog descriptions to `(num_partitions - 2) * rotation_period`
- adds two-sided README and installed-catalog contract guards
- uses literal `position()` assertions so underscores cannot act as `LIKE` wildcards

## Rebase integration

Rebased by hand onto `origin/main` at `fb5935d`. The known #129 workflow conflict was resolved by retaining the complete `Test 2.0: raw retention ignores the excluded ring slot` step first, immediately followed by the renamed `Test v1.4: status output and admin catalog contracts` step. Both pass.

## Validation

Postgres 17.9:

- RED catalog: the branch's named contract step against base SQL exits 3 at `rotate description has the wrong raw-retention formula`
- GREEN catalog at `3769482`: exact corrected wording is present and both old formulas are absent
- RED docs: the branch guard against base docs exits 1 because the broad confirmation claim remains
- GREEN docs: the complete guard passes
- preserved #129 N=6 and retention-boundary step passes
- actionlint, YAML ordering assertion, and `git diff --check` pass
- `codex review --base origin/main` reports no actionable regression

Closes #147

PR #185, opened 2026-07-27
release: prevent cleanup from deadlocking rotation

### LINKED ISSUE #162: Release blocker: rollup_cleanup can permanently deadlock rotation
## Summary

`ash.rollup_cleanup()` can delete the minute-rollup rows that `ash.rotate()` requires before truncating an endangered raw slot. `ash.rollup_minute()` then cannot reconstruct them because its watermark only advances. Rotation fails forever, the current raw partition grows without bound, and pg_cron records the failed rotation as SUCCESS because `rotate()` returns text instead of raising.

Audit finding: B2 in #160.

## Location

- Rotation completeness comparison and failure: `sql/ash-install.sql:1408-1433`.
- Forward-only minute-rollup watermark: read at `sql/ash-install.sql:2895-2897`, processing starts from it at `sql/ash-install.sql:2948-2951`, and it advances at `sql/ash-install.sql:3097-3103`.
- Minute-rollup retention deletion: `sql/ash-install.sql:3258`.

## Reproduction

With the default three partitions and one-day rotation, lower only the allowed minute-rollup retention:

```sql
update ash.config
set rollup_1m_retention_days = 1
where singleton;
```

Seed a 36-hour-old raw minute, then run the three shipped maintenance jobs in their scheduled order. Rotation returns this forever:

```text
failed: pre-truncation rollup incomplete for 1 group(s) in slot 2; slot not truncated
```

The documented catch-up call cannot repair it:

```sql
select ash.rollup_minute(1000000);
```

```text
0
```

The same state is reachable with stock 30-day minute-rollup retention through the README's documented raw-retention runbook:

```sql
select ash.rebuild_partitions(32, 'yes');
```

At truncation, endangered rows are 30–31 days old. The confirmed truncate-target sequence is `2 3 4 … 31 0 1 2 3 4 5`.

## Observed vs expected

Observed: cleanup removes evidence required by the rotation guard; the watermark prevents reconstruction; every later rotation returns the same failure text while pg_cron records successful execution.

Expected: the scheduled cleanup and rotation jobs must remain mutually compatible. An endangered slot should be rolled up and truncated, or an incompatible retention/partition configuration should be rejected before it can deadlock the ring. Repeated failures must be observable.

## Impact

Rotation never resumes. Every new sample accumulates in the frozen current slot until storage fills, defeating the ring's purpose. `ash.status()` has no failure metric; only `time_since_rotation` grows. Recovery requires manually rewinding `ash.config.last_rollup_1m_ts` or performing a destructive rebuild.

## Suggested fix

- Exclude raw groups already past `rollup_1m` retention from the completeness gate, and/or let `rotate()` rewind `last_rollup_1m_ts` to the endangered slot's `min(sample_ts)` before calling `rollup_minute()`.
- Validate `(num_partitions - 1) * rotation_period <= rollup_1m_retention_days` when rebuilding partitions or changing retention.
- Surface consecutive rotation failures in `ash.status()`.

## Interaction

No in-flight PR addresses this finding. The owner decision in #160 §4 C3 affects whether sub-day `rotation_period` should be rejected or scheduled dynamically, and one validation may share this fix. It does not resolve this blocker: the deadlock also reproduces with one-day rotation, 32 partitions, and stock 30-day rollup retention.

NikolayS:
Fixes #162
Refs #167

## Summary

`ash.rollup_cleanup()` could delete minute rollups that `ash.rotate()`'s pre-truncation completeness guard still required. Because `ash.rollup_minute()` advances its watermark only forward, rotation then failed forever and the raw partition ring stopped draining.

This change deliberately excludes raw minute groups already older than `rollup_1m_retention_days` from the pre-truncation gate. Those groups are consciously abandoned: cleanup is entitled to delete their rollups, rewinding would recreate data that cleanup would immediately remove, and it would turn rotation into an unbounded historical catch-up. Supported configurations are prevented from reaching this state by validating:

```text
(num_partitions - 1) * rotation_period <= rollup_1m_retention_days
```

The validator runs on relevant `ash.config` writes and before `ash.rebuild_partitions()` changes any state. Existing unsafe legacy geometry can still load the installer and drain through the expired-data escape hatch; the next geometry-changing write must make it safe.

## Observability and contract

Known text-return failure paths increment `ash.status()` metric `consecutive_rotate_failures`; a successful rotation resets it to `0`. Counter recording has its own bounded, best-effort lock timeout so observability cannot mask the stable `rotate()` result when `ash.config` is concurrently locked.

I retained the text return contract. Raising this late would change the public/admin API and, in the same transaction, roll back the counter meant to expose the failure. The status metric is the conservative compatible signal.

## Rotation-period decision

The C3 half of #167 is implemented as decided: sub-day and fractional-day `rotation_period` values are unsupported. A declarative CHECK plus trigger validator accepts whole days only (minimum one day), with actionable remediation. A file-level installer transaction makes a legacy sub-day value return the exact actionable validation error and rolls the attempted upgrade back atomically; after correcting the value, the operator reruns the installer. `ash.start()` keeps the owned rotation job on the documented daily `0 0 * * *` schedule; multi-day periods work through `rotate()`'s due-time check. README, catalog comments, and `ash.status()` document the contract.

## RED / GREEN

Both RED paths produced exactly:

```text
failed: pre-truncation rollup incomplete for 1 group(s) in slot 2; slot not truncated
```

The one-day-retention reproduction also proved forward catch-up returned `0`. The stock 32-partition run reproduced the truncate-target sequence:

```text
2 3 4 5 6 7 8 9 10 11 12 13 14 15 16 17 18 19 20 21 22 23 24 25 26 27 28 29 30 31 0 1 2 3 4 5
```

GREEN is exactly:

```text
rotated: slot 0 -> 1, truncated slot 2 (sample + query_map)
```

The expired rollup stays deleted, the raw target is truncated, and the failure metric is `0` after success.

## Validation

- PostgreSQL 14.23, 15.18, 16.14, 17.9, 18.4, and 19beta2: fresh install, #162 behavior/geometry/sub-day/32-slot coverage, installer reapply — PASS
- Complete #81 concurrency suite, including counter-row lock timeout and outer exception-handler timeout — PASS on PostgreSQL 14–19
- Expired-only rollup/rotation lock-order regression: RED deadlock, GREEN bounded busy result, raw preservation, successful retry — PASS on PostgreSQL 17
- pg_cron same-name cross-owner collision: unrelated job unchanged, caller-owned daily job created — PASS on PostgreSQL 17
- Full discovered upgrade chain and fresh-vs-upgrade schema equivalence on PostgreSQL 17.9 — PASS
- Legacy 12-hour installer path: exact actionable rejection with full rollback, then correction to one day and CHECK restoration — PASS on PostgreSQL 14–19
- Workflow YAML, extracted Bash syntax, docs lint, and `git diff --check` — PASS
- Local five-lens REV fallback — PASS; cloud `claude ultrareview` could not launch because its free-review quota was exhausted. Initial actionable findings were fixed and re-reviewed.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -5491,7 +5491,7 @@ jobs:
             select 1 from ash.config where singleton for update;
             select pg_advisory_xact_lock(
               hashtext('pg_ash')::int4,
-              hashtext('pg_ash_rollup')::int4
+              hashtext('pg_ash_config_test')::int4
             );
             select pg_sleep(5);
             commit;
@@ -5503,7 +5503,7 @@ jobs:
               from pg_locks
               where locktype = 'advisory'
                 and classid = hashtext('pg_ash')::int4::oid
-                and objid = hashtext('pg_ash_rollup')::int4::oid
+                and objid = hashtext('pg_ash_config_test')::int4::oid
                 and granted
                 and pid <> pg_backend_pid()
             )
@@ -5521,6 +5521,86 @@ jobs:
           ")"
           if [ "$rotate_state" != "0 0" ]; then
             echo "::error::outer lock failure changed state: $rotate_state" >&2
+            exit 1
+          fi
+
+          echo "=== #81F: expired-only rotation must serialize with rollup ==="
+          reset_issue_81_state
+          "${psql_base[@]}" << 'SQL'
+          begin;
+          set local session_replication_role = replica;
+          update ash.config
+          set rollup_1m_retention_days = 1
+          where singleton;
+          insert into ash.sample (
+            sample_ts, datid, active_count, data, slot
+          )
+          values (
+            ash.ts_from_timestamptz(
+              date_trunc('minute', now()) - interval '36 hours'
+            ),
+            0::oid,
+            1::smallint,
+            array[-1,1,0]::integer[],
+            2::smallint
+          );
+          commit;
+          SQL
+          "${psql_base[@]}" -c "
+            begin;
+            select pg_advisory_xact_lock(
+              hashtext('pg_ash')::int4,
+              hashtext('pg_ash_rollup')::int4
+            );
+            lock table ash.sample_2 in access share mode;
+            select pg_sleep(1);
+            update ash.config
+            set last_rollup_1m_ts = last_rollup_1m_ts
+            where singleton;
+            commit;
+          " &
+          expired_rollup_pid=$!
+          wait_until_sql_true "rollup advisory and sample_2 locks" "
+            select exists (
+              select 1
+              from pg_locks advisory_lock
+              where advisory_lock.locktype = 'advisory'
+                and advisory_lock.classid = hashtext('pg_ash')::int4::oid
+                and advisory_lock.objid = hashtext('pg_ash_rollup')::int4::oid
+                and advisory_lock.granted
+                and advisory_lock.pid <> pg_backend_pid()
+                and exists (
+                  select 1
+                  from pg_locks sample_lock
+                  where sample_lock.pid = advisory_lock.pid
+                    and sample_lock.locktype = 'relation'
+                    and sample_lock.relation = 'ash.sample_2'::regclass
+                    and sample_lock.mode = 'AccessShareLock'
+                    and sample_lock.granted
+                )
+            )
+          "
+          rotate_result="$("${psql_base[@]}" -tAc "select ash.rotate()")"
+          if [ "$rotate_result" != "failed: pre-truncation rollup busy; slot 2 not truncated" ]; then
+            echo "::error::expired-only busy-rollup result mismatch: $rotate_result" >&2
+            exit 1
+          fi
+          wait "$expired_rollup_pid"
+          assert_raw_sample_survived "expired-only rollup race"
+          rotate_result="$("${psql_base[@]}" -tAc "select ash.rotate()")"
+          if [ "$rotate_result" != "rotated: slot 0 -> 1, truncated slot 2 (sample + query_map)" ]; then
+            echo "::error::expired-only rotation did not recover: $rotate_result" >&2
+            exit 1
+          fi
+          expired_state="$("${psql_base[@]}" -tAc "
+            select current_slot || ' ' ||
+                   (select count(*) from ash.sample_2) || ' ' ||
+                   consecutive_rotate_failures
+            from ash.config
+            where singleton
+          ")"
+          if [ "$expired_state" != "1 0 0" ]; then
+            echo "::error::expired-only recovery state mismatch: $expired_state" >&2
             exit 1
           fi
 
@@ -5649,6 +5729,7 @@ jobs:
             v_enabled boolean;
             v_input interval;
             v_message text;
+            v_other_rotation_job bigint;
             v_raised boolean;
             v_rotation_job bigint;
             v_sqlstate text;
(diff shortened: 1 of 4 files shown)


NikolayS:
## REV code review report

Cloud `claude ultrareview origin/main --timeout 30` could not launch because the account reported: `Free ultrareviews used. Turn on usage credits to continue.` I therefore ran the current cloned REV agent prompts locally through Claude Code under all five standard lenses: bug hunter, security reviewer, test analyzer, guidelines checker, and docs reviewer. Three independent Codex reviewers also audited the SQL and PostgreSQL 14–19 behavior.

### Findings resolved

1. **Failure-counter lock contention could mask `rotate()`'s stable text result.** The counter write is now isolated in `_record_rotate_failure()` with its own function-scoped 250 ms `lock_timeout`; failure to record warns but returns the original result. #81D covers the advisory-busy path and #81E covers the outer `lock_not_available` exception path where the caller's `SET LOCAL` is rolled back.
2. **Legacy sub-day validation originally ran too early for an autocommit installer.** The hard validation is now the installer's final block. The regression uses ordinary `psql -f`, replaces `rotate()` with a sentinel first, and proves the real cutoff-aware body was committed before exit status `3` and the exact actionable error.
3. **Legacy upgrade coverage and remediation text were incomplete.** The workflow now asserts the actual invalid legacy reapply, exact message, correction to one day, CHECK restoration, and successful reapply.
4. **Duplicated day-granularity predicates lacked synchronization guidance.** All declarative/imperative copies now cross-reference one another and exact 12-hour, 36-hour, one-day, and two-day behavior is covered.
5. **One catalog comment had a missing relative clause.** Corrected after the final review.

The frozen-SQL/branch-name warning was dismissed because TASK.md explicitly mandates `sql/ash-install.sql` and `agent/w2-b2`; this PR's `release:` title satisfies the repository's SQL guard. The daily rotation schedule repair was re-reviewed and confirmed intentional: it matches the documented owner decision and the existing sampler schedule-resync semantics.

### Final verdict

- **Blocking:** 0
- **Security:** no findings
- **Runtime/correctness:** no unresolved findings at confidence ≥ 4
- **Tests:** no unresolved coverage gaps at confidence ≥ 4
- **Docs/guidelines:** no unresolved findings after the cosmetic correction
- **REV verdict: PASS**

The final review ran on the logical patch at `3814b89`; it was then rebased onto `88b69cf` as `62bf1ef`. The only conflict combined upstream's terminal hourly-rollup schedule migration before this PR's terminal legacy rotation-period validator. Post-rebase PostgreSQL 17.9 #81A–E, complete #162, full upgrade chain, schema equivalence, YAML/Bash syntax, docs lint, and `git diff --check` all passed. Cross-major validation before the final rebase passed on PostgreSQL 14.23, 15.18, 16.14, 18.4, and 19beta2.

NikolayS:
## Merge verification / REV report

Reviewed the rebased patch at `cdf1f7a` against `8540219` with Codex review plus an independent PostgreSQL verification.

### Rebase and integration

- Rebased by hand onto current `main`; the installer conflict preserves #142 sampling behavior, #184 file-level transaction ownership, #181 reader-bundle replay/admin exclusions, and #185's retention/rotation geometry.
- Corrected the legacy 12-hour regression for #184's atomic installer contract: the rejected install now proves full rollback, then correction and reapply converge.
- Added the requested release-note disclosure that upgrades intentionally remove legacy reader `EXECUTE` on `ash._admin_funcs()` while preserving every non-admin reader grant.

### Review findings resolved

1. **Expired-only rotation could still deadlock a concurrent rollup.** RED reproduced PostgreSQL `deadlock detected` at the slot-2 `TRUNCATE`. `rotate()` now takes the rollup advisory lock before every truncate, including empty/expired-only targets. GREEN returns the bounded busy result, preserves the raw row and slot, then succeeds on retry with the failure counter reset.
2. **Daily schedule repair could mutate another user's same-named pg_cron job.** The lookup is now constrained by both `username = current_user` and `database = current_database()`. The regression removes the caller's job, installs an unrelated same-name row, and proves `ash.start()` leaves it unchanged while creating the caller's daily job.
3. **The new race fixture leaked unsafe geometry into the later legacy-upgrade recovery.** The #81 fixture now restores the safe baseline. The complete #81 → #162 sequence, including legacy rejection/recovery, passes with pg_cron.

A review probe that manually granted only `rotate()` or `rebuild_partitions()` to a non-owner was not treated as a regression: the shipped contract explicitly defines operations/admin and their internal helpers as owner-only, and individual SECURITY INVOKER admin grants have never supplied a transitive helper privilege closure.

### Verification

- Exact RED/GREEN deadlock reproduction and retry state: PASS
- Existing #162 cleanup → forward catch-up `0` → successful truncate: PASS
- Complete #81A–F followed by complete #162 and atomic legacy recovery on PostgreSQL 17 with pg_cron: PASS
- #181 privilege/slot assertions on the prospective tree: PASS
- pg_cron cross-owner schedule regression: PASS
- Workflow YAML, actionlint, docs guard, and `git diff --check`: PASS
- Final exact-head delta review: no findings

**Verdict: PASS — no blocking findings.**

NikolayS:
CI follow-up: the first rebased matrix run failed uniformly in the restored #152 catalog-contract assertion. The conflict resolution kept the correct `(num_partitions - 2) * rotation_period` boundary but changed the exact public sentence from “is approximately ….” to “is roughly … plus …”, so both `rotate()` and `rebuild_partitions()` descriptions failed the existing contract gate.

`ee63c73` restores the exact required sentence and retains the added current-partial-period clarification as a separate sentence. The exact catalog assertions pass on a fresh PostgreSQL 17 install; `git diff --check` is clean. This is a real integration correction, not a CI retry.

PR #186, opened 2026-07-27
release: clarify AAS cadence and coverage limits

NikolayS:
## Summary

- documents that 2.0 weights retained raw and rollup appearances with the current `ash.config.sample_interval`, so changing cadence rescales history
- documents that successful idle ticks write no row, so stored activity cannot prove sampler coverage or distinguish idle sampling from an outage
- documents that intervals greater than one minute can assign a full tick to one minute and overstate minute extrema
- marks all affected AAS user stories partial and adds the limitation to the README, release notes, API/example blueprints, and the `sample_interval` catalog comment

## Release decision

No genuinely low-risk behavioral fix exists this late in 2.0. A heartbeat alone would not fix historical weighting, while a per-row interval alone would not represent sampled idle time. The complete storage fix remains tracked for a later release in #137; this PR contains the release-local documentation and catalog containment, including #175's corrected no-observation language.

## Rebase and conflict resolution

Hand-rebased onto current `main` at `28136cfcadb6628975363aaaf17dd7b46d2f6f1a`.

The overlapping documentation was resolved by retaining:

- #182's effective grain, per-window provenance, and `rollup_1h_flat` contracts
- #180's reader inversion, until-only defaults, report last-day default, rollup_1m-only report disclosure, and NULL+NOTICE behavior
- #189's distinction between the logical `raw_retention_start` planning boundary and physical attribution evidence
- #187's statement that no maintained 2.0 benchmark replaces the retired suite

The examples-wide default sentence was also corrected after review: bare readers mean the last hour except `report()` (last day), while until-only means the preceding hour.

## RED reproduction

On current main before this documentation/catalog change:

- the same 12 appearances read as `backend_seconds = 12.00`, `avg_aas = 0.20` at a 1-second config, then `60.00` / `1.00` after changing only the current config to 5 seconds; the same 5× rescaling persisted after minute rollup and raw truncation
- 12 successful idle `take_sample()` calls stored zero rows and produced the same `data_points = 0`, NULL-AAS timeline representation as a minute with no sampler calls
- issue #137's greater-than-one-minute reproduction assigns a whole coarse tick to one minute and can report a one-minute peak above observed concurrency

## GREEN verification

Exact head `18a90f1fa54bc04db3fe2092af1515892ac7ebd4`:

- fresh PostgreSQL 17 install from an exact archive: PASS
- installed `ash.config.sample_interval` catalog comment contains current-cadence rescaling, >1-minute extrema overstatement, and idle-vs-outage limitations: PASS
[...]

PR #182, opened 2026-07-27
release: fix rollup_1h grain honesty

NikolayS:
## Summary

Fixes the `rollup_1h` grain collapse by separating requested/effective buckets from the grain actually retained, exposing provenance, and returning `NULL` instead of fabricated peak/p99 values when retained data is too coarse.

- makes `ash.aas()`, `ash.timeline()`, `ash.top()`, and `ash.compare()` follow the same grain-honesty rule
- routes the `database` dimension through valid per-database `minute_counts`
- labels upgraded/synthesized arrays as `rollup_1h_flat`
- rejects incomplete or sum-inconsistent `minute_counts` in both `rollup_hour()` and the v1.5 → 2.0 backfill
- snaps hour-only dimensional windows outward and exposes effective bounds instead of raising
- gives `ash.summary()` separate headline and wait/query drill provenance so widened drill values cannot inherit narrower labels
- corrects the N8 catalog claim and points the coverage-model limitation to #137
- keeps zero-argument and `now()`-relative APIs working

Fixes #161.
Fixes #168.
Fixes #175.
Fixes #130.

This supersedes #131; I recommend closing #131 because this branch satisfies #130 without its hard-error/default-call and database-precision regressions.

**This PR needs a fresh rebase onto `origin/main` immediately before merge.** It is intentionally a draft and must not be merged as-is later without that rebase.

## Audit fixture behavior

| Case | Before | After |
|---|---|---|
| B1a overall `aas()` | avg/peak/p99 `1.15 / 10.00 / 4.69` | unchanged |
| B1a `top('wait_event')` | `0.77 / 0.77 / 0.77` | avg `0.77`; peak/p99 `NULL` because the read is hourly |
| B1a `top('database')` | `1.15 / 1.15 / 1.15` | `1.15 / 10.00 / 4.69` from the per-database minute array |
| B1b filtered IO | avg/peak/p99 `2.30 / 2.30 / 2.30` at a requested minute bucket | avg `2.30`, seconds `8280.00`, effective bucket `1 hour`, peak/p99 `NULL`; an explicit one-hour request remains `2.30 / 2.30` |
| B1c identical dimensional windows | no source columns; avg delta `0.00` but peaks `10.98 / 600.00` and p99 `10.98 / 246.59` | sources `rollup_1h / rollup_1m`, averages `10.98 / 10.98`, delta `0.00`, and both peak/p99 pairs `NULL` because retained grains differ |
[...]

PR #180, opened 2026-07-27
release: reject inverted reader windows and clarify diagnostics

NikolayS:
## Summary

- Reject inverted windows in `ash.aas`, `timeline`, `top`, `report`, `chart`, `samples`, `summary`, and both `compare` windows; anchor an `until`-only call to the preceding hour while retaining the #63 overflow/degenerate clamp. Closes #165.
- Correct the pg_monitor opt-out and reader re-apply documentation, and warn with the roles whose complete reader bundle is restored. Closes #170 and #171.
- Document the `rollup_1m`-only report contract in the catalog and README and emit a NOTICE naming raw or `rollup_1h` alternate coverage without changing the SQL NULL return behavior. This implements C1 Option A. Closes #172; refs #167.
- Correct the pg_cron database-scoping diagnostic and distinguish unset, current-database, and alternate-database targets. Closes #173.
- Clarify that lifecycle/admin signatures are unchanged while parameter names were de-prefixed. Closes #174.
- Append the user-visible changes to the release notes.

## Rebase resolution

Hand-rebased onto main `c28fce3`. The resolution preserves #181's dynamic reader/helper grants, #182's honest `rollup_1h_flat` planning, and #189's logical/physical raw-retention boundary and active-slot exclusion. No new reader helper was introduced.

## Validation

- Current main RED: until-only calls answered a now-relative window, and every affected reader accepted explicit inverted bounds.
- Rebased head GREEN: exact B5 defaults/errors for all readers and both compare windows; #63 horizon clamps remain intact.
- Alternate raw/hourly coverage NOTICEs occur exactly once while `ash.report()` remains SQL NULL; uncovered and sparse invalid-hour windows remain silent.
- Installer re-grant warning and pg_monitor behavior pass; #181 helper bundle remains complete.
- #182 grain/source behavior and #189 retention/active-slot probes pass.
- PG17 cron-off and real pg_cron-preloaded cron-on diagnostic branches pass; the postmaster-only GUC is never mutated on cron-enabled servers.
- YAML parse, documentation guards, fresh install, and `git diff --check` pass.
- Codex review (standing in for REV): final pass, no actionable regressions.

PR #189, opened 2026-07-27
release: fix raw retention boundary

### LINKED ISSUE #163: Release blocker: fresh-install tie drills reject retained raw samples
## Summary

The README's flagship wait-to-query drill hard-errors on a fresh install even though all requested raw data is present. `ash.status().raw_retention_start` is the oldest sample held, but the tie guard compares it with a minute-floored reader boundary. The error's own round-up advice then discards the only partial minute and returns no rows.

Audit finding: B3 in #160.

## Location

- `_raw_retention_start()` definition and physical `min(sample_ts)`: `sql/ash-install.sql:3430-3436`.
- Tie-retention guard: `sql/ash-install.sql:3527-3572`; the mismatched comparison is at `sql/ash-install.sql:3541-3543` and the round-up remediation is at `sql/ash-install.sql:3558-3572`.
- `ash.report()` already floors the raw start to a minute at `sql/ash-install.sql:5130-5131`.
- Documented drill: `README.md:190-200`.

## Reproduction

On a fresh install with eight samples, run the README query verbatim:

```sql
select *
from ash.top(
  'query_id',
  since => now() - interval '5 minutes',
  wait_event => 'Lock:tuple',
  order_by => 'peak',
  n => 5
);
```

Observed:

```text
ERROR: pg_ash: this drill needs raw samples; raw retention starts at 22:34:51 but the requested window starts at 22:30:01. Narrow the window to start at or after 22:35:00…
```

The untied `ash.top('wait_event')` reads the same data successfully. Passing the boundary exposed by status still raises:

```sql
select *
from ash.top(
  'query_id',
  since => (
    select value::timestamptz
    from ash.status()
    where metric = 'raw_retention_start'
  ),
  wait_event => 'Lock:tuple'
);
```

```text
ERROR: ... raw retention starts at 22:34:51 but the requested window starts at 22:34:51
```

Following the error's advice and using `since => '22:35:00'` returns zero rows because the entire partial first minute is discarded.

## Observed vs expected

Observed: the guard reports retention loss for data that was never lost. Its published status boundary cannot be used as a valid `since`, and its suggested correction silently erases the available drill window.

Expected: fully retained fresh data must be queryable. Status must expose a planning boundary that readers accept, and remediation must not round past retained data.

## Impact

The payoff step of the documented investigation flow fails on first use, then its remediation can falsely report that no query was responsible. The same trap recurs after every `ash.rebuild_partitions()` and after sampler outages. With `ash.start()`'s one-second default, the first sample lands mid-minute with 59/60 probability, so this is the default path rather than an edge case.

## Suggested fix

Separate “oldest sample held” from the usable raw-retention boundary, floor both sides of the tie-guard comparison consistently, and round the remediation boundary down rather than up.

## Interaction

PR #129 fixes #128's excluded-slot boundary by moving `_raw_retention_start()` newer. That fixes a distinct too-old boundary and makes this partial-first-minute defect strictly worse. Resolve B3 with, or before, #129; do not fold it into or close #128.

samo-agent:
## Summary

- separate exact physical raw coverage (`ash._raw_oldest_sample()`) from the reusable minute-aligned planning/loss boundary (`ash._raw_retention_start()`)
- floor both sides of the wait/query tie guard and print the same down-rounded remediation boundary it accepts
- keep source selection and report query attribution tied to physical coverage while status/report coverage expose the logical boundary
- preserve still-readable evidence after delayed rotation by choosing the older of ring geometry and the physical oldest minute
- add three independent exact B3 regressions, including the README query copied verbatim after a no-argument `ash.start()` call
- clarify the logical-vs-physical report and status contracts

## Root cause

`ash._raw_retention_start()` exposed the exact first sample second. Readers query minute buckets, so status could not be fed back into `since`, and the guard rounded its advice up past the only partial minute. The same physical timestamp was also being used for source attribution and public planning even though those are different contracts.

## Verification

- all three exact failures reproduced independently on current `main` (`1fcf9e7`): README drill errors, status self-feed errors, and advice rounds up to an empty result
- PostgreSQL 17.9 fresh install and full discovered upgrade chain: pass
- exact README/default-start/status/advice B3 step: pass
- report attribution, retention split, delayed-rotation range guard, compare, and chart step: pass
- behavioral reader suite through the changed retention section: pass
- #185 cleanup/rotation regression: pass
- `grant_reader` dynamic helper closure: 41/41 current non-admin functions after #181, including direct EXECUTE/call of `ash._raw_oldest_sample()`; rebuild preserves it and revoke removes it
- workflow YAML parse and `git diff --check`: pass
- Codex review plus independent adversarial verification: approved with no blockers

## Rebase resolution

Rebased onto current `main` (`1fcf9e7`). The `ash.status()` catalog-comment conflict preserves both #185's day-granular rotation / `consecutive_rotate_failures` contract and this PR's logical-vs-physical retention contract. Two test fixtures that still used a now-invalid 30-minute `rotation_period` were converted to the deliberate one-day contract and their entirely-past/partial/delayed-evidence ranges were scaled without weakening any assertion.

## Merge sequencing

Land this retention-boundary change before rebasing sibling PRs #182 and #180. PR #181's dynamic non-admin helper enumeration is retained and verified for the new helper. Expect documentation conflicts in `blueprints/AAS_API.md` and `blueprints/AAS_EXAMPLES.md` when #182 and #186 are rebased.

Fixes #163

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -3060,7 +3060,7 @@ jobs:
           -- --- raw-retention exception (tie drill before raw retention) ----------------
           update ash.config
           set num_partitions = 3,
-              rotation_period = interval '30 minutes',
+              rotation_period = interval '1 day',
               rotated_at = date_trunc('minute', now())
           where singleton;
 
@@ -3070,7 +3070,8 @@ jobs:
             select base_ts into v_base from t2anchor;
             begin
               perform * from ash.top('query_id',
-                ash.ts_to_timestamptz(v_base - 3600), ash.ts_to_timestamptz(v_base - 3000),
+                ash.ts_to_timestamptz(v_base - 172800),
+                ash.ts_to_timestamptz(v_base - 172200),
                 wait_event => 'DataFileRead');
               raise exception 'expected raw-retention exception, none raised';
             exception when others then
@@ -6563,7 +6564,7 @@ jobs:
             v_slots smallint[];
           begin
             update ash.config
-            set rotation_period = interval '30 minutes',
+            set rotation_period = interval '1 day',
                 rotated_at = v_end
             where singleton;
             select ash._register_wait('active','CPU*','CPU*') into v_cpu;
@@ -6582,8 +6583,9 @@ jobs:
                            else array[v_cpu, 60]::int4[] end,
                       case when i = 5 then array[424242::bigint, 540]::int8[] else '{}'::int8[] end);
             end loop;
-            -- raw samples for the last 30 minutes ONLY (raw retention starts at
-            -- v_end-30min); worst minute carries the 424242 tie.
+            -- Raw samples cover the last 30 minutes only (physical coverage);
+            -- the logical planning boundary is v_end-1day. The worst minute
+            -- carries the 424242 tie.
             for i in 1..30 loop
               insert into ash.sample (sample_ts, datid, active_count, data, slot)
               values (ash.ts_from_timestamptz(v_end - (i || ' minutes')::interval), v_dp,
@@ -6610,8 +6612,10 @@ jobs:
             assert (j->'coverage'->>'minutes_with_data')::int = 120, 'coverage.minutes_with_data 120';
             assert (j->'coverage'->>'from')::timestamptz = v_end - interval '2 hours', 'coverage.from';
             assert (j->'coverage'->>'to')::timestamptz = v_end, 'coverage.to';
-            assert (j->'coverage'->>'raw_retention_start')::timestamptz = v_end - interval '30 minutes',
-              'coverage.raw_retention_start -30min, got ' || (j->'coverage'->>'raw_retention_start');
+            assert (j->'coverage'->>'raw_retention_start')::timestamptz
+                = v_end - interval '1 day',
+              'coverage.raw_retention_start -1day, got '
+                || (j->'coverage'->>'raw_retention_start');
 
             -- window whose extremes all predate raw retention: aas_* still present,
             -- attribution keys absent, availability flag false, never raises.
@@ -6628,8 +6632,8 @@ jobs:
             -- entirely past raw retention: unrecoverable, no "narrow" advice.
             v_raised := false;
             begin
-              perform * from ash.top('query_id', v_end - interval '2 hours',
-                                     v_end - interval '1 hour', wait_event => 'DataFileRead');
+              perform * from ash.top('query_id', v_end - interval '2 days',
+                                     v_end - interval '1 day', wait_event => 'DataFileRead');
               raise exception 'entirely-past tie drill should have raised';
             exception when others then
               if sqlerrm like '%entirely outside raw retention%'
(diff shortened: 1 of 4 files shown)


NikolayS:
## REV Code Review Report

- **PR:** NikolayS/pg_ash#189 — release: fix raw retention boundary
- **Reviewed commit:** `f1d7aa7`
- **Pipeline:** 8/8 checks passed (docs-lint; PostgreSQL 14, 15, 16, 17 cron on/off, 18, 19beta)

No issues found. Reviewed the final pushed diff for security, runtime/SQL correctness, exact regression coverage, repository guidelines, API contracts, documentation, upgrade behavior, and merge sequencing.

**Result: PASSED**

|---|---:|---:|---:|
| Security | 0 | 0 | 0 |
| Bugs | 0 | 0 | 0 |
| Tests | 0 | 0 | 0 |
| Guidelines | 0 | 0 | 0 |
| Docs | 0 | 0 | 0 |
| Metadata / API contract | 0 | 0 | 0 |

Manual evidence included the three independent #163 RED assertions on the #129 base, the copy-pasted README drill on a fresh default-start install, status-boundary self-feed, down-rounded remediation preserving all eight samples, delayed-rotation evidence, the full discovered upgrade chain, fresh/upgrade schema equivalence, and the complete reader-helper privilege closure.

Merge-sequencing note (non-blocking): land B3 before rebasing #180/#182; reconcile their shared report/workflow/source-provenance regions and rerun the exact boundary tests. #181 must retain dynamic non-admin helper enumeration.

---
*REV-assisted review (parallel analysis following [postgres-ai/rev](https://gitlab.com/postgres-ai/rev))*

NikolayS (commented):
## Codex review report — PASS

Reviewed rebased head `e724827` against current `main` (`1fcf9e7`). No remaining blocker.

- **RED:** all three exact #163 regressions fail independently on main: the README tie drill raises and advises the next minute; feeding `status.raw_retention_start` back into the drill raises; following the advice yields `{up,0,NULL,…}`.
- **GREEN:** the consolidated README/default-start/status/advice block passes on PostgreSQL 17.9, as do the full behavioral reader step and the report/compare/retention/chart step.
- **Rebase conflict:** combined #185's day-granular rotation and `consecutive_rotate_failures` status wording with #189's logical-vs-physical retention wording.
- **Semantic collision fixed:** two #189 fixtures still assigned a 30-minute `rotation_period`, which #185 deliberately forbids. Commit `e724827` retains the product validation and scales the synthetic windows to one-day geometry; the entirely-past, partial-overlap, delayed-evidence, and down-rounding assertions are unchanged in substance.
- **#181 privilege bundle:** 41/41 current non-admin functions are granted. `ash._raw_oldest_sample()` is explicitly EXECUTE-able/callable by the reader, survives `rebuild_partitions`, and is removed by `revoke_reader`. The historical PR-body count of 42 became 41 because #181 intentionally moved `_admin_funcs()` out of the reader set.
- **Previous-fix cross-check:** #185's cleanup/rotation regression still passes on this installer.
- **Additional verification:** fresh install, discovered full upgrade chain, YAML parse, and `git diff --check` pass. Automated Codex review and an independent adversarial review found no issues.

This is verification and collision resolution against the existing #160 audit; no assertions were deleted or weakened.

Issue #191, opened 2026-07-27
rollup_minute() and rollup_hour() return per-database row counts, not processed grains


samo-agent:
## Summary

On current `main` (`85402194873ccc607630e68b225e142649f2cd3b`), both public rollup functions violate their documented return-value contracts when one time grain contains more than one database:

- `ash.rollup_minute(integer)` is documented as “Returns minutes processed.”
- `ash.rollup_hour()` is documented as “Returns hours processed.”

Each instead returns the number of `(time grain, datid)` rows upserted. One completed minute/hour with two databases returns `2`, although exactly one minute/hour was processed.

Found while filling the systematic behavioral release-gate coverage requested by #160.

## Exact reproduction

Run this after a fresh install in a database other than `postgres` so the fixture has two distinct database OIDs:

```sql
truncate table ash.sample;
truncate table ash.rollup_1m, ash.rollup_1h;
truncate table ash.wait_event_map restart identity;

insert into ash.wait_event_map (state, type, event)
values ('active', 'CPU*', 'CPU*');

update ash.config
set current_slot = 0,
    sample_interval = interval '1 second',
    last_rollup_1m_ts = ash.ts_from_timestamptz(
      date_trunc('hour', statement_timestamp()) - interval '1 hour'
    ),
    last_rollup_1h_ts = null
where singleton;

with anchor as (
  select ash.ts_from_timestamptz(
    date_trunc('hour', statement_timestamp()) - interval '1 hour'
  ) as sample_ts
), database_ids as (
  select oid as datid
  from pg_database
  where datname in (current_database(), 'postgres')
), wait_id as (
  select id
  from ash.wait_event_map
  where state = 'active' and type = 'CPU*' and event = 'CPU*'
)
insert into ash.sample (sample_ts, datid, active_count, data, slot)
select anchor.sample_ts,
       database_ids.datid,
       1,
       array[-wait_id.id::int, 1, 0]::integer[],
       0
from anchor
cross join database_ids
cross join wait_id;

select count(distinct datid) as distinct_datids_in_one_minute
from ash.sample;

select 1 as expected_documented_minutes_processed,
       ash.rollup_minute(1) as actual_rollup_minute_return;
select count(*) as minute_rows_written from ash.rollup_1m;

update ash.config
set last_rollup_1m_ts = ash.ts_from_timestamptz(
      date_trunc('hour', statement_timestamp())
    ),
    last_rollup_1h_ts = ash.ts_from_timestamptz(
[...]

PR #188, opened 2026-07-27
release: add Docker release gate and normalize config upgrade

NikolayS:
## Summary

- adds repeatable Docker release-gate images and a safe local runner for PostgreSQL 14, 15, 16, 17, 18, and 19beta2;
- reuses the workflow's SQL assertions through shared test files, including previously missing degraded-mode coverage and real pg_cron scheduling/firing;
- closes the schema-equivalence blind spot by snapshotting columns in physical `attnum` order;
- fixes the resulting upgrade finding in the editable `ash-1.5-to-2.0.sql` migration: `ash.config` had fresh-vs-upgraded column ordinals that diverged beginning with v1.1;
- rebased onto current `main`, including #159, #145, #154, #151, #149, #133, and #127.

Relates to #160.

## `ash.config` red/green

The new ordinal-aware upgrade assertion was red before the migration fix on PostgreSQL 14 (and the same divergence affected every supported major):

```text
ash.config column order differs: {singleton,current_slot,sample_interval,rotation_period,include_bg_workers,encoding_version,rotated_at,installed_at,version,debug_logging,num_partitions,sampling_enabled,skipped_samples,missed_samples,rollup_1m_retention_days,rollup_1h_retention_days,rollup_min_backend_seconds,last_rollup_1m_ts,last_rollup_1h_ts,insert_errors,register_wait_cap_hits}
```

No allowlist was added. The final migration takes an `ACCESS EXCLUSIVE` lock, validates the exact expected columns/types/nullability and supported catalog properties, snapshots the singleton row plus defaults, constraints, indexes, owner, effective table/column ACLs, and comments, then performs an atomic rename/create/copy/`DROP RESTRICT`/replay. Unsupported dependencies or properties abort and roll back. A canonical table is a no-op on reapplication and retains its OID.

The regression seeds non-default data, custom defaults/ACL state, a check constraint, a partial index, and a standalone-UNIQUE/self-FK pair. It verifies preservation and applies the migration twice.

## Release-gate matrix

Command used on rebased HEAD:

```bash
PG_MAJORS="14 15 16 17 18 19beta2" \
RELEASE_GATE_JOBS=2 \
RELEASE_GATE_PULL=0 \
RELEASE_GATE_OUTPUT_DIR=/tmp/pg_ash-release-gate-postrebase-final3-20260727 \
devel/scripts/release_gate.sh all
```

| Surface | PG14 | PG15 | PG16 | PG17 | PG18 | PG19beta2 |
[...]

PR #192, opened 2026-07-27
release: add behavioral feature-gate coverage

samo-agent:
## Summary

Fills the per-feature behavioral release-gate gap documented in #160 and absorbs the complete release-gate foundation developed in #188.

- adds deterministic fixtures and exact assertions for every installed non-underscore public surface
- covers readers, lifecycle/admin side effects, reader grants/revokes/default `pg_monitor`, documented reader edges, and pgss text enrichment/degradation
- runs the canonical feature suite with both optional extensions, without pg_cron, without pg_stat_statements, and without either
- includes #188's pg_cron Docker images, runner, shared SQL assertions, `attnum`-ordered schema snapshot, and `ash.config` ordinal normalization
- retains #194's processed-grain contract: empty minute grains count as processed, caught-up calls return zero, and physical rows remain independently asserted

Refs #160, #188, #191, and #194.

## Final integration

Hand-rebased onto current main `2ccdf65` after #194. The two stacked #194 commits were dropped: the test patch was recognized as upstream, and the implementation duplicate differed only at the migration header where main carries the corrected, non-contradictory wording. Range-diff preserves every other PR commit.

Three full-workflow findings were fixed without changing product behavior or weakening an assertion:

1. `DROP/CREATE EXTENSION pg_stat_statements` does not clear its preloaded shared state. After thousands of earlier workflow statements, the exact marker could be evicted. The feature driver now calls `pg_stat_statements_reset()` immediately after recreating the extension.
2. H-CI-3 launched three synthetic `pg_sleep(120)` clients but retained neither backend identities nor client PIDs. The later lifecycle fixture correctly observed those three plus its own sleeper. H-CI-3 now tags only its workload, terminates those tagged backends, and reaps all three clients on success and failure. The exact `active_count = 1` and decoded-row assertions are unchanged.
3. PostgreSQL 19 beta 2 resolves `text || jsonb` in two ASSERT diagnostic expressions as JSON concatenation. Both coverage objects are now explicitly cast to text. The asserted bounds and `minutes_with_data` predicates are unchanged.

## #188 absorption and assertion conservation

All original 42 #188-owned shared assertion sites remain and are reachable through the canonical runner. Their files now contain 44 sites after adding the current-main failure-counter and trigger-preservation checks.

Using the established `^\s*assert\b` source-site convention across the workflow and `devel/tests/*.sql`:

- current main before: 858 inline + 0 shared = **858**
- this PR after: 848 inline + 145 shared = **993**
- 11 existing inline assertions moved into shared SQL; the other 847 remain inline
- 1 new inline plus 134 other net-new shared sites yields +135 overall

No assertion was lost. At #188's original extraction point, its own before/after inventory was 594/594 workflow assertions.

## RED/GREEN evidence

- Ordinals: current main's dev-upgrade leaves `ash.config.num_partitions` at `attnum=11`; the new assertion exits 3. This head normalizes canonical attnums 1–22 (`num_partitions=3`) and passes preservation plus idempotent reapply checks.
- #191: on exact head `08b5f6b`, the four feature-bearing surfaces (`features`, `degraded-no-cron`, `degraded-no-pgss`, `degraded-neither`) all PASS on PG14 and PG17: **8/8**, with no `Expected FAIL: #191` remaining.
- pgss isolation: after prefilling PG17 pgss to 4,910 entries/17 deallocations, the old no-reset step executes the marker then fails at `features_readers.sql:1400`; the reset passes the unchanged enrichment assertions.
- H-CI-3 isolation: old head `041a450` fails every cron-on matrix job with four live sleepers. An isolated PG17 check of the fix removes all three tagged backends and tracked clients while leaving a differently tagged control active.
- PG19 diagnostics: old head `61a7547` fails on PG19beta2 with `invalid input syntax for type json` at the unchanged B5 assertion; exact explicit casts make the canonical PG19beta2 fresh-install surface PASS.

## Validation

On exact head `08b5f6b`:

- PG14/PG17 four-surface feature gate: 8/8 PASS; no `Expected FAIL: #191`
- PG19beta2 canonical fresh-install surface: PASS
- PG17 canonical upgrade chain: PASS
- fresh install, full dev-upgrade, and actual v1.5-to-current snapshots: byte-identical
- config row/default/owner/ACL/constraint/index/self-FK/trigger preservation and OID-stable reapply: PASS
- #188 shared assertions: 42/42 retained; final inventory 993 static sites
- `bash -n devel/scripts/release_gate.sh`: PASS
- release-gate helper unit tests: 7/7 PASS
- workflow parse, extracted Bash syntax, Python AST parse, and `git diff --check`: PASS
- independent Codex and focused cleanup reviews: no merge-blocking finding

## Final landing rebase

The merge candidate is hand-rebased onto current `main` at `9efad9c` (after #198 and #199). Exact head: `fcd37d7`.

- GitHub Actions has a 512,000-byte workflow-file limit. Combining #192 with the new #199 inline selector gate would have produced 513,005 bytes, so the unchanged 110-line/9-assert selector transaction was moved byte-for-byte to `devel/tests/stale_rollup_source.sql`; the workflow is now 508,364 bytes and invokes it in the same psql session.
- The final assertion inventory is **848 inline + 187 shared = 1,035**. All **42/42** original #188-owned shared assertions remain; those files now contain 44 sites after later additions. The #198 query-attribution gate remains 30 assertions and the #199 source-selector gate remains 9.
- Exact PG14/PG17 hard gates passed `features`, `degraded-no-cron`, `degraded-no-pgss`, and `degraded-neither`: **8/8 PASS**, with no `Expected FAIL: #191` in source or output.
- The first exact combined run correctly failed because #198 added the dynamic reader helper `ash._exact_query_uses_coarser(integer,integer,name)`, taking the direct reader-function count from 43 to 44. The final gate expects 44 and also names that helper explicitly for both the feature reader and `pg_monitor`, so an unrelated 44th grant cannot mask its loss.
- A deterministic fractional-boundary fixture exposed a real `ash.report` bug: convert-then-floor rounded `:59.9` endpoints into the following minute. The implementation now floors each bound before epoch conversion. Parent RED shifted both bounds by one minute and reported 4 expected/3 with data; final GREEN covers the exact four-minute range. This user-visible fix is recorded in `RELEASE_NOTES.md`; no assertion was weakened.
- Fresh-install and upgrade-chain release gates pass on PG17; workflow YAML, shell syntax, helper tests (7/7), and `git diff --check` pass.

This section supersedes the historical base/head and assertion totals above; those remain as the audit trail for the earlier review point.


NikolayS (approved):
Codex review standing in for REV — audit verdict: PASS on exact head 041a450. Hand rebase onto main 2ccdf65 correctly dropped the two stacked #194 commits; the only stopped conflict was their obsolete migration-header wording, for which current main is the corrected superset. Range-diff preserves the other 12 commits. Ordinal RED/GREEN reproduced (main attnum 11 and assertion exit 3; head canonical 1–22 with preservation/reapply PASS). Both schema-equivalence paths are byte-identical. The mandated #191 feature gate is 8/8 PASS across four surfaces on PG14 and PG17. All 42 #188-owned shared assertions remain; source-site inventory is 858 before and 993 after (848 inline + 145 shared), with none lost. Old CI's pgss-marker failure was independently reproduced under saturated shared state and fixed narrowly by reset-after-create; enrichment assertions are byte-identical and pass. Codex static review found no blocker. APPROVED.

NikolayS (approved):
Final exact-head verification on `fcd37d71efbe5c2b86b2374b52a07d42194db4b8`: APPROVE.

- Hand-rebased onto exact current base `9efad9c`. Range-diff maps all 15 original #192 commits patch-identically; only the documented selector extraction, report-boundary fix, and helper-grant assertions follow.
- Exact-head CI is green across PG14–PG19beta2, docs, and cron-off: https://github.com/NikolayS/pg_ash/actions/runs/30315785772
- Required hard gate is 8/8 PASS: PG14+PG17 × features/no-cron/no-pgss/neither, with zero `Expected FAIL: #191`.
- Parent RED proves convert-before-floor shifted both `:59.9` report bounds +1 minute and returned 3/4 covered minutes. Final GREEN floors before conversion, preserves all four minutes, and release-notes the behavior change; no assertion was weakened.
- The #198 dynamic reader helper made the correct direct function-grant count 44. The final gate asserts both that exact count and named EXECUTE on `_exact_query_uses_coarser(integer,integer,name)` for the explicit reader and `pg_monitor`, while admin negative controls remain.
- Assertion inventory is 848 inline + 187 shared = 1,035. All 42/42 original #188-owned shared sites execute on passing paths; those files now contain 44 sites.
- #198 is preserved (30 query-attribution assertions); #199 is preserved byte-for-byte after extraction (9 source-discriminating assertions and the exact `least(until, date_trunc(minute, now()))` clamp).
- Fresh/dev-upgrade/actual-v1.5 schemas are byte-identical; ordinal, reapply, and OID-stability checks pass.
- Workflow size is 508,364 bytes, below GitHub Actions’ 512,000-byte limit. Shell/YAML/Python syntax, helper unit tests, conflict-marker scan, and diff check pass.

No merge blocker found.

PR #194, opened 2026-07-27
release: return processed grains from rollups

### LINKED ISSUE #191: rollup_minute() and rollup_hour() return per-database row counts, not processed grains

NikolayS:
Fixes #191.

Unblocks #192.

## Contract decision

`ash.rollup_minute(integer)` and `ash.rollup_hour()` now return the number of completed time grains processed, independent of how many `(grain, datid)` rows are upserted. This matches their catalog comments, the batch-limit model, and the operator-visible meaning of catch-up progress.

Each completed loop iteration counts once, including an empty grain: it consumes one batch slot and advances its watermark. A call with an already-caught-up watermark still returns exactly `0` because it processes no loop iterations.

## Changes

- increment rollup results once per processed minute/hour rather than by DML `ROW_COUNT`
- add exact CI assertions for one and two databases, physical row counts, watermark deltas, empty grains, and already-caught-up `rollup_minute(1000000)`
- align the historical rollup blueprint and 1.5-to-2.0 upgrade comment with the corrected contract
- disclose the user-visible return correction in `RELEASE_NOTES.md`

## Call-site audit

- `ash.rotate()` stores the minute-rollup result but never reads or compares it. Exceptions and its independent `(minute, datid)` completeness query remain authoritative, so #185's truncation safety is unchanged.
- `ash.rollup_cleanup()` does not consume either return value; it only shares the rollup advisory lock and continues returning physical deletion counts.
- `ash.status()` does not call either rollup function; it reports physical table rows and watermarks directly.
- pg_cron and documented external scheduler bodies execute bare `SELECT` statements and do not branch on the scalar.
- existing tests either discard results, assert `0` on no-work/busy/incomplete paths, assert `> 0`, or assert `1` for a single grain/database. Those expectations remain valid. PR #192's retained two-database assertion changes from RED to GREEN.

## Validation

After the required hand rebase onto main `3af54a7`:

- exact issue #191 RED on main: two datids produced two physical rows and returned `2` for one minute/hour grain
- exact issue #191 GREEN on this head: two physical rows remain, both functions return `1`, and watermarks advance by 60/3600 seconds
- exact added contract block: PASS for one/two datids, empty grains, and already-caught-up `0`
- complete #185/#162 cleanup and rotation regression: PASS
- independent static caller/contract review: no blocker
- composed with #192: all four feature-bearing surfaces PASS on both PG17 and PG14 (8/8)
- workflow YAML parse and `git diff --check`: PASS

The migration header was also corrected during rebase so it no longer broadly claims all admin/lifecycle functions are unchanged while documenting this intentional return-value correction.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -5667,6 +5667,200 @@ jobs:
             raise notice 'Rollup hour different-length arrays test PASSED';
           end;
           $$;
+
+          /*
+           * #191: return processed time grains, independently of the number of
+           * per-database rows written. Exercise both the historically invisible
+           * single-database case and the failing two-database case.
+           */
+          do $rollup_return_contract$
+          declare
+            v_anchor int4 := ash.ts_from_timestamptz(
+              date_trunc('hour', statement_timestamp()) - interval '1 hour'
+            );
+            v_datids bigint;
+            v_expected_datids int;
+            v_hour_result int;
+            v_hour_rows bigint;
+            v_hour_watermark int4;
+            v_minute_result int;
+            v_minute_rows bigint;
+            v_minute_watermark int4;
+            v_wait_id smallint;
+          begin
+            select ash._register_wait(
+              'active',
+              'Issue191',
+              'ReturnContract'
+            )
+            into v_wait_id;
+
+            for v_expected_datids in 1..2 loop
+              truncate table ash.sample;
+              truncate table ash.rollup_1m, ash.rollup_1h;
+              update ash.config
+              set current_slot = 0,
+                sample_interval = interval '1 second',
+                last_rollup_1m_ts = v_anchor,
+                last_rollup_1h_ts = null
+              where singleton;
+
+              with database_ids as (
+                select database_row.oid as datid
+                from pg_catalog.pg_database as database_row
+                order by
+                  (
+                    database_row.datname = pg_catalog.current_database()
+                  ) desc,
+                  database_row.oid
+                limit v_expected_datids
+              )
+              insert into ash.sample (
+                sample_ts,
+                datid,
+                active_count,
+                data,
+                slot
+              )
+              select
+                v_anchor,
+                database_ids.datid,
+                1,
+                array[-v_wait_id::int, 1, 0]::int4[],
+                0
+              from database_ids;
+
+              select pg_catalog.count(distinct sample_row.datid)
+              into v_datids
+              from ash.sample as sample_row;
+              select ash.rollup_minute(1)
+              into v_minute_result;
+              select pg_catalog.count(*)
+              into v_minute_rows
+              from ash.rollup_1m;
+              select last_rollup_1m_ts
+              into v_minute_watermark
+              from ash.config
+              where singleton;
+
+              update ash.config
+              set last_rollup_1m_ts = v_anchor + 3600,
+                last_rollup_1h_ts = v_anchor
+              where singleton;
+              select ash.rollup_hour()
+              into v_hour_result;
+              select pg_catalog.count(*)
+              into v_hour_rows
+              from ash.rollup_1h;
+              select last_rollup_1h_ts
+              into v_hour_watermark
+              from ash.config
+              where singleton;
+
+              assert v_datids = v_expected_datids
+                and v_minute_result = 1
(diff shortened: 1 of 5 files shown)


samo-agent (approved):
Verified on exact rebased head 717545e. Issue #191 is reproducibly RED on main (one processed grain with two datids returns 2) and GREEN here (returns 1 while preserving two physical rows and 60/3600-second watermark movement). The exact added contract covers one/two datids, empty grains, and caught-up zero-work. The complete #185/#162 rotation regression passes. Independent static review found no blocker, and composing this fix with #192 makes all four feature-bearing surfaces pass on both PG17 and PG14 (8/8). Migration wording was corrected to describe the intentional return-value change without claiming all lifecycle behavior is unchanged. APPROVED.

PR #198, opened 2026-07-27
release: keep compacted query attribution honest

### LINKED ISSUE #122: Bug: 2.0 readers prefer a stale rollup over complete raw data

### LINKED ISSUE #156: Performance: combine rollup prefixes with raw tails during routine lag
## Summary

PR #123 correctly requires the `rollup_1m` watermark to reach the requested `until` boundary whenever raw samples cover the window start. A watermark at `until - interval '1 minute'` is not harmless scheduler slack: it means the latest requested complete minute has not been processed.

That strict completeness predicate should remain in #123. This follow-up should reduce the cost of routine lag by combining a complete rollup prefix with a raw tail, without accepting incomplete rollup data.

## Measured Postgres 18.3 evidence

A deterministic two-hour fixture contained:

- 7,200 raw rows, one per second;
- 100 packed active backends per raw row;
- 119 complete `rollup_1m` rows; and
- `last_rollup_1m_ts = until - interval '1 minute'`.

The strict selector chose `raw` and produced:

```text
source  buckets_with_data  backend_seconds
raw                   120        720000.00
```

`EXPLAIN (ANALYZE, BUFFERS, TIMING OFF)` reported:

```text
Execution Time: 2925.794 ms
temp read=81375 written=81462
```

At 8 KiB per block, that is approximately 636 MiB read and 636 MiB written through temporary storage.

Advancing the watermark to `until` forced the same fixture through `rollup_1m` in 2.106 ms, approximately 1,389 times faster. Because the latest rollup row was deliberately absent, the result also demonstrated the correctness failure:

```text
source      buckets_with_data  backend_seconds
rollup_1m                 119        714000.00
```

The missing minute accounted for 6,000 backend-seconds. If that minute contains the incident spike or the only activity in the window, the relative error is not bounded by one minute divided by the window length.

### Fixed time tolerance

Accepting `watermark >= until - interval '1 minute'` selects a source that is known not to cover `[until - 1 minute, until)`. It can silently omit the newest complete minute and recreate the stale-data bug fixed by #123.

### Maximum rollup timestamp

Checking `max(rollup_1m.ts)` cannot prove completeness. The minute worker advances its watermark across processed minutes with no samples, but those zero-activity minutes have no rollup row. A maximum-row check therefore cannot distinguish:

- a completely processed zero-activity tail;
- an unprocessed tail;
- coverage for one database but not another; or
- a genuine gap followed by later rows.

The transactional watermark is the authoritative processed boundary.

## Proposed design

For an eligible wide aggregate window, define the minute-aligned watermark as `W` and read two half-open ranges:

- `rollup_1m` for `[since, W)`;
- raw samples for `[W, until)`.

The seam must have neither overlap nor gap. Eligibility must require raw retention to cover `W`; otherwise preserve the existing safe fallback behavior.

## Acceptance criteria

- Keep #123's strict `last_rollup_1m_ts >= until` predicate for selecting a pure `rollup_1m` source.
- Introduce a hybrid path only when:
  - `rollup_1m` covers `since`;
  - `since < W < until`;
  - `W` is the authoritative minute-aligned watermark; and
  - raw retention covers `W`.
- Read exactly `rollup_1m [since, W)` plus raw `[W, until)`, with exact-value assertions proving no double count and no omitted seam minute.
- Handle processed zero-activity minutes from the watermark, never from `max(rollup_1m.ts)` or row presence.
- Expose an honest hybrid source label, such as `rollup_1m+raw`, and update the documented source contract rather than reporting either component as the sole source.
- Apply the design consistently to every aggregate consumer of the selector and its delegates:
  - `ash.aas()`;
  - `ash.timeline()`;
  - `ash.periods()`;
  - non-tie `ash.top()`;
  - `ash.compare()`;
  - `ash.chart()`; and
  - `ash.summary()`.
- Keep wait/event-to-query tie drills on their existing raw-only path.
- Cover both aggregate helper shapes used by those readers, including count-series and dimensional reads.
- Follow strict RED/GREEN TDD with exact source, bucket, AAS, and backend-second assertions for:
  - routine one-minute lag;
  - a high-load latest-minute spike;
  - a zero-activity seam minute;
  - `watermark = until` pure-rollup selection;
  - NULL or substantially stale watermarks;
  - raw retention starting after `W`; and
  - database-filtered and unfiltered reads.
- Record repeatable Postgres performance evidence comparing full-window raw decoding with hybrid tail decoding. Include execution time, buffers, temporary I/O, fixture size, and the number of raw rows decoded; do not make wall-clock timing alone a flaky CI assertion.
- Run fresh-install, installer reapply, relevant upgrade/schema-equivalence checks, and the supported PostgreSQL CI matrix.

## Relationship to #123

This is a performance follow-up to #123, not a reason to weaken it. Until a hybrid path proves exact coverage, #123 should retain the strict watermark completeness rule and use full raw fallback when raw covers the requested window.

NikolayS:
## Recovery / scope

Addresses #136.

This draft replaces auto-closed #190. GitHub closed that PR when its stacked base branch, `agent/w2-b1`, was deleted after #182 merged; it was not rejected and had no submitted reviews, inline threads, or issue comments.

This reworks and supersedes the approach in #150 without modifying that branch. It is now based directly on `main` at `2ccdf65c613b7f79e00e83fcd4d1d7566e60d5c6`, which contains #180, #182, #186, #189, and #194. This PR does **not** resolve or close #122. It also does not claim to resolve #156.

## Rebase recovery

Dropped as already upstream:

- The four #182 stack commits (`6f9a5af`, `1182691`, `eddda1d`, `2af229a`) and their `blueprints/SPEC.md` delta; #182 is now squash-merged as `a1f0a30`.
- The duplicate creation of `ash._raw_oldest_sample()` and the old `_raw_retention_start() := _raw_oldest_sample()` wiring. #189 already landed the physical helper plus the logical, minute-aligned ring boundary; restoring the old wiring would regress #163/#189.
- Pre-#189 boundary comments and fixture assumptions. The recovered test now uses `_raw_oldest_sample()` for physical evidence and exact-asserts a deterministic `logical < old rollup < physical` split.

The rebased series is:

- `1683b78` — test: expose compacted query attribution gaps
- `21f1bab` — fix: keep compacted query attribution honest
- `ac28efc` — docs: explain compacted query attribution
- `12f30cf` — test: bound query-attribution rollup setup
- `8660412` — test: verify rebased query-attribution interactions

### Current-main rebase

The only textual conflict was in `blueprints/AAS_USER_STORIES.md`. It was resolved by retaining #186's cadence/coverage limitation and #180's until-only/inverted-window contract while adding this PR's exact-query/raw and compacted-residual rules.

#194 changed `rollup_minute()` / `rollup_hour()` to return processed time grains rather than physical database rows. The #136 minute fixture now calls `rollup_minute(1)`, and both residual fixtures separately assert their processed-grain count and their single physical row.

## What changed

- Rollup-backed, unfiltered `top('query_id')` emits a positive `NULL` residual equal to `max(total wait counts - preserved named query counts, 0)` per source row. The residual enters the ranking stream before `n`, so compacted-away load can win a top-N slot.
- The residual scans each row's `query_counts` once with `GROUPING SETS` and its `wait_counts` once. It avoids #150's second query-array scan and window-wide materialized expansion.
- Every explicit `query_id` filter uses raw samples. It raises only when ordinary planning selects a matching, database-scoped rollup with retained coverage in the requested slice before the physical oldest raw sample.
- A fallback rollup name alone is not treated as data-loss proof. Young installs remain usable when their first minute/hour rollup covers only the same retained minute as raw.
- Valid `rollup_1h.minute_counts` must contain a non-NULL pre-raw minute before the hourly envelope counts as lost exact-query history; legacy/incomplete detail remains conservative.
- `compare()` NULL-safely pairs rollup residuals, and `summary()` renders them as `(other / unattributed)`.

## Combined rule on the landed #182/#189 base

| Reader | Exact query behavior | Rollup query residual | Grain contract |
|---|---|---|---|
| `aas` / `timeline` | Any explicit `query_id` forces raw; matching pre-raw rollup history raises, while a young same-minute rollup does not. | None | #182 effective bounds and honest extrema remain authoritative. |
| `top` | Explicit `query_id`, or query breakdown plus a wait filter, forces raw. | Unfiltered rollup query breakdowns emit the residual before ranking and `n`. | Hourly query/wait arrays remain hour-grain with NULL sub-hour extrema. |
| `compare` | Exact query filters force both sides raw. | Each rollup-backed query side emits one residual; the full join pairs NULL keys exactly once. | Inherits each side's disclosed source/effective grain. |

The known `_pick_source_agg()` end-watermark/completeness gap remains #122 and is intentionally not claimed fixed.

### Rollup minute residual competes for top-N

RED recorded by #190:

```text
ERROR:  residual top-N mismatch: key=111 source=rollup_1m seconds=3.00 pct=100.00
CONTEXT:  PL/pgSQL function inline_code_block line 98 at ASSERT
```

GREEN on the recovered branch:

```text
NOTICE:  issue #136 rollup_1m residual GREEN: named=3.00/11.54 residual=23.00/88.46 top1=NULL compare=NULL/0.00
DO
```

The block also exact-asserts two rows, `26.00` backend-seconds, `100.00%`, one NULL-safe compare residual with delta `0.00`, and the exact summary label.

### `rollup_1h` top-100 compaction

RED on current `main` (`2ccdf65`):

```text
ERROR:  rollup_1h top-100 mismatch: rows=100 total=300.00 pct=100.00 residual=/ named=3.00/1.00 source= bucket= peak= p99=
CONTEXT:  PL/pgSQL function inline_code_block line 126 at ASSERT
```

GREEN:

```text
NOTICE:  issue #136 rollup_1h top-100 GREEN: rows=101 total=400.00 pct=100.00 residual=100.00/25.00
DO
```

The fixture stores 100 named rows / 300 seconds after hourly top-set truncation while wait totals remain 400 seconds. The result is 101 rows / 400 seconds / 100%, with a 100-second residual and NULL peak/p99 at hour grain.

### Young-install exact query filters

Current-main RED is a source-honesty failure, not an exception:

```text
ERROR:  young-install aas mismatch: source=rollup_1m buckets=1 seconds=7.00
CONTEXT:  PL/pgSQL function inline_code_block line 143 at ASSERT

ERROR:  same-minute hourly query mismatch: error=<none> source=rollup_1h seconds=7.00
CONTEXT:  PL/pgSQL function inline_code_block line 166 at ASSERT
```

GREEN:

```text
NOTICE:  issue #136 young-install GREEN: same-minute 1m/1h rollups ignored; aas=raw/7.00 until-only=raw/7.00 inverted=validated timeline=raw/0.12 top=raw/7.00 compare=raw/raw; other-db allowed; old rollup rejected
DO
```

Exact assertions cover `aas` (`raw`, one bucket, 7 seconds), `timeline` (60 rows, one data row, 0.12 AAS), `top` (7 seconds, 100%), `compare` (raw/raw, delta 0, 100/100), database scoping, and matching old-history rejection.

The strengthened fixture retains the young raw row and asserts the ordering between the logical ring boundary, old rollup window, and physical oldest sample. A mutation that passed `_raw_retention_start()` instead of `_raw_oldest_sample()` failed as intended:

```text
ERROR:  ash.aas query_id retention error mismatch: ash.aas query_id filter should require raw
mutant_exit=3
```

## Residual cost measurement retained from #190

PostgreSQL 17.9, 43,200 `rollup_1m` rows, 100 query pairs and 32 wait pairs per row:

| Variant | Median | Relative |
|---|---:|---:|
| #182 base, no residual | 27.425866s | baseline |
| Straight #150 residual / second query scan | 34.828003s | 27.0% slower than base |
| Final per-row grouped stream | 4.159544s | 88.1% faster than naive; 84.8% faster than base |

Naive #150 spilled about 20 GiB of temp reads/writes; the final stream spilled about 79.5 MiB. Naive and final outputs were identical: 4,363,200 rows and total count 29,160,000. This is adjacent to #156's cost class but does not resolve it.

## Validation

- PostgreSQL 17.9: helper-discovered fresh install, installer reapply, and full 1.0-to-current upgrade chain — GREEN.
- PostgreSQL 18.4 disposable container: all three exact #136 blocks — GREEN; post-step sample partitions/view all had zero rows.
- Current-main RED reproduced for the young-install compacted-source regression and hourly top-100 residual.
- #180 until-only/inverted-window/cross-horizon regression block — GREEN.
- #182 minute-grain seam and rollup_1h grain-honesty/partial-hour blocks — GREEN.
- #194 processed-grain return-contract block — GREEN; #136 setup separately asserts processed grains and physical rows.
- Exact-query until-only re-anchoring and inverted-window-before-source-validation assertions — GREEN.
- Mutated logical-boundary implementation — expected RED, proving the strengthened fixture is discriminating.
- Installer idempotent reapply, workflow YAML parse, docs guards, and `git diff --check` — GREEN.
- Independent SQL semantic review and independent docs/test review — no findings after the fixture-strengthening cleanup.

CI remains responsible for the full PostgreSQL 14-19 matrix. This PR stays draft.

## Landing verification (current main)

- Hand-rebased onto current main 2ccdf65, including #180, #186, and #194.
- The only textual conflict was blueprints/AAS_USER_STORIES.md. The resolution retains #186 historical-cadence/#137 limitations alongside this PRs exact-query/raw and rollup-residual rules.
- #194 deliberately changed `rollup_minute()` to count processed time grains. The #136 fixture now bounds its setup with `batch_limit = 1`; both residual fixtures separately assert processed-grain counts and physical rows. `RELEASE_NOTES.md` already records #194's return-contract change.
- Exact head `8660412` preserves #180 until-only/inversion, #182 grain honesty, #189 physical/logical retention boundaries, #194 one-grain/two-datid plus empty/caught-up behavior, and #181's dynamic helper grant closure; it also directly asserts until-only and inverted exact-query behavior.
- #122 remains explicitly unclaimed.


samo-agent (approved):
Codex review standing in for REV — audit verdict: PASS on exact head 12f30cfb0c1c161a4e107e33c80105a6753f672c.

Hand rebase: based directly on main 2ccdf65. The only conflict was blueprints/AAS_USER_STORIES.md; I retained both #186s historical-cadence/#137 limitation and #198s exact-query/raw plus NULL-residual rules.

RED/GREEN: current main is discriminating RED after the #194 setup adaptation: named compacted query load is 3.00 seconds at 100.00% instead of 11.54%. Exact head is GREEN: minute named/residual 3.00/11.54 and 23.00/88.46 (the NULL residual wins top-1); hourly top-100 returns 101 rows/400.00 seconds/100%, with NULL residual 100.00/25.00; young same-minute rollups stay raw while genuinely older coarser history is rejected.

#194 assertion adaptation: before, an unbounded rollup_minute() setup asserted return=1 and described one database row. After #194, that function intentionally counts every processed completed minute, so the 20-minute-old fixture correctly returned 20. The setup now calls rollup_minute(1), still asserts 1, and says one completed minute. No #136 attribution predicate changed. RELEASE_NOTES.md already records #194s return-contract behavior.

Cross-checks: #180 until-only/inversion/cross-horizon PASS; #182 grain honesty/partial-hour PASS; #194 one/two-datid, empty-grain, caught-up, row-count, and watermark contracts PASS. #189 physical/logical raw boundaries remain; #181s dynamic non-admin reader bundle covers the new helper. #122 remains explicitly unclaimed. Workflow YAML and diff checks pass. No merge-blocking finding.

samo-agent (approved):
Codex review standing in for REV — head correction / final delta: PASS on exact head 8660412a2c2a9523ce07733ac5b817b355d5208a.

The prior review was submitted after this head had already advanced, so its GitHub commit_id is correctly 8660412 although its opening sentence named the previously verified 12f30cf. The only new commit is test-only: it separately asserts #194s processed grain and physical row count, and adds until-only plus inverted-window coverage to the young exact-query fixture. The full canonical #136 step passes on exact 8660412. No product assertion was weakened and the PASS verdict stands.

samo-agent (approved):
Codex review standing in for REV — final CI-trigger head: PASS on exact head 5ef411f14133b1c0335108dfcbb61158acd11477. This final commit is deliberately empty: its tree is byte-identical to reviewed head 8660412 (tree 47e7e3b296727cbf10043b9362bc53d61b4c5137) and exists only because draft-to-ready did not trigger the pull_request workflow. All prior RED/GREEN and cross-check conclusions therefore apply unchanged; the newly triggered full PR workflow must still pass before merge.

NikolayS:
Temporarily closing and reopening to trigger the required pull_request CI event; the draft-to-ready and fork synchronize transitions created no workflow run.

samo-agent (approved):
Codex review standing in for REV — final verdict: PASS on exact head e349eddd357d958523f0a2dac2a84a463fdb7019. Full PR CI is green across docs lint, PG14/15/16/17/18/19beta1, and the PG17 cron-off lane: https://github.com/NikolayS/pg_ash/actions/runs/30314259819. The final CI repair is assertion-neutral: GitHub rejected the prior 535,600-byte workflow above its 512,000-byte ceiling before creating any check suite. The 749-line #136 SQL payload is byte-identical after extraction (22,068 bytes, 30 assertions, SHA-256 26cdc4e29aa87e4de9664f26a73de1c40bd1b3a82d831e889ce3059eb0b08cb7), the psql connection and ON_ERROR_STOP behavior are unchanged, and test.yml is now 506,353 bytes. Prior discriminating RED/GREEN and #180/#182/#189/#194 cross-checks stand unchanged. No merge-blocking finding.

PR #199, opened 2026-07-27
release: keep stale rollups honest without losing fast path

### LINKED ISSUE #122: Bug: 2.0 readers prefer a stale rollup over complete raw data

NikolayS:
## Summary

Relates to #122. Supersedes and reworks #123; this deliberately does **not** close #122.

#123 correctly prevented a lagging `rollup_1m` from hiding newer raw load, but compared its watermark with the unclamped `until`. A standard dashboard boundary of `date_trunc('minute', now()) + interval '1 minute'` is unreachable by that watermark, forcing a healthy wide read onto raw data.

This rework:

- delegates canonical fallback selection to `_pick_source(since)`;
- when that fallback is `raw`, requires the rollup watermark to reach `least(until, date_trunc('minute', now()))`;
- otherwise preserves the established picker result instead of duplicating B3's physical/logical raw-boundary logic.

## Measured performance

Five runs per implementation on the same local PostgreSQL 17.9 fixture: 21,600 one-second raw rows over six hours, 40 active backends per row, 6,096 kB raw table, and 360 healthy minute-rollup rows. Each call used the rounded-up dashboard boundary.

| Implementation / chosen source | Runs (ms) | Median |
| --- | --- | ---: |
| #123 (`6123780`), `raw` | 1494.746, 1345.164, 1386.175, 1364.884, 1351.251 | **1364.884 ms** |
| This branch (`100e95c`), `rollup_1m` | 7.521, 2.193, 1.980, 1.961, 1.914 | **1.980 ms** |

The rejected #123 path was **689.3× slower** on this reproduction. Both paths returned 360 covered buckets and `864000.00` backend-seconds; the discriminating difference was the selected source.

## Exact regression coverage

`Test 2.0: stale and rounded-up rollup source selection` runs in an isolated `BEGIN`/`ROLLBACK` fixture and asserts:

- stalled rollup with complete raw coverage: `source = 'raw'`, `buckets_with_data = 2`, `backend_seconds = 2.00`;
- healthy historical rollup: `source = 'rollup_1m'`, `buckets_with_data = 90`, `backend_seconds = 90.00`;
- healthy rollup with `until` rounded up one minute: `source = 'rollup_1m'`, `buckets_with_data = 120`, `backend_seconds = 120.00`.

The literal final test block is RED on untouched #123 head `6123780`:

```text
BEGIN
ERROR:  healthy rollup must keep the fast path for a rounded-up until, got raw
CONTEXT:  PL/pgSQL function inline_code_block line 89 at ASSERT
```

It is GREEN on this rebased branch:

```text
BEGIN
NOTICE:  stale/rounded-up rollup source selection PASSED
DO
ROLLBACK
```

## Explicit non-goal

A stalled rollup for a window beginning before physical raw coverage can still return a silently truncated rollup answer. The post-rebase probe remains:

```text
NOTICE:  older-than-raw stalled rollup: source=rollup_1m, buckets=1, backend_seconds=1.00
```

Completeness disclosure for that case is intentionally out of scope here and remains tracked by #122, which stays open.

## Post-rebase validation

Rebased immediately before push onto `origin/main` `3af54a7e552ad3f29225a1f7278138de4570ac78`, including #189, #182, #197, #180, and the later #186 documentation update.

- Fresh install and idempotent installer reapply: passed on PostgreSQL 17.9.
- Exact selector fixture: passed with the GREEN output above.
- Full #180 `until`-only/inverted-window fixture: passed (`B5 until-only reader windows PASSED`, `B5 inverted-window validation PASSED`, and `B5 #63 cross-horizon clamp PASSED`).
- `git diff --check`, workflow YAML parsing, and changed shell-block syntax: passed.
- Independent final diff review after the last rebase: no actionable blockers.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -7081,6 +7081,126 @@ jobs:
         run: |
           psql -h localhost -U postgres -d postgres -v ON_ERROR_STOP=1 -f devel/tests/query_attribution_assert.sql
 
+      - name: "Test 2.0: stale and rounded-up rollup source selection"
+        env:
+          PGPASSWORD: postgres
+        run: |
+          psql -h localhost -U postgres -d postgres -v ON_ERROR_STOP=1 << 'EOF'
+          -- A wide aggregate read may prefer rollup_1m only when that rollup
+          -- covers the requested end. If raw covers the whole window while the
+          -- rollup worker lags, choosing rollup silently drops the recent load.
+          -- Conversely, a healthy rollup must keep the fast path when a
+          -- dashboard rounds until up to the next minute boundary.
+          do $$
+          declare
+            v_wait smallint;
+            v_start timestamptz := date_trunc('minute', now()) - interval '2 hours';
+            v_end timestamptz := date_trunc('minute', now());
+            v_start_ts int4;
+            v_recent_ts int4;
+            v_config record;
+            a record;
+          begin
+            select
+              current_slot,
+              num_partitions,
+              sample_interval,
+              rotation_period,
+              rotated_at,
+              last_rollup_1m_ts,
+              last_rollup_1h_ts
+            into v_config
+            from ash.config
+            where singleton;
+
+            truncate ash.sample_0, ash.sample_1, ash.sample_2;
+            truncate ash.rollup_1m, ash.rollup_1h;
+            update ash.config
+            set current_slot = 0,
+                num_partitions = 3,
+                sample_interval = interval '1 second',
+                rotation_period = interval '1 day',
+                rotated_at = v_end,
+                last_rollup_1m_ts = ash.ts_from_timestamptz(
+                  v_start + interval '1 minute'
+                ),
+                last_rollup_1h_ts = null
+            where singleton;
+
+            select ash._register_wait('active', 'IO', 'StaleRollupProof')
+            into v_wait;
+            v_start_ts := ash.ts_from_timestamptz(v_start);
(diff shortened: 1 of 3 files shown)


samo-agent (approved):
Codex review standing in for REV — audit verdict: PASS on exact rebased head de169f46533d54ad312694bb760b28f1998df73c. Rebased by hand onto main 6d72583. Workflow conflict: retained #198s extracted query-attribution gate first, then #199s isolated stale/rounded-up selector gate. Installer comment conflicts: preserved #198s exact-query terminology while adding #199s watermark/fallback contract and explicit #122 non-goal. Product predicate is exact: last_rollup_1m_ts reaches least(until, date_trunc('minute', now())) only when canonical fallback is raw. The gate directly asserts source=raw for stale coverage and source=rollup_1m for both historical and rounded-up healthy windows; it has no timing assertion. Current main RED is "stale rollup must fall back to raw, got rollup_1m". Exact head GREEN is "stale/rounded-up rollup source selection PASSED". Focused PG17 #180 and #198 cross-checks pass; #182/#189/#194 bodies and #198s helper are unchanged. RELEASE_NOTES records the behavior and leaves older-than-physical-raw completeness under #122. Full PG14-19beta1 plus cron-off CI is green: https://github.com/NikolayS/pg_ash/actions/runs/30314863835. No merge-blocking finding.

PR #204, opened 2026-07-28
fix: make lifecycle job teardown atomic

### LINKED ISSUE #203: ash.stop() suppresses rollup unschedule failures, leaving jobs after rebuild/uninstall
## Summary

On release-candidate SHA `d1971fd4b984f88375c4e1fc0c110488483d809b`, `ash.stop()` suppresses every error raised while unscheduling the three rollup jobs. Both `ash.rebuild_partitions()` and `ash.uninstall()` call `ash.stop()` and therefore can report success while an active pg_cron job remains.

This was found while generalizing #202. It is not #202's installer-commit/post-work mechanism: each lifecycle call is one PostgreSQL statement and ordinary uncaught failures roll the statement back. The inconsistency comes from `EXCEPTION WHEN OTHERS THEN NULL` in `ash.stop()`.

## Affected paths

- `sql/ash-install.sql:2493-2522`: errors from `cron.unschedule('ash_rollup_1m')`, `ash_rollup_1h`, and `ash_rollup_gc` are silently discarded.
- `sql/ash-install.sql:1754-1757`: `ash.rebuild_partitions()` calls `ash.stop()` before rebuilding.
- `sql/ash-install.sql:2580-2585`: `ash.uninstall()` calls `ash.stop()` before dropping the schema.

A lock timeout on an existing `cron.job` row is enough to trigger the suppressed error:

- `ash.rebuild_partitions(4, 'yes')` returns its normal success string, leaves `sampling_enabled=false`, but leaves `ash_rollup_1m` active.
- `ash.uninstall('yes')` reports that four jobs were removed and drops the `ash` schema, but leaves `ash_rollup_1m` active with command `select ash.rollup_minute()`, which now targets a missing function.

## Copy-pasteable reproduction

Run from the repository at the SHA above. This uses the release-gate Dockerfile to provide PostgreSQL 17 with pg_cron.

```bash
docker build \
  --build-arg=POSTGRES_TAG=17 \
  --tag=pg-ash-release-gate:17 \
  devel/docker

set -Eeuo pipefail

test "$(git rev-parse HEAD)" = d1971fd4b984f88375c4e1fc0c110488483d809b
repo_root=$PWD
image=pg-ash-release-gate:17
container_name="pgash-stop-repro-${BASHPID}"
container_id=
locker_shell_pid=
locker_backend_pid=

cleanup() {
  if [[ -n "${locker_backend_pid}" && -n "${pg_port:-}" ]]; then
    PGPASSWORD=pgash_stop_repro psql --no-psqlrc \
      --host=127.0.0.1 --port="${pg_port}" --username=postgres \
      --dbname=postgres --tuples-only --no-align \
      --command="select pg_terminate_backend(${locker_backend_pid});" \
      >/dev/null 2>&1 || true
  fi
  if [[ -n "${locker_shell_pid}" ]]; then
    wait "${locker_shell_pid}" 2>/dev/null || true
  fi
  if [[ -n "${container_id}" ]]; then
    docker rm --force "${container_id}" >/dev/null 2>&1 || true
  fi
}
trap cleanup EXIT INT TERM

container_id=$(docker create \
  --name="${container_name}" \
  --label=org.pg-ash.stop-repro=true \
  --env=POSTGRES_DB=postgres \
  --env=POSTGRES_PASSWORD=pgash_stop_repro \
  --publish=127.0.0.1::5432 \
  "${image}" \
  postgres \
  -c shared_preload_libraries=pg_cron \
  -c cron.database_name=postgres \
  -c cron.use_background_workers=on \
  -c cron.launch_active_jobs=off)
docker start "${container_id}" >/dev/null
port_mapping=$(docker port "${container_id}" 5432/tcp)
pg_port="${port_mapping##*:}"

export PGPASSWORD=pgash_stop_repro
export PAGER=cat
export PGOPTIONS='-c client_min_messages=warning'
psql_base=(
  psql --no-psqlrc
  --host=127.0.0.1
  --port="${pg_port}"
  --username=postgres
  --dbname=postgres
  --set=ON_ERROR_STOP=1
)

for attempt in $(seq 1 60); do
  if pg_isready --host=127.0.0.1 --port="${pg_port}" \
      --username=postgres --dbname=postgres >/dev/null 2>&1; then
    break
  fi
  if [[ "${attempt}" == 60 ]]; then
    echo 'PostgreSQL did not become ready' >&2
    exit 1
  fi
  sleep 1
done

"${psql_base[@]}" --quiet --command='create extension pg_cron;'
"${psql_base[@]}" --quiet --file="${repo_root}/sql/ash-install.sql" >/dev/null
"${psql_base[@]}" --quiet \
  --command="select * from ash.start(interval '5 seconds');" >/dev/null

start_locker() {
  local app_name=$1
  local attempt

  PGAPPNAME="${app_name}" "${psql_base[@]}" --quiet --command="
    begin;
    select jobname
    from cron.job
    where jobname = 'ash_rollup_1m'
    for update;
    select pg_sleep(300);
  " >/dev/null 2>&1 &
  locker_shell_pid=$!

  locker_backend_pid=
  for attempt in $(seq 1 100); do
    locker_backend_pid=$("${psql_base[@]}" --tuples-only --no-align --command="
      select pid
      from pg_stat_activity
      where application_name = '${app_name}'
        and wait_event_type = 'Timeout'
        and wait_event = 'PgSleep';
    ")
    locker_backend_pid=$(printf '%s' "${locker_backend_pid}" | tr -d '[:space:]')
    if [[ "${locker_backend_pid}" =~ ^[0-9]+$ ]]; then
      return
    fi
    sleep 0.1
  done

  echo "lock holder ${app_name} did not reach pg_sleep" >&2
  exit 1
}

stop_locker() {
  "${psql_base[@]}" --tuples-only --no-align \
    --command="select pg_terminate_backend(${locker_backend_pid});" >/dev/null
  wait "${locker_shell_pid}" 2>/dev/null || true
  locker_shell_pid=
  locker_backend_pid=
}

echo '=== rebuild_partitions ==='
start_locker pgash_rebuild_lock_holder
"${psql_base[@]}" --tuples-only --no-align --field-separator='|' <<'SQL'
set lock_timeout = '500ms';
reset lock_timeout;
select metric, value
from ash.status()
where metric in ('version', 'sampling_enabled', 'cron_job_ash_rollup_1m')
order by metric;
select jobname, active, command
from cron.job
where jobname like 'ash_%'
order by jobname;
SQL
stop_locker

# Recreate the four jobs that stop() removed. start() also replaces the
# surviving rollup job, leaving five jobs for the uninstall reproduction.
"${psql_base[@]}" --quiet \
  --command="select * from ash.start(interval '5 seconds');" >/dev/null

echo '=== uninstall ==='
start_locker pgash_uninstall_lock_holder
"${psql_base[@]}" --tuples-only --no-align --field-separator='|' <<'SQL'
set lock_timeout = '500ms';
select ash.uninstall('yes');
reset lock_timeout;
select 'ash_schema', coalesce(to_regnamespace('ash')::text, '<absent>');
select jobname, active, command
from cron.job
where jobname like 'ash_%'
order by jobname;
SQL
stop_locker

"${psql_base[@]}" --tuples-only --no-align \
  --command="select current_setting('server_version');"
```

## Actual output

```text
=== rebuild_partitions ===
SET
rebuilt: 3 -> 4 partitions. all raw data cleared. call ash.start() to resume sampling.
RESET
cron_job_ash_rollup_1m|id=3, schedule=* * * * *, active=t
sampling_enabled|false
version|2.0
ash_rollup_1m|t|select ash.rollup_minute()
=== uninstall ===
SET
uninstalled: removed 4 pg_cron jobs, dropped ash schema
RESET
ash_schema|<absent>
ash_rollup_1m|t|select ash.rollup_minute()
17.10 (Debian 17.10-1.pgdg13+1)
```

## Expected

If an existing pg_cron job cannot be unscheduled, the failure must not be silently converted into lifecycle success. The error should be surfaced so the enclosing `rebuild_partitions()` or `uninstall()` statement can roll back, or the caller must receive an explicit failure result that accurately identifies the surviving job.

A genuinely absent job may remain idempotent, but unrelated errors such as lock timeouts and permission failures must not be treated as absence.

NikolayS:
Fixes #203.

## Contract decision

Lifecycle functions now **raise and roll back** when they cannot complete managed pg_cron work. A warning or partial-success row would still be unsafe because `rebuild_partitions()` and `uninstall()` consume `ash.stop()` internally; an uncaught error gives all four entry points one honest, statement-atomic contract.

- `ash.stop()` pre-resolves genuine absence, unschedules exact job IDs, requires `cron.unschedule()` to return true, and emits a `removed` row only after proven deletion (with the real job ID).
- Ordinary unschedule errors preserve their original SQLSTATE and name the job/jobid. Cancellation still propagates naturally.
- `ash.rebuild_partitions()` and `ash.uninstall()` inherit that error, so config changes, earlier unschedules, DDL/data changes, and schema removal roll back together.
- `ash.start()` applies the same exact-ID/true-result rule to rollup replacement. Existing sampler/rotation jobs have schedule, command, target database, and active state repaired.
- Lifecycle calls must run as the `ash` schema owner. This closes the cross-superuser variant where role B bypasses owner-only ACLs, misses role A's jobs, and drops the schema. Visible foreign-owned managed jobs targeting the installation database make teardown fail closed with the job and owner named.
- Genuinely absent jobs remain idempotent.

The catalog comments, README, release notes, and the rollup blueprint now describe this contract; the blueprint no longer prescribes `EXCEPTION WHEN OTHERS THEN NULL`.

## RED

Captured before the fix with the issue's locked-row reproduction:

```text
ERROR:  issue #203: lifecycle functions reported success while ash_rollup_1m could not be unscheduled: ash.stop(), ash.rebuild_partitions(), ash.uninstall()
```

Artifact: `/tmp/pgash-203-red-TunI9x/results.tsv` (log line 2298 in `/tmp/pgash-203-red-TunI9x/17_fresh-install.log`).

## GREEN

The exact deterministic workflow assertion passed on every supported surface major, including the row-lock rollback, real-ID/idempotency checks, inactive/drifted-job repair, cross-superuser rejection, and foreign-owned-job fail-closed probe:

| PostgreSQL | #203 assertion |
|---|---|
| 14 | PASS |
| 15 | PASS |
| 16 | PASS |
| 17 | PASS |
| 18 | PASS |
| 19beta2 | PASS |

Complete PG17 release gate:

| Surface | Result |
|---|---|
| fresh-install | PASS |
| upgrade-chain | PASS |
| features | PASS |
| degraded-no-cron | PASS |
| degraded-no-pgss | PASS |
| degraded-neither | PASS |
| cron-path | PASS |

Machine-readable TSV: `/tmp/pgash-203-owner-all-D4Izt0/results.tsv`.

Also green: `test_release_stamp.py` (17), `test_ci_step_script.py` (7), extracted-step `bash -n`, `git diff --check`, and overlay discovery (`2.0-beta1`, `devel/sql/ash-install.sql`).

## Further lifecycle-result sites found

These remain intentionally deferred; they are separate operational-result or frozen-migration contracts, but they share the same family of misleading SQL success and should be tracked:

1. `sql/migrations/ash-1.1-to-1.2.sql:21-30` catches every sampler-command update error and does nothing. A lock/permission failure can leave the old command while the migration continues.
2. `sql/migrations/ash-1.2-to-1.3.sql:2612-2630` repeats that pattern around `cron.alter_job()`. Both are frozen legacy migration legs, so this PR does not rewrite them casually.
3. `devel/sql/ash-install.sql:1590-1652,1680-1682`: `ash.rotate()` returns `failed: ...` text for rollup contention/failure/incompleteness and truncate lock timeout, so pg_cron records SQL success (the known B2 result-contract family).
4. `devel/sql/ash-install.sql:3541-3546,3561-3566`: `ash.rollup_minute()` returns `0` both when caught up and when rollup/sampler locks are busy; callers cannot distinguish no work from no progress.
5. `devel/sql/ash-install.sql:3905-3910`: `ash.rollup_cleanup()` returns a skipped string on lock contention; a direct caller sees it, but pg_cron still records SQL success.
6. Without pg_cron, `ash.stop()` returns an external-scheduler reminder, but `ash.uninstall()` consumes that row and drops the schema. It truthfully counts zero pg_cron removals, but the reminder is lost; external scheduler ownership needs its own policy.

## Landing coordination

This branch is rebased on current `origin/main` (`87c22a3`) and uses the post-#179 development-overlay policy: released `sql/ash-install.sql` and frozen migrations are unchanged; the complete next payload is `devel/sql/ash-install.sql`.

PR #201 also adds that complete overlay and already contains the same overlay-aware config-ordinal chain split, so landing #201 first will create an add/add overlap. Rebase this PR after #201 and apply the lifecycle delta on top of #201's copy-reviewed development installer; keep #201's wording where unrelated.

PR #200 should then promote the composed #201 + #203 development payload while retaining #200/#202's transaction-local installer wrapper and #200's release identity/RELEASE_NOTES organization. Port the #203 lifecycle note deliberately rather than letting either full-file version overwrite the other.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -8680,7 +8680,8 @@ jobs:
               ), 'B3 ash.start() cron result was not successful';
               assert v_sampler_status in (
                 'created',
-                'already exists — schedule updated to 1 seconds'
+                'already exists — schedule updated to 1 seconds, '
+                  'command repaired, activated'
               ), 'B3 ash.start() sampler status mismatch: '
                 || v_sampler_status;
             else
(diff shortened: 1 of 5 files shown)


NikolayS:

## samorev Code Review Report

- **PR:** NikolayS/pg_ash#204 - fix: make lifecycle job teardown atomic
- **Author:** @NikolayS
- **AI-Assisted:** Yes

| Pipeline | Coverage |
|----------|----------|
| none | Not reported |

---

### BLOCKING ISSUES (3)

**MEDIUM** [tests] The #203 regression test is gated to `matrix.postgres == 17 && matrix.cron == 'on'` only, while the PR description claims the assertion was verified on 14–19. CI will not catch a regression in pg_cron/lock-handling behavior on other majors (the `cron.unschedule(bigint)` true/false contract and `cron.job` column set vary across pg_cron versions bundled per major).
> `if: matrix.postgres == 17 && matrix.cron == 'on'`
> **Fix:** Gate on `matrix.cron == 'on'` only (all majors), or document why PG17-only is sufficient.

**MEDIUM** [docs] Binding lifecycle calls to the `ash` schema owner is a behavioral break: operators who today run `ash.stop()`/`ash.uninstall()`/`ash.rebuild_partitions()` as a different superuser will now get a hard error. RELEASE_NOTES lists this inside a "fix" bullet rather than calling it out as a breaking/upgrade-action item, and the README does not state what error a non-owner sees or that `ash.start()` is affected too.
> `Lifecycle calls are bound to the `ash` schema owner, and visible foreign-owned managed jobs ... make teardown fail closed. (issue #203)`
> **Fix:** Add an explicit "Breaking / upgrade note" entry naming the four functions and the `SET ROLE` workaround; mention the error text so it is searchable.

---

### Summary

| CI/Pipeline | 0 | 0 | 0 |
| Security | 0 | 0 | 0 |
| Bugs | 0 | 0 | 0 |
| Tests | 0 | 1 | 0 |
| Guidelines | 0 | 0 | 0 |
| Docs | 0 | 1 | 0 |
| Metadata | 1 | 0 | 0 |

Note:

<details>

```text
provider=github
kind=pr
project=NikolayS/pg_ash
number=204
target=github:NikolayS/pg_ash#204
state=OPEN
draft=true
diff_lines=8817
diff_added=8708
diff_removed=22
diff_bytes=340699
comments_count=0
commits_count=1
ci_status=none
blocking=true
samorev_review=1
review_round=1/10
escalated=false
posted_by=local
no_comment=true
live_posting=not-run
defender_dropped=2
```

**Contested / dropped**

- [docs] The blueprint snippet silently changes the hourly rollup schedule from `'0 * * * *'` to `'1 * * * *'`. This is unrelated to the atomicity contract and, if it does not match what `ash.start()` actually schedules, the blueprint now disagrees with the installer. — The installer in this same diff schedules `ash_rollup_1h` at `'1 * * * *'` (with an explicit minute-1 rationale and a migration block), so the blueprint now matches `devel/sql/ash-install.sql` rather than diverging from it.
- [tests] The foreign-owned-job fail-closed check relies on the caller being able to *see* other roles' rows in `cron.job`. pg_cron's RLS policy on `cron.job` restricts non-superusers to `username = current_user`, so when the `ash` owner is not a superuser the check is a no-op and the orphan scenario in #203 is not actually closed. The PR acknowledges "visible" jobs but the visible test fixture only exercises the superuser case. — The PR's code comments, README, and release notes already explicitly scope the foreign-job guard to "visible" jobs and explain the non-superuser limitation, and the owner-bound gate on ash.start() prevents a non-superuser owner from ever creating the orphan state, so this is a documented, intentional limitation rather than an unhandled gap.

</details>

---

---

## Reviewer notes

Reviewed at head `fc810ef7`. I reproduced the bug and verified the fix myself against real PostgreSQL 18.4 with pg_cron 1.6 in Docker, rather than relying on the transcript in the description — the `/tmp/pgash-203-*` artifact paths it cites are gone, so nothing in the PR body is independently checkable as written.

**The bug is real, it is on `main` today, and it is worse than "reports success".**

### Red — reproduced on current `main` (083e9c1, `sql/ash-install.sql`)

Five managed jobs scheduled, then another session holds a row lock on `cron.job` for `ash_rollup_1m`, and `ash.stop()` runs with `lock_timeout = '2s'`:

```text
 job_type  | job_id | status
-----------+--------+---------
 sampler   |     11 | removed
 rotation  |     12 | removed
 rollup_1h |        | removed
 rollup_gc |        | removed
(4 rows)
```

No error. No warning. Four rows claiming `removed`, and `ash_rollup_1m` is simply missing from the output — still scheduled, still firing. Note also that `rollup_1h` and `rollup_gc` report `job_id` as NULL, so the report cannot even name what it claims to have deleted.

Then the part that makes this a release blocker. `ash.uninstall()` consumes `ash.stop()` internally, so under the same lock:

```text
select ash.uninstall('yes');
NOTICE:  drop cascades to 79 other objects
-- succeeded

select count(*) from pg_namespace where nspname = 'ash';   -- 0

 jobid |    jobname    |          command
-------+---------------+----------------------------
    13 | ash_rollup_1m | select ash.rollup_minute()
```

The schema is gone and a pg_cron job is left behind calling `ash.rollup_minute()` forever against a schema that no longer exists — erroring every minute into `cron.job_run_details`, on a database the operator believes they have cleanly uninstalled from. `uninstall` reported success.

### Green — verified on this PR's `devel/sql/ash-install.sql`

Same fixture, same lock:

```text
ERROR:  ash.stop: could not unschedule pg_cron job "ash_rollup_1m" (jobid 18):
        canceling statement due to lock timeout
```

and, critically, the state afterwards:

```text
-- schema ash still exists: 1
 jobid |    jobname
-------+---------------
    16 | ash_sampler
    17 | ash_rotation
    18 | ash_rollup_1m
    19 | ash_rollup_1h
    20 | ash_rollup_gc
```

All five jobs survive. `sampler` and `rotation` had already been unscheduled before the loop reached `rollup_1m`, and they are correctly rolled back — that is the statement-atomicity claim, and it holds. `uninstall()` under the same conditions now fails closed and leaves the installation intact and recoverable, instead of half-destroyed.

Also verified: the happy path reports real job IDs for all five job types (the old code reported NULL for the three rollup jobs), and a second `ash.stop()` on already-stopped state returns zero rows without erroring, so idempotency is preserved.

### The fix itself

The core change is the right one and it is worth naming plainly: the old code wrapped each rollup unschedule in

```sql
exception when others then
  null;
```

which is the anti-pattern that produced every symptom above. The replacement resolves absence **before** calling `cron.unschedule()`, unschedules by exact `jobid` rather than by name, requires the return value to be `true`, and only then emits a `removed` row. That ordering is what makes idempotency and failure distinguishable, which a `when others` handler fundamentally cannot do given pg_cron raises the same generic error for both.

Re-raising with `errcode = sqlstate` preserves the original SQLSTATE, so `lock_timeout` surfaces as `55P03`-class behavior and a cancellation still propagates with its own code rather than being flattened into `P0001`. I checked that specifically because the description claims it.

Removing `exception when others then null` from the blueprint prose is the right call too — it should not be prescribed anywhere.

### Test quality

Strong. The CI block at `test.yml:9487` builds a real lock-holder fixture with a background session, asserts its own preconditions before testing anything (five active jobs present, the minute-rollup command as expected, and that the lock holder actually reached `pg_sleep` before proceeding), and then exercises `stop()`, `rebuild_partitions()` and `uninstall()` through a dedicated `P0203` SQLSTATE, plus absent-job idempotency, drifted/inactive-job repair, and the cross-superuser rejection. These are behavioral assertions with exact expectations, not catalog checks.

### Findings

1. **The #203 regression test is gated to one major.** `test.yml:9488` is `if: matrix.postgres == 17 && matrix.cron == 'on'`, but the description presents a PG 14/15/16/17/18/19beta2 PASS table for this exact assertion. CI reproduces one cell of that table. This matters more than usual here because the contract under test is pg_cron's — the `cron.unschedule(bigint)` true/false return and the `cron.job` column set vary by bundled pg_cron version across majors, which is precisely what could regress silently. Gate on `matrix.cron == 'on'` across all majors, or state in the workflow why PG17 alone is sufficient. (Two neighbouring steps use the same gate, so there may be a deliberate cost reason — but then the description should not claim six-major verification.)

2. **Owner-binding is a breaking change presented as a fix.** Lifecycle calls now require the `ash` schema owner, so an operator who today runs `ash.stop()` / `ash.uninstall()` / `ash.rebuild_partitions()` as a *different* superuser gets a hard error after upgrading. `RELEASE_NOTES.md` folds this into a fix bullet; it needs its own breaking/upgrade-action entry naming all four affected functions (`ash.start()` is affected too), quoting the error text so operators can search for it, and giving the `SET ROLE` workaround. This is a real behavior change for the managed-Postgres audience the project targets, and it should not be discovered at upgrade time.

3. **No CI has run on this PR at all** — zero check runs, zero check suites, zero workflow runs on the head SHA. This head is on a fork whose workflow runs are waiting on maintainer approval that was never given. Everything above is my local verification standing in for a gate that never executed. The same is true of #205 and #201.

I did check the merge result rather than assuming: `pr204` merges into `origin/main` cleanly, and I extracted and ran the merged **Guard docs against stale privilege and reader names** step against the merged tree — it exits 0. No docs-lint breakage from this PR.

### Landing

The description's own sequencing is right and should be treated as a constraint, not a suggestion: `devel/sql/ash-install.sql` does not exist on `main` (only `devel/sql/README.md` does), and both this PR and #201 add that whole file, so they are a guaranteed add/add conflict. Whichever lands second must be rebuilt on top of the first rather than resolved file-wise, or one of them will silently lose its half.

### Verdict

The fix is correct, the bug it fixes is live on `main` and genuinely severe, and the tests are honest. Not mergeable yet:

- draft, and gated behind the #201 landing-order decision;
- no CI has ever run on it;
- finding 1 (single-major gating vs. a six-major claim) and finding 2 (undocumented breaking change) should be addressed first.

I could not push these fixes to the branch myself; it is on a fork.

NikolayS:

The atomic lifecycle behavior is being reconciled in #251 with the newer CALL wrappers, custom cron commands, recovery guards and cross-database ownership checks. Samorev found additional permission cases; those are being fixed and tested before integration.

Issue #202, opened 2026-07-28
1.5-to-2.0 migration can stamp 2.0 before config normalization fails


NikolayS:
## Release blocker

A failed `1.5 -> 2.0` migration can publish the `2.0` release identity before `ash.config` has physically converged. The installer currently stamps and commits first; config ordinal normalization then runs in a second transaction.

Origin: #192. Found while adversarially reviewing draft release-stamp PR #200.

## Exact reproduction

Reproduced on PostgreSQL 17.10 from the immutable released payload (do **not** use the development migration chain, whose historical shim can replay the current installer):

1. Extract the released SQL with `git archive v1.5 -- sql`.
2. Apply, in order, the archived `sql/ash-1.0.sql`, `ash-1.0-to-1.1.sql`, `ash-1.1-to-1.2.sql`, `ash-1.2-to-1.3.sql`, `ash-1.3-to-1.4.sql`, and `ash-1.4-to-1.5.sql`.
3. Confirm the released precondition is `1.5|1.5|11`: the `ash.config.version` row, `ash.status()` version metric, and `num_partitions` physical `attnum`, respectively.
4. Add an inert custom trigger that the migration promises to reject safely:

```sql
create function public.ash_config_atomicity_trigger()
returns trigger language plpgsql as $$
begin
  return null;
end
$$;

create trigger ash_config_atomicity_custom
after update on ash.config
for each row
execute function public.ash_config_atomicity_trigger();
```

5. In a separate `psql --set=ON_ERROR_STOP=1` process, run the current public wrapper `sql/ash-1.5-to-2.0.sql`.
6. From a fresh connection, inspect the same two release identities and physical ordinal.

Observed before the fix:

```text
migration_exit=3
ERROR: cannot normalize ash.config: unsupported custom triggers are present
after_version=2.0
after_status=2.0
after_num_partitions_attnum=11
```

A fresh 2.0 install has `num_partitions` at `attnum=3`, so this database claims 2.0 while physically unconverged.

## Late-failure discriminator

A preflight for the custom-trigger condition alone is insufficient. A `public` view depending on `ash.config` passes the migration's explicit catalog checks, follows the old relation through its rename, and then makes the later `DROP TABLE ash.config_ordinal_legacy RESTRICT` fail. Before the fix, this path also leaves `2.0|2.0|11`.

[...]

PR #205, opened 2026-07-28
fix: bind installer include mode to its migration transaction

samo-agent:
## Draft stacked #202 fix — do not merge

This isolates the follow-up delta on the held `release/2.0` branch. It does **not** authorize merging #200, #204, or this PR, and it does not authorize tagging. Before the final release candidate, this fix must be folded into the candidate and the release-stamp commit recreated last so payload identity remains unambiguous.

Relates to #202.

## Red reproduction

The previous held head `ba96f7858c0827b8afaf031cd3f832de37db256a` trusted the literal value of `pg_ash.install_in_migration_transaction`. Supplying it before connection startup bypassed both installer transaction boundaries:

```text
Issue #202 external-marker v1.5 precondition: 1.5|1.5|11
Issue #202 external-marker v1.5 install_exit=3
Issue #202 external-marker v1.5 after failure: 2.0|2.0|11
ERROR: forced late installer failure under externally preset marker
```

The new regression supplies `PGOPTIONS=-c pg_ash.install_in_migration_transaction=on` before direct installer invocation. It covers both the immutable v1.5 payload and an empty database, forces a late failure at `ash.summary(...)`, and compares identity, config data, config relation OID, schema, and survival of the blocker from a fresh connection.

## Fix and rationale

A finite preflight cannot cover event triggers, dependency changes, concurrent DDL, vanished roles, permissions, cancellation, deadlock/timeout, storage/resource failures, or commit failure. The robust boundary remains one transaction around installer replay plus normalization.

The migration now stores its assigned transaction ID in the transaction-local internal setting. The installer accepts include mode only when that token matches `pg_current_xact_id_if_assigned()`; an ambient literal such as `on` has no matching assigned transaction and therefore fails closed into direct-installer transaction ownership. The installer computes that decision once and reuses it at the footer, so later SQL or an event trigger cannot change ownership by mutating the setting. The migration retains one final `COMMIT`; direct invocation retains its documented final `COMMIT`.

## Green verification

Exact candidate: `38af9f2f9b462027a484bd42ae0c273ecfe32a28`

```text
PostgreSQL 14 full #202 suite: PASS
PostgreSQL 17 full #202 suite: PASS
PostgreSQL 19beta2 full #202 suite: PASS

external-marker v1.5: install_exit=3, after=1.5|1.5|11, rollback PASS
external-marker fresh: install_exit=3, after=<absent>|<absent>|<absent>, rollback PASS
custom-trigger: migration_exit=3, after=1.5|1.5|11, rollback PASS
dependent-view: migration_exit=3, after=1.5|1.5|11, rollback PASS
same-client transaction-marker rollback: PASS

PostgreSQL 17 release gate:
  fresh-install PASS
  upgrade-chain PASS
  features PASS
  degraded-no-cron PASS
  degraded-no-pgss PASS
  degraded-neither PASS
  cron-path PASS

17/17 test_release_stamp.py: PASS
7/7 test_ci_step_script.py: PASS
check_release_stamp.py --tag v2.0: PASS
git diff --check: PASS
```

No merge or tag has been performed.

### DIFF
--- a/devel/tests/config_normalization_atomicity.sh
+++ b/devel/tests/config_normalization_atomicity.sh
@@ -11,8 +11,11 @@ IFS=$'\n\t'
 readonly REQUESTED_CASE="${1:-all}"
 if [[ "${REQUESTED_CASE}" != "all" \
   && "${REQUESTED_CASE}" != "custom-trigger" \
-  && "${REQUESTED_CASE}" != "dependent-view" ]]; then
-  printf 'usage: %s [all|custom-trigger|dependent-view]\n' "$0" >&2
+  && "${REQUESTED_CASE}" != "dependent-view" \
+  && "${REQUESTED_CASE}" != "external-marker" ]]; then
+  printf \
+    'usage: %s [all|custom-trigger|dependent-view|external-marker]\n' \
+    "$0" >&2
   exit 2
 fi
 
@@ -65,6 +68,8 @@ cleanup_database() {
   "${PSQL[@]}" --quiet >/dev/null 2>&1 <<'SQL' || true
 drop event trigger if exists fail_ash_install_transaction_marker;
 drop function if exists public.fail_ash_install_transaction_marker();
+drop event trigger if exists fail_ash_install_external_marker;
+drop function if exists public.fail_ash_install_external_marker();
 do $cleanup$
 begin
   if pg_catalog.to_regprocedure('ash.uninstall(text)') is not null then
@@ -102,6 +107,15 @@ reset_test_database() {
       create database ${TEST_DATABASE}
       with template ${SEED_DATABASE};
     "
+}
+
+reset_empty_test_database() {
+  "${PSQL_MAINTENANCE[@]}" \
+    --quiet \
+    --command="drop database if exists ${TEST_DATABASE} with (force);"
+  "${PSQL_MAINTENANCE[@]}" \
+    --quiet \
+    --command="create database ${TEST_DATABASE} template template0;"
 }
 
 cleanup() {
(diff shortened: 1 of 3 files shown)


NikolayS:

## samorev Code Review Report

- **PR:** NikolayS/pg_ash#205 - fix: bind installer include mode to its migration transaction
- **Author:** @samo-agent
- **AI-Assisted:** Yes

| Pipeline | Coverage |
|----------|----------|
| none | Not reported |

---

### BLOCKING ISSUES (2)

**LOW** [docs] The new include-mode check relies on `pg_current_xact_id()` / `pg_current_xact_id_if_assigned()`, which only exist on PostgreSQL 13+ (pre-13 uses `txid_current()`). The diff raises the minimum server version for both the installer and the migration but neither file's header comment nor the PR notes state this; verification covers 14/17/19 only.
> `) = pg_catalog.pg_current_xact_id_if_assigned()::text then 'true'` and `pg_catalog.pg_current_xact_id()::text` in the migration.
> **Fix:** Document the PG 13+ requirement in the installer/migration header comments (or confirm the documented minimum is already ≥13), and if older servers must remain supported, fall back to `txid_current()`/`txid_current_if_assigned()`.

---

### Summary

| CI/Pipeline | 0 | 0 | 0 |
| Security | 0 | 0 | 0 |
| Bugs | 0 | 0 | 0 |
| Tests | 0 | 0 | 0 |
| Guidelines | 0 | 0 | 0 |
| Docs | 0 | 1 | 0 |
| Metadata | 1 | 0 | 0 |

Note:

<details>

```text
provider=github
kind=pr
project=NikolayS/pg_ash
number=205
target=github:NikolayS/pg_ash#205
state=OPEN
draft=true
diff_lines=459
diff_added=317
diff_removed=31
diff_bytes=15848
comments_count=0
commits_count=2
ci_status=none
blocking=true
samorev_review=1
review_round=1/10
escalated=false
posted_by=local
no_comment=true
live_posting=not-run
defender_dropped=2
```

**Contested / dropped**

- [bugs] The psql variable `pg_ash_install_in_migration_transaction` is no longer `\unset` after use; it now persists in the caller's psql session after `\i`/`\ir` of the installer. The old code unset it right after the header and recomputed at the footer; the new code intentionally caches it through the footer but never cleans it up afterwards, leaving an internal variable visible in interactive sessions and to any subsequent script in the same session. — The variable is unconditionally recomputed via `\gset` at every header before any `\if` reads it, so the leaked psql variable can never influence a later include or any behavior; it is a well-namespaced (`pg_ash_`-prefixed), purely cosmetic residue in the caller's session — a non-actionable nitpick, not a bug, and the retention through the footer is the fix's deliberate defense against mid-install mutation.
- [tests] The new `external-marker` cases only cover the negative path (ambient `on` is rejected). There is no positive test that an ambient setting equal to a *different* assigned xid string (e.g. `PGOPTIONS=-c pg_ash.install_in_migration_transaction=12345`) is also rejected, nor that a stale/garbage value (non-numeric) fails closed rather than erroring at the `\gset` comparison. Both exercise the same code path but a numeric mismatch is the closer adversarial case to the real token. — The rejection is value-independent — at the installer's `\gset` no xid is assigned, so `pg_current_xact_id_if_assigned()` is NULL and any ambient literal (`on`, numeric, garbage) fails the same comparison identically; the existing test already exercises the sole code path, making extra parameterized values a non-actionable LOW nitpick.

</details>

---

---

## Reviewer notes

Reviewed at head `38af9f2f9b462027a484bd42ae0c273ecfe32a28`. I re-derived the red reproduction and the green result myself against real PostgreSQL 18.4 in Docker rather than taking the PR description on trust.

### Both gate findings are real, and neither is about the code

The gate raises two blockers: `draft=true`, and `ci_status=none` (`total=0`). Both are accurate, and the CI one is the more interesting of the two.

**No CI has ever run on this PR** — zero check runs, zero check suites, zero workflow runs on the head SHA. Two independent causes:

1. This head lives on a fork, and fork workflow runs on this repository are waiting on maintainer approval that has not been given. No suites are created until then, which is why even `demo.yml` (which has unfiltered `push`/`pull_request` triggers) shows nothing.
2. More structurally: `test.yml` is gated on `pull_request: branches: [main]`, and this PR targets `release/2.0`. **Even with fork approval, the test matrix would not run on it.** Every PR into the held `release/2.0` stack has the same blind spot.

That second point is worth acting on independently of this PR. A change to `sql/ash-install.sql` — the installer's transaction-ownership decision, no less — currently reaches `main` through a branch that the test matrix structurally cannot see. The coverage is deferred rather than absent (`test.yml` on `release/2.0` does invoke `devel/tests/config_normalization_atomicity.sh`, so it will run once the stack lands on `main`), but "we will find out after the merge" is the wrong side of the gate for installer atomicity.

### Red reproduction — confirmed independently

I built the pre-fix tree (`c8acc209`, the PR's own red commit) and ran the new case on PostgreSQL 18.4:

```text
Issue #202 external-marker v1.5 precondition: 1.5|1.5|11
Issue #202 external-marker v1.5 install_exit=3
Issue #202 external-marker v1.5 after failure: 2.0|2.0|11
ERROR:  forced late installer failure under externally preset marker
external marker published a partial v1.5 upgrade: before=1.5|1.5|11 after=2.0|2.0|11
```

The bug is exactly as described and it is not theoretical: an ambient `PGOPTIONS=-c pg_ash.install_in_migration_transaction=on` suppresses the installer's own `BEGIN`/`COMMIT`, so a failure partway through leaves the database **stamped 2.0 with a half-applied install**. The version says the upgrade succeeded; the schema says otherwise.

### Green — confirmed independently

Same tree, fix applied, full suite:

```text
Issue #202 custom-trigger rollback PASSED
Issue #202 dependent-view rollback PASSED
Issue #202 same-client transaction-marker rollback PASSED
Issue #202 external-marker v1.5 rollback PASSED          (after failure: 1.5|1.5|11)
Issue #202 external-marker fresh rollback PASSED         (after failure: <absent>|<absent>|<absent>)
Issue #202 failed-normalization release-state atomicity PASSED
```

And the happy paths, which the atomicity suite does not cover on its own:

```text
v1.5 chain seeded              -> 1.5
sql/migrations/ash-1.5-to-2.0  -> exit 0, ash.config.version = 2.0
sql/ash-install.sql on empty   -> exit 0, ash.config.version = 2.0
```

Worth stating why that matters: the `custom-trigger` and `dependent-view` cases only roll back correctly **if include mode still engages** for the real migration. Their passing is positive evidence that the token handshake did not simply disable include mode altogether, which is the obvious way a fix like this could be quietly wrong.

### The fix itself

The mechanism is sound. The migration stores `pg_current_xact_id()::text` in the transaction-local setting; the installer accepts include mode only when the stored token equals `pg_current_xact_id_if_assigned()::text`. Walking the cases:

- Setting absent — `current_setting(..., true)` returns NULL, `nullif` keeps NULL, comparison is NULL, `CASE` falls to `'false'`. Installer owns its transaction.
- Setting is `''` — `nullif` yields NULL, same outcome.
- Setting is an ambient literal such as `on` — no assigned xid at connection start means `pg_current_xact_id_if_assigned()` is NULL and the comparison is NULL; even with an xid assigned, `'on'` never equals a numeric xid8. Fails closed, which is the correct direction.
- Migration path — `pg_current_xact_id()` assigns the xid and stores it; the installer reads back the same value. Match.

Two details I checked specifically:

- **`is_local => true`** on `set_config` is correct, so the token cannot outlive a rolled-back migration.
- **Caching through the footer.** Dropping the `\unset` and deleting the footer's second `\gset` is the other half of the fix, not incidental cleanup: recomputing at the footer would let later SQL or an event trigger flip transaction ownership mid-install. The header `\gset` always overwrites, so a stale psql variable from an earlier `\i` in the same session cannot leak in.
- **Version floor.** `pg_current_xact_id()` and `pg_current_xact_id_if_assigned()` are PostgreSQL 13+; the supported floor is 14. Fine.
- **No regression to the documented "install inside my own transaction" flow.** With the marker unset that path is unchanged — `BEGIN` warns, the final `COMMIT` commits the caller's outer work, exactly as the header comment has always said.

### Test quality

Genuinely strong, and I want to be specific about why, since "asserts a row exists" would not have caught this. The new case asserts the exact version triple (`1.5|1.5|11`), the exact installer exit code (`3`), an md5 of the `ash.config` row, the `ash.config` relation OID (so a drop-and-recreate cannot pass as untouched), a full schema snapshot diff, and survival of the blocking event trigger from a fresh connection. It covers both the immutable v1.5 payload and an empty database. It is registered in the `all` runner and `test.yml` on this base invokes the script with no argument, so it is wired in rather than orphaned.

The commit split is the right shape too: `c8acc209` adds the failing test, `38af9f2f` fixes it.

### Verdict

The code and the tests pass on the merits. What is unmet is process, not quality:

- **Blocking:** no CI, for the two structural reasons above.
- **Blocking:** draft, and this PR's own description asks that it not be merged — it targets the held `release/2.0` stack and says the fix must be folded into the release candidate with the stamp commit recreated last.

So this is not a "merge it" recommendation. It is a "the fix is correct, verified, and should be carried into the release candidate" recommendation, plus a request that the `pull_request: branches: [main]` blind spot over `release/2.0` be closed on its own.

NikolayS:

Transaction-bound installer ownership is being incorporated with the complete migration fix and the new sample_unlogged config shape. Temporary promotion tests cover the canonical migration and public wrapper against actual historical tags. No final release stamp is authorized.

Issue #215, opened 2026-08-03
Bug: revoke_reader() bricks the installation for a non-superuser owner; grant_reader() cannot repair it


NikolayS:
Found by the pre-tag audit (#160). Affects the **non-superuser owner** case — i.e. the RDS / Cloud SQL / Supabase / Neon audience pg_ash targets. A superuser owner is immune via ACL bypass, which is why this went unnoticed.

## Repro

As a non-superuser owner of schema `ash`:

```sql
select ash.revoke_reader('<owner>');   -- returns cleanly, no warning
```

Then everything breaks:

```
ash.take_sample()  -> ERROR: permission denied for schema ash
ash.status()       -> ERROR: permission denied for schema ash
every reader       -> ERROR: permission denied for schema ash
ash.grant_reader() -> ERROR: permission denied for schema ash
```

Re-granting schema `USAGE` by hand still leaves `grant_reader()` failing:

```
ERROR: permission denied for function _admin_funcs
```

`ash._admin_funcs()` is not in the admin exclusion list, so `revoke_reader` strips EXECUTE on the very helper `grant_reader` initialises from.

## Impact

Silent at call time. Downs sampling and the entire reader surface for that database. **Both documented repair paths are broken by the same call** — recovery needs hand-written `GRANT EXECUTE ON ALL FUNCTIONS` + `GRANT SELECT ON ALL TABLES`, or a full installer re-apply.

Location: `:6195-6263`, specifically `:6256 revoke usage on schema ash from %s`.

## Fix

One `if` block: refuse when the target owns schema `ash`, is `current_user`, or is a superuser. Add `ash._admin_funcs()` to the admin exclusion list.

NikolayS:
## Verification against current `main`

Tested on `main` at `083e9c1` (`ash.config.version` = `2.0-beta1`), fresh install from `sql/ash-install.sql`, real PostgreSQL 18.4 in Docker. Posting evidence only — no action taken on this item.

**Result: does not reproduce. The exact guard this issue asks for is present.**

```sql
select ash.revoke_reader(current_user);
```
```text
ERROR:  ash.revoke_reader: refusing protected role postgres
        (schema owner, current user, or superuser)
CONTEXT:  PL/pgSQL function revoke_reader(name) line 39 at RAISE
```

Same result when the owner is named indirectly rather than via `current_user`:

```sql
select ash.revoke_reader(
  (select pg_get_userbyid(nspowner) from pg_namespace where nspname='ash')
);
[...]

Issue #210, opened 2026-08-03
Release blocker: rollup_1h grain collapse — top()/compare() emit hour averages in peak_aas/p99_aas, fabricating regressions


NikolayS:
Found by the pre-tag audit (#160). **Not covered by any in-flight PR** — re-tested against PR #131's head (`1aa6bf3`) and the behaviour is byte-identical to 2.0-beta1.

Four audit lenses converged on this independently. **It is one structural gap with one fix, not four bugs** — hence one issue with four sub-parts.

## The gap

`ash.aas()` (`:3897`) and `ash.timeline()` (`:4044`) upgrade a `rollup_1h` read to the minute-grain `rollup_1h_minutes` view so `peak_aas`/`p99_aas` stay per-minute. `ash.top()` never does: `:4470` leaves `v_grain_secs = 3600` and silently snaps the caller's `bucket => '1 minute'` up to an hour. `ash.compare()` delegates *both* windows to `ash.top()`.

### B1a — `top()` contradicts `aas()`/`summary()` by 13×

Seed one `rollup_1h` hour = 600 counts in minute 0 + 60 in each other minute:

```
ash.aas()                 -> peak_aas 10.00
ash.top('wait_event')     -> peak_aas  0.77     -- same window
```

`ash.summary()` prints `peak_aas 10.00` directly above per-wait rows whose max peak is `0.77`. `order_by => 'peak'` — advertised in the shipped catalog comment as *the* incident-triage ordering — provably cannot surface a one-minute spike on any `rollup_1h`-backed window.

### B1b — any filter silently degrades the grain

The `rollup_1h_minutes` upgrade at `:3897` is gated on `wait_event_type is null and wait_event is null and query_id is null`:

```
ash.aas(...)                          -> peak_aas 21.00   source rollup_1h
ash.aas(..., wait_event_type => 'IO') -> peak_aas  2.30   source rollup_1h
```

`ash.timeline()` NULLs p99 in exactly this case, with the comment *"a percentile over hour averages would masquerade as a minute percentile"* — `aas` and `top` violate the invariant the file writes down two functions away.

### B1c — `compare()` fabricates regressions

Two windows, byte-identical load, window 1 past the 30-day `rollup_1m` retention:

```
key  | avg_aas_1 | avg_aas_2 | avg_delta | peak_aas_1 | peak_aas_2 | p99_aas_1 | p99_aas_2
[...]

PR #131, opened 2026-07-10
fix: reject partial hourly drills

NikolayS:
## What changed

- preserves exact unfiltered/database-only partial-hour `aas` and `timeline` reads through `rollup_1h.minute_counts`
- requires hour-aligned bounds when a reader must use hourly wait/query arrays
- rejects explicitly requested sub-hour `timeline` and `chart` buckets on `rollup_1h`
- widens omitted buckets to the safe one-hour grain automatically
- documents the retained-precision boundary in README and catalog comments

## Why

An hourly dimensional row represents a full hour. The readers previously selected whole rows by their start timestamp and divided their complete counts by a clipped partial window. A single-backend half-hour IO read could therefore report AAS 2.0, while a cross-hour read could omit the overlapping first hour.

The current storage format cannot reconstruct partial-hour wait/query dimensions exactly. Rejecting unsupported shapes is safer than fabricating precision.

## Impact

The data API now raises an actionable error for unsupported partial-hour dimensional reads. Unfiltered minute-total reads remain exact and unchanged.

## Validation

Postgres 18.3:

- RED at `5da0e13`: filtered half-hour `aas` returned instead of rejecting
- GREEN at `1aa6bf3`:
  - unfiltered half-hour remains `avg_aas = 1.00`, `backend_seconds = 1800`
  - filtered `aas`, `top`, dimensional `compare`, explicit sub-hour `timeline`, and explicit sub-hour `chart` reject
  - aligned filtered hour remains exact at `avg_aas = 1.00`, `backend_seconds = 3600`
  - omitted timeline/chart buckets auto-select two one-hour rows
- existing 2.0 rollup-source reader tests pass
- existing rollup_1h minute-seam tests pass
- existing calendar-bucket and spike-first ordering tests pass
- installer re-apply with `ON_ERROR_STOP=1` passes
- `git diff --check main...HEAD` passes

Closes #130

PR #216, opened 2026-08-03
fix: revoke_reader owner guard (#215), until-only window anchoring (#214), take_sample slot race

### LINKED ISSUE #214: Bug: passing only until silently answers a completely different window

## Repro

Load exists at `now() - 1h`. Ask for a window ending three hours ago:

```sql
select * from ash.top('wait_event', until => now() - interval '3 hours');
```
```
key              | avg_aas | pct
IO:DataFileRead  |   40.00 | 100.00
```

That is real data from a window **three hours away from the one requested**, returned with no window columns and no NOTICE. `ash.summary()` goes from "no data in this time range" to a full top-waits profile from the wrong hour. `ash.aas(until => <3h ago>)` returned `avg_aas 40.00 / backend_seconds 2400` for `22:13–22:14` while the intended window had `4.75`.

## Cause

`since` and `until` defaults are computed independently (`:3832-3833` and siblings), and `since > until` is never validated. The #63 degenerate-window guard — whose stated purpose is INT4 overflow safety — then rewrites the inverted window to `[now()-1h, now()-1h+1min)`.

Affected: `:3856` (aas), `:3996` (timeline), `:4451` (top), `:5105` (report), `:5492` (chart).

## Why it matters

`ash.periods()` anchors every window to `until` and gets this right, which sets exactly the opposite expectation for the rest of the family. `ash.samples()` returns 0 rows in the same situation — the readers also disagree with each other.

"Show me load up to \<past incident time\>" is a natural named-arg call the README teaches (`until => '2026-02-14 03:05'`), and it returns fabricated attribution during post-incident review.

## Fix

- Raise on `since > until`
- Default `since` to `until - interval '1 hour'` when only `until` is given
- Keep the #63 clamp for its actual purpose, applied **after** validation

NikolayS:
Fixes #214 and #215, plus the `take_sample()` slot race and an ACL-convergence gap the first two
uncovered. All four are independent of the `rollup_1h` grain cluster (#210 / #131 / #150), so this
can land without waiting on that decision.

Found by the pre-tag audit — see #160.

**Draft on purpose.** Per `CLAUDE.md` this needs REV and your explicit approval. It has neither.

## What's here

| Commit | Fix |
|---|---|
| `01c54ce` | **#215** — `revoke_reader()` bricked a non-superuser-owned install. It now refuses when the target owns schema `ash`, is `current_user`, or is a superuser. `ash._admin_funcs()` moved into the admin set so `revoke_reader` stops stripping the helper `grant_reader()` initialises from. |
| `f1d9d7e` | **#214** — passing only `until` silently answered a window an arbitrary distance away. `until` now resolves first, `since` defaults to `until` minus the reader's span, and `since > until` raises instead of being rewritten by the #63 clamp. |
| `724a0fa` | **slot race** — `take_sample()` resolved query-map ids against the slot it read at the top but omitted `slot` from the INSERT, so the column DEFAULT re-evaluated under a fresh snapshot. A rotation landing mid-sample stamped old-dictionary ids with the new slot: NULL attribution, then *unrelated* query ids once the new slot reused map ids 1,2,3. Passed validation, never touched `insert_errors` — undetectable after the fact. |
| `037896d` | Aligns the shipped catalog comments and `blueprints/AAS_API.md` with the #214 rule. |
| `04233f9` | **ACL convergence** — because `_admin_funcs()` became admin-only, an EXECUTE grant handed out by 2.0 beta 1 could never be withdrawn (`CREATE OR REPLACE` preserves ACLs; the corrected `revoke_reader` skips admin functions). A beta1-upgraded install kept privileges a fresh install does not grant, `pg_monitor` included. Hardening now revokes from named non-owner grantees, not just PUBLIC. |
| `1b1fac0` | Release notes for the above. |

## Verification

Every fix was proven **RED against unpatched SQL first**, then GREEN — and independently
re-derived by a second reviewer working from a clean checkout of the same base commit, not from
the author's evidence.

- Full CI replay: **51/51 on the unpatched baseline, 54/54 on this branch**, across PG 14, 18 and 19beta1.
- A 60-case behavioural battery diffed unpatched-vs-patched: the *only* differences are the six
  until-only cases and nine inverted-window cases that now raise. Everything else byte-identical —
  all ten zero-arg readers, `periods()` still anchoring on `until`, the #63 clamp still applying.
- Fresh install, full 1.0→2.0 chain, double re-apply, and fresh-vs-chain schema equivalence
  (288 catalog rows) all clean.
- `04233f9` additionally asserts a **second re-apply is a fixed point**, which is the property
  that actually matters for convergence.

The zero-argument regression guard was explicitly checked for tautology: run alone against
unpatched SQL it *passes*, while the inverted-window block still fails. That is the PR #131
failure mode being guarded rather than reproduced.

## Review notes

- **`ash.report`'s "never raises" contract** was reworded rather than exempted:
  *"the only raise is the malformed request `since > until`, so a null always means no data and
  never a failed call."* That keeps the guarantee machine consumers depend on (null ⇒ no data)
  while being honest about argument validation. Worth a second opinion.
- **Behaviour change worth your attention:** `04233f9` means an admin-function grant an operator
  gave *directly* to a role is now removed on every installer apply. The supported path becomes
  membership in the schema-owner role. Called out explicitly in the release notes.
- Interacts with nothing else in the open backlog — no shared hunks with #123/#125/#129/#131/#150.


NikolayS:

**#215 — `revoke_reader` owner guard.** `main` already refuses:
```text
select ash.revoke_reader(current_user);
ERROR:  ash.revoke_reader: refusing protected role postgres
        (schema owner, current user, or superuser)
```
Same result when the owner is named indirectly via `pg_get_userbyid(nspowner)`.

**#214 — until-only window anchoring.** With server time ~05:25 UTC:
```text
select * from ash.aas(until => now() - interval '3 hours');
 period_start           | period_end
------------------------+------------------------
 2026-08-21 01:25:00+00 | 2026-08-21 02:25:00+00
```
`since` correctly defaults to `until - 1 hour`, and an explicitly inverted window now raises `ash.aas: since must be less than or equal to until` instead of being rewritten by the #63 clamp. `ash.top(until => ...)` returns 0 rows rather than fabricated attribution.

The `take_sample()` slot race is covered by the same merged #181.

Full reproduction transcript: [#214](https://github.com/NikolayS/pg_ash/issues/214#issuecomment-5365756888) and [#215](https://github.com/NikolayS/pg_ash/issues/215#issuecomment-5365757022).

This branch's merge base is `37a0e2a` ("docs: polish user-facing text for clarity and consistency", #158) — **27 commits behind `main`**. Sitting between that base and `main` is `28136cf` ("release: reject inverted reader windows and clarify diagnostics", #180) and `8540219` ("release: fix reader privilege blockers and sample slot race", #181), merged **2026-07-27**.

Issues #214, #215 were filed **2026-08-03**, a week after that merge, and this PR was opened the same day from the pre-fix base. The same pattern holds for #216, #217, #218, #219 and #220 as a group: all five share the `37a0e2a` base, and the fixes they propose all merged on 2026-07-27 in #180, #181, #182, #185 and #189. That is consistent with the audit having been run against the stale tree rather than against `main`.

### Recommendation

NikolayS:

PR #217, opened 2026-08-03
fix: prevent rollup cleanup from permanently freezing rotation (#211)

### LINKED ISSUE #211: Release blocker: rollup_cleanup() and rotate() deadlock permanently; ash.sample grows unbounded while pg_cron reports SUCCESS
Found by the pre-tag audit (#160). **Two of pg_ash's own scheduled jobs deadlock each other.**

## Mechanism

`rollup_cleanup()` (`:3258`) deletes the `ash.rollup_1m` rows that `rotate()`'s pre-truncation completeness gate (`:1428`, the #81 guard) requires. `rollup_minute()`'s watermark (`:2943`) only walks forward, so the deleted minutes can never be re-rolled.

## Repro

Default 3 partitions / 1-day rotation, only `rollup_1m_retention_days` lowered to 1 (the shipped CHECK allows `>= 1`). Seed a 36 h-old minute, run the three shipped cron jobs in their scheduled order:

```
failed: pre-truncation rollup incomplete for 1 group(s) in slot 2; slot not truncated
```

…forever. `select ash.rollup_minute(1000000)` returns `0` — the watermark is already past those minutes.

**Also reachable with stock 30-day rollup retention** via README's documented "increase raw retention" runbook: `ash.rebuild_partitions(32,'yes')` puts endangered rows at 30–31 days old at truncation time. Truncate-target sequence confirmed: `2 3 4 … 31 0 1 2 3 4 5`.

## Impact

Rotation never happens again. Every new sample piles into the frozen current slot until the volume fills — the exact failure the ring exists to prevent.

**It fails silently.** `rotate()` returns a plain text value on the failure path, so pg_cron records **SUCCESS** and `ash.status()` exposes no failure metric; only `time_since_rotation` quietly grows. Recovery requires hand-rewinding `ash.config.last_rollup_1m_ts` or a destructive rebuild.

## Fix

- Exclude raw groups already past `rollup_1m` retention from the gate, and/or let `rotate()` rewind `last_rollup_1m_ts` to the endangered slot's `min(sample_ts)` before calling `rollup_minute()`
- Validate `(num_partitions - 1) * rotation_period <= rollup_1m_retention_days` at `rebuild_partitions()` / retention-change time
- Surface consecutive rotate failures in `ash.status()` — a silent permanent failure is the worst part of this

NikolayS:
Fixes #211 — two of pg_ash's own scheduled jobs deadlock each other, permanently and silently.

Found by the pre-tag audit — see #160. **Draft on purpose**: needs REV and your explicit approval.

## The bug

`rollup_cleanup()` deletes the `ash.rollup_1m` rows that `rotate()`'s pre-truncation completeness
gate (#81) requires, and `rollup_minute()`'s watermark only walks forward — so those minutes can
never be re-rolled and the gate can never be satisfied again.

    failed: pre-truncation rollup incomplete for 1 group(s) in slot 2; slot not truncated

…forever. Rotation stops, so every new sample piles into the frozen current slot until the volume
fills — exactly the failure the ring exists to prevent. Reachable on **stock 30-day retention** via
the README's own "increase raw retention" runbook (`rebuild_partitions(32,'yes')`).

Worst part: it is **silent**. `rotate()` returns text on the failure path, so pg_cron records
SUCCESS and `ash.status()` showed nothing. Only `time_since_rotation` crept up.

## Three parts

1. **Break the deadlock** — the gate now ignores raw minute groups already past
   `rollup_1m_retention_days`. Rolling those up would only create rows `rollup_cleanup()` deletes
   on its next pass, so there is nothing left to protect. Groups still *inside* retention keep the
   full #81 check.
2. **Make the bad geometry unreachable** — new `config_raw_rollup_geometry_check` constraint:
   `(num_partitions - 1) * rotation_period <= rollup_1m_retention_days * interval '1 day'`,
   enforced both as a table CHECK and in `rebuild_partitions()`. A table CHECK is used because
   there is no supported setter for these values — direct `UPDATE` is the interface.
3. **Make failure loud** — new `consecutive_rotate_failures` counter on `ash.config`, surfaced in
   `ash.status()`, incremented on every silent `failed:` return and reset by a real rotation.
   Neutral `skipped:` returns leave it alone.

## Verification

RED first against unpatched SQL, with exact-value assertions:

    ERROR:  rotation did not recover after cleanup: failed: pre-truncation rollup incomplete ...
    ERROR:  status failure metric should be exactly {1}, got NULL
    ERROR:  unexpected rebuild geometry result: <accepted>

Then GREEN, including that the **#81 gate still refuses** to truncate when a group inside retention
is genuinely unrolled (the control that proves the fix didn't just disable the guard), and that the
counter goes `{1}` → `{2}` across repeated failures and resets on success.

Fresh install, full 1.0→2.0 chain, double re-apply, fresh-vs-chain schema equivalence, and a
zero-argument smoke of every reader all clean.

## Two things I want a reviewer's eye on

- **`(num_partitions - 1)` is deliberately stricter than the `(num_partitions - 2)` retention
  formula used everywhere else** (`:2268`, `:2374`, `:2315`, `:2475`, README:364 — and documented
  by in-flight #152). The `-1` is correct *here*: it bounds the age of the oldest row in the slot
  about to be truncated, not the user-visible retention window. But the two formulas now differ in
  the same file with no comment saying why, so a later "consistency fix" to `-2` would silently
  reintroduce this deadlock for the boundary slot. **Worth adding that comment before merge.**
- **Upgrade path mutates operator config.** An existing install with unsafe geometry has its
  `rollup_1m_retention_days` raised to the minimum safe whole-day value so the constraint can be
  added. It emits `raise warning` first, and raising minute retention is the only non-destructive
  direction (reducing `num_partitions` would drop raw data, and rejecting the installer would deny
  affected systems the very fix they need) — but it does increase disk usage without asking.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -3578,6 +3578,316 @@ jobs:
 
           echo "Critical bug #81 regression tests PASSED"
 
+      - name: "Critical bug #211: rollup cleanup must not freeze rotation"
+        env:
+          PGPASSWORD: postgres
+        run: |
+          psql -h localhost -U postgres -d postgres -v ON_ERROR_STOP=1 << 'EOF'
+          -- Recreate the legacy geometry that made cleanup and rotation
+          -- deadlock. The new constraint is removed only in this transaction;
+          -- ROLLBACK restores it and every fixture below.
+          begin;
+          alter table ash.config
+            drop constraint if exists config_raw_rollup_geometry_check;
+
+          do $$
+          declare
+            v_minute_ts int4;
+            v_wait_id smallint;
+            v_rollup_result int;
+            v_cleanup_result text;
+            v_rotate_result text;
+            v_current_slot smallint;
+            v_raw_rows bigint;
+            v_rollup_rows bigint;
+          begin
+            perform ash.stop();
+            truncate ash.sample;
+            truncate ash.rollup_1m;
+            truncate ash.rollup_1h;
+            truncate ash.wait_event_map restart identity cascade;
+            truncate ash.query_map_0 restart identity;
+            truncate ash.query_map_1 restart identity;
+            truncate ash.query_map_2 restart identity;
+
+            v_minute_ts := ash.ts_from_timestamptz(
+              date_trunc('minute', now() - interval '36 hours')
+            );
+            select ash._register_wait('active', 'Issue211', 'ExpiredRollup')
+            into v_wait_id;
+
+            update ash.config
+            set current_slot = 0,
+              rotated_at = now() - interval '2 days',
+              rotation_period = interval '1 day',
+              rollup_1m_retention_days = 1,
+              last_rollup_1m_ts = ash.ts_from_timestamptz(
+                date_trunc('minute', now())
+              ),
+              last_rollup_1h_ts = null,
+              sampling_enabled = true
+            where singleton;
+
+            insert into ash.sample (
+              sample_ts, datid, active_count, data, slot
+            ) values (
+              v_minute_ts,
+              0::oid,
+              1::smallint,
+              array[-v_wait_id, 1, 0]::integer[],
+              2::smallint
+            );
+            insert into ash.rollup_1m (
+              ts, datid, samples, peak_backends, wait_counts, query_counts
+            ) values (
+              v_minute_ts,
+              0::oid,
+              1::smallint,
+              1::smallint,
+              array[v_wait_id, 1]::int4[],
+              '{}'::int8[]
+            );
+
+            -- Shipped job order: the forward-only watermark cannot revisit
+            -- this minute, then cleanup removes its already-built rollup.
+            select ash.rollup_minute(1000000) into v_rollup_result;
+            assert v_rollup_result = 0,
+              format('expired minute rollup result should be 0, got %s',
+                     v_rollup_result);
+
+            select ash.rollup_cleanup() into v_cleanup_result;
+            assert v_cleanup_result =
+              'cleanup: deleted 1 minute rows, 0 hourly rows',
+              format('unexpected cleanup result: %s', v_cleanup_result);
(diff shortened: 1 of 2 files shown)


NikolayS:

Both configuration paths the deadlock depends on are now rejected outright:
```text
update ash.config set rollup_1m_retention_days = 1 where singleton;
ERROR:  pg_ash: invalid retention geometry:
        (num_partitions - 1) * rotation_period = (3 - 1) * 1 day = 2 days
        exceeds rollup_1m_retention_days = 1 day; ...

select ash.rebuild_partitions(32, 'yes');
ERROR:  pg_ash: invalid retention geometry:
        (num_partitions - 1) * rotation_period = (32 - 1) * 1 day = 31 days
        exceeds rollup_1m_retention_days = 30 days; ...
```
That error message is literally the invariant issue #211's fix bullet 2 asks for, so the endangered-rows geometry cannot be configured on `main` at all. Fix bullet 3 is also present — `ash.status()` returns a `consecutive_rotate_failures` metric.

Full reproduction transcript: [#211](https://github.com/NikolayS/pg_ash/issues/211#issuecomment-5365756389).

This branch's merge base is `37a0e2a` ("docs: polish user-facing text for clarity and consistency", #158) — **27 commits behind `main`**. Sitting between that base and `main` is `0b40b91` ("release: prevent cleanup from deadlocking rotation", #185), merged **2026-07-27**.

Issue #211 was filed **2026-08-03**, a week after that merge, and this PR was opened the same day from the pre-fix base. The same pattern holds for #216, #217, #218, #219 and #220 as a group: all five share the `37a0e2a` base, and the fixes they propose all merged on 2026-07-27 in #180, #181, #182, #185 and #189. That is consistent with the audit having been run against the stale tree rather than against `main`.

### Recommendation

NikolayS:

PR #218, opened 2026-08-03
fix: preserve reader grants across ash.rebuild_partitions() (#213)

### LINKED ISSUE #213: Release blocker: rebuild_partitions() strips every reader grant, including the default pg_monitor grant

## Repro

```
reader role works
select ash.rebuild_partitions(6, 'yes');
-> ash.aas(), ash.summary(), ash.samples(), ash.top(...), ash.chart(),
   ash.timeline(), ash.status()  ALL fail:
   ERROR: permission denied for view query_map_all
```

Same for a plain `pg_monitor` member that never called `grant_reader()`. ACL proof: `query_map_all`, `query_map_0`, `sample_0` come back with `relacl = NULL`, while `sample` / `rollup_1m` keep theirs.

Cause: `:1575` drops `ash.query_map_all`, `:1584-1610` create fresh `sample_N` / `query_map_N`, `:1614` rebuilds the view — none of it replays grants.

## Impact

A documented admin operation takes down every configured dashboard with a hard ERROR — blank panels, not degraded ones.

**Onset is delayed, which defeats the obvious smoke test.** Immediately after the rebuild, raw storage is empty and every reader returns fine; the outage lands minutes later when the first samples arrive.

README:376-378 tells operators to re-grant "monitoring roles", but README:403 advertises that the installer grants `pg_monitor` by default and **nothing** warns that a rebuild destroys that specific grant. The doc also attributes the loss to "new partitions" when the first blocker is the recreated **view**.

## The fix machinery already exists

The installer at `:6283-6360` snapshots reader-role bundles into `pg_temp._ash_install_reader_roles` and replays `grant_reader()` after its own drop/recreate (added for #107). `rebuild_partitions()` performs the same class of drop/recreate and simply does not use it.

NikolayS:
Fixes #213 — `ash.rebuild_partitions()` silently strips every reader's grants.

Found by the pre-tag audit (#160), and independently rediscovered during the PG 17 gate run.
**Draft on purpose**: needs REV and your explicit approval.

## The bug

`rebuild_partitions()` drops `ash.query_map_all`, recreates `sample_N` / `query_map_N`, and rebuilds
the view — replaying none of the grants. Afterwards, as a reader:

    select * from ash.status();
    ERROR:  permission denied for view query_map_all
    CONTEXT:  SQL statement "select count(*) from ash.query_map_all"
              PL/pgSQL function status() line 31 at SQL statement

`aas()`, `summary()`, `samples()`, `top()`, `chart()`, `timeline()` all fail the same way. A plain
`pg_monitor` member is hit too, since 2.0 grants it the reader bundle by default.

**Onset is delayed**, which is what makes it nasty: right after the rebuild, raw storage is empty
and every reader returns fine. The outage lands minutes later when the first samples arrive — so
the obvious smoke test passes. The regression test here therefore rebuilds, *then samples*, *then
reads as the reader role*.

## The fix

Snapshots explicit SELECT ACLs for every recreated relation, then restores them:

- Complete reader bundles → re-run `ash.grant_reader()`, so newly numbered partitions are covered
- Partial / hand-narrowed grants → restored **by exact relation name**, including grant options,
  never widened to the full bundle
- Roles with no grants → get nothing

Also fixes README:376-378, which attributed the loss to "new partitions" when the recreated *view*
is the first blocker, and never warned that a rebuild destroys the default `pg_monitor` grant.

## Verification

RED against unpatched SQL, after `take_sample()` inserted exactly one row — the real error above,
reproduced through the delayed-onset path rather than an ACL-only check.

GREEN on **PG 18.3 and PG 17**, through a 3→4→3 rebuild cycle:
- named reader and `pg_monitor` both call `ash.status()` and `ash.samples()` successfully after sampling
- both receive all 14 reader relations after 3→4
- partial role stays exactly `query_map_0:true, sample_1:false` — not widened
- no-access role stays `{}`
- decoded sample is exactly one `Timeout:PgSleep` row; `samples_total = 1`, `num_partitions = 4`

Fresh install, double re-apply, and the full 1.0→2.0 chain all clean with zero errors or warnings.

## Review note

**No shared helper was factored out** with the installer's equivalent #107 block (`:6283-6360`),
deliberately: that path spans a script-wide function-recreation cycle while this one is
transaction-local and relation-specific, and unifying them would drag install-order and hardening
dependencies into an already-verified path. Reasonable, but it does mean two places now implement
grant preservation — worth a look if you'd rather they converge.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -3000,6 +3000,262 @@ jobs:
           $$;
           EOF
 
+      - name: "Issue #213: rebuild_partitions preserves reader grants"
+        env:
+          PGPASSWORD: postgres
+        run: |
+          # A rebuild drops and recreates query_map_all plus every sample_N /
+          # query_map_N relation. The outage used to be latent: empty storage
+          # let readers appear healthy until the first sample made status()
+          # touch the freshly-created, ungranted query_map_all view. Drive that
+          # path before inspecting ACLs so this test fails on the user-visible
+          # bug, not only on a catalog difference.
+          psql -h localhost -U postgres -d postgres -v ON_ERROR_STOP=1 << 'EOF'
+          do $$
+          begin
+            if not exists (
+              select from pg_roles where rolname = 'cx213_rebuild_reader'
+            ) then
+              create role cx213_rebuild_reader;
+            end if;
+            if not exists (
+              select from pg_roles where rolname = 'cx213_rebuild_partial'
+            ) then
+              create role cx213_rebuild_partial;
+            end if;
+            if not exists (
+              select from pg_roles where rolname = 'cx213_rebuild_none'
+            ) then
+              create role cx213_rebuild_none;
+            end if;
+          end $$;
+
+          select ash.grant_reader('cx213_rebuild_reader');
+          grant select on ash.query_map_0 to cx213_rebuild_partial
+            with grant option;
+          grant select on ash.sample_1 to cx213_rebuild_partial;
+
+          do $$
+          declare
+            v_acl text[];
+            v_full_acl constant text[] := array[
+              'query_map_0:false', 'query_map_1:false',
+              'query_map_2:false', 'query_map_all:false',
+              'sample_0:false', 'sample_1:false', 'sample_2:false'
+            ];
+            v_role constant name[] := array[
+              'cx213_rebuild_reader'::name, 'pg_monitor'::name,
+              'cx213_rebuild_partial'::name, 'cx213_rebuild_none'::name
+            ];
+            v_expected text[];
+          begin
+            for i in 1 .. cardinality(v_role) loop
+              v_expected := case
+                when i <= 2 then v_full_acl
+                when i = 3 then array['query_map_0:true', 'sample_1:false']
+                else array[]::text[]
+              end;
+              select coalesce(
+                       array_agg(
+                         relation.relname || ':' || acl.is_grantable
+                         order by relation.relname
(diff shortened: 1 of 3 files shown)


NikolayS:

Reader grants survive `rebuild_partitions()` intact:
```text
BEFORE: granted_tables=12  qm_all=t  sample_tbl=t  schema_usage=t  pg_monitor_qm=t
AFTER : granted_tables=14  qm_all=t  sample_tbl=t  schema_usage=t  pg_monitor_qm=t
```
12 to 14 only because the rebuild went from 3 partitions to 4. `ash.query_map_all` stays readable, schema `USAGE` is intact, and the default `pg_monitor` grant is intact.

Full reproduction transcript: [#213](https://github.com/NikolayS/pg_ash/issues/213#issuecomment-5365756726).

This branch's merge base is `37a0e2a` ("docs: polish user-facing text for clarity and consistency", #158) — **27 commits behind `main`**. Sitting between that base and `main` is `8540219` ("release: fix reader privilege blockers and sample slot race", #181), merged **2026-07-27**.

Issue #213 was filed **2026-08-03**, a week after that merge, and this PR was opened the same day from the pre-fix base. The same pattern holds for #216, #217, #218, #219 and #220 as a group: all five share the `37a0e2a` base, and the fixes they propose all merged on 2026-07-27 in #180, #181, #182, #185 and #189. That is consistent with the audit having been run against the stale tree rather than against `main`.

### Recommendation

NikolayS:

PR #219, opened 2026-08-03
fix: separate raw retention boundary from oldest held sample (#212)

### LINKED ISSUE #212: Release blocker: README's flagship drill hard-errors on any fresh install; raw_retention_start is unusable as a planning boundary

## Two conflated concepts

`ash._raw_retention_start()` (`:3430`) returns `min(sample_ts)` — the *oldest sample currently held*, not a retention boundary. `ash._raise_tie_retention()` (`:3527`) then treats it as one.

## Repro — fresh install, 8 samples, data 100% present

The untied `ash.top('wait_event')` reads the same data fine. But:

```
-- README.md:193, verbatim
ERROR: pg_ash: this drill needs raw samples; raw retention starts at 22:34:51
       but the requested window starts at 22:30:01. Narrow the window to start
       at or after 22:35:00...
```

Feed it the value the docs point at:

```sql
since => (select value::timestamptz from ash.status() where metric='raw_retention_start')
-- ERROR: raw retention starts at 22:34:51 but the requested window starts at 22:34:51
```

…because the reader floors `since` to the minute *before* the guard runs. Follow the error's own printed advice (`since => 22:35:00`, rounded **up**):

```
(0 rows)
```

The entire partial first minute is discarded.

## Impact

The payoff step of the documented investigation flow errors on first use, and blames retention for data that was never lost. Obeying the printed remediation returns a silent empty result that reads as "no query was responsible."

`ash.start()` defaults to a 1-second interval, so `raw_start` lands mid-minute with 59/60 probability — **this is the default path, not an edge case.** Same trap after every `ash.rebuild_partitions()` (i.e. immediately after README's own retention runbook) and after any sampler outage.

`ash.report()` at `:5130-5131` already floors `_raw_retention_start()` to the minute for exactly this reason — the tie guard is internally inconsistent with the report path in the same file.

## ⚠️ Interaction with PR #129

**#129 moves this boundary newer.** It correctly fixes #128 (boundary too old) and makes this issue strictly worse. Fix this with, or before, #129.

## Fix

- Separate "oldest sample held" from "raw retention boundary"; `ash.status().raw_retention_start` should report the latter, since that is what it is documented to be
- Floor both sides of the tie guard consistently
- Round the remediation boundary **down**, not up, so following the advice returns the data

NikolayS:
Fixes #212 — the README's flagship drill hard-errors on any fresh install, and
`ash.status().raw_retention_start` is unusable as the planning boundary it is documented to be.

Found by the pre-tag audit (#160). **Draft on purpose**: needs REV and your explicit approval.

## The bug

`ash._raw_retention_start()` returns `min(sample_ts)` — the *oldest sample currently held* — but
`_raise_tie_retention()` treats it as a *retention boundary*, and `status()` publishes it under a
name documented for window planning. On a young install with the data 100% present:

    -- README.md:193, verbatim
    ERROR: this drill needs raw samples; raw retention starts at 22:34:51 but the
           requested window starts at 22:30:01. Narrow the window to start at or after 22:35:00

Feed it `status().raw_retention_start` and it *still* raises (the reader floors `since` to the
minute before the guard runs). Follow the error's own printed advice and you get **0 rows** — the
whole partial first minute discarded. With `ash.start()`'s 1s default this fires 59/60 of the time.

## The fix

Splits the two conflated concepts rather than fudging one:

- **`ash._raw_retention_boundary()`** (new) — the configured ring boundary,
  `rotated_at - (num_partitions - 2) * rotation_period`, floored to the minute, honouring an older
  sample if one is still physically retained. Used by `status()` and every tie guard.
- **`ash._raw_retention_start()`** (unchanged) — still the exact oldest held sample. Still used by
  source selection and report attribution, where "what data exists" is the right question.
- Both sides of the tie comparison are floored consistently, so a window starting exactly at the
  boundary is accepted.
- The remediation hint in the error is rounded **down**, so following it returns data.

## Verification

RED against unpatched SQL — the real README:193 failure, reproduced:

    ERROR: README:193 raised on a young install: pg_ash: this drill needs raw samples...

GREEN, with exact-value assertions rather than "returns rows":

- README:193 → `1 / 212212 / raw`, metrics `(0.03, 0.13, 0.13, 8.00, 100.00)`
- passing `status().raw_retention_start` straight back in → same eight backend-seconds
- following the printed remediation verbatim → same non-empty result
- **genuine age-out still raises**: a row is inserted into slot 2, `rotate()` truncates it, its
  count is asserted to be exactly zero, and the old window raises SQLSTATE `P0001` with the exact
  unrecoverable message. That is the control proving the guard was fixed, not disabled.

Fresh install, double re-apply, full 1.0→2.0 chain, and fresh-vs-chain schema/ACL equivalence all
pass on PG 18.3.

## ⚠️ Merge order and a required edit to #129

**#129 must merge first, then this.** Composition was tested against #129 head `2006159`, not
assumed.

The code cherry-picks with **no textual conflict** — but **one of #129's own test assertions
becomes wrong** once this lands. #129 asserts that `status().raw_retention_start` equals the exact
oldest retained sample; this PR redefines that field as the configured ring boundary. Whoever
merges must update that assertion to compute the boundary from `rotated_at`, `num_partitions` and
`rotation_period`. Verified composed and green after exactly that change:

    NOTICE: raw retention excludes the oldest ring slot PASSED
    NOTICE: issue #212 raw retention boundary tests PASSED
    NOTICE: retention error split PASSED (boundary hint is followable, echoes passed start)

This branch deliberately remains one clean commit on `37a0e2a`; the composition was validated on a
scratch branch and not included here.

## Review note

`raw_retention_start` now means **two different things** on two user-facing surfaces:
`status()` reports the *planning boundary*, while `report().coverage.raw_retention_start` keeps
the *exact oldest sample* (it describes attribution availability, not planning). Each choice is
defensible on its own; sharing the name is the wart. Worth renaming one before this ships —
`report()`'s payload contract is frozen per 2.0 minor line, so it is easier to change now than later.

Related but distinct: #217 introduces a geometry constraint using `(num_partitions - 1)`. That is
not an inconsistency with the `- 2` used here — `-1` bounds the physical age of rows in the slot
about to be truncated, `-2` is the user-visible retention window. Both PRs would benefit from a
comment saying so.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -3987,6 +3987,233 @@ jobs:
           end $$;
           EOF
 
+      - name: "Regression #212: raw retention boundary preserves the first minute"
+        env:
+          PGPASSWORD: postgres
+        run: |
+          psql -h localhost -U postgres -d postgres -v ON_ERROR_STOP=1 << 'EOF'
+          do $$
+          declare
+            v_minute timestamptz := date_trunc('minute', now());
+            v_oldest timestamptz := date_trunc('minute', now())
+                                      - interval '1 minute' + interval '11 seconds';
+            v_boundary timestamptz := date_trunc('minute', now())
+                                        - interval '11 minutes';
+            v_aged_sample timestamptz := date_trunc('minute', now())
+                                          - interval '20 minutes'
+                                          + interval '11 seconds';
+            v_aged_boundary timestamptz := date_trunc('minute', now())
+                                             - interval '10 minutes';
+            v_wait smallint;
+            v_query_map_id int4;
+            v_datid oid;
+            v_status_boundary timestamptz;
+            v_count int;
+            v_key text;
+            v_source text;
+            v_avg numeric;
+            v_peak numeric;
+            v_p99 numeric;
+            v_seconds numeric;
+            v_pct numeric;
+            v_message text;
+            v_state text;
+            v_rotate_result text;
+          begin
+            truncate ash.sample_0, ash.sample_1, ash.sample_2;
+            truncate ash.query_map_0, ash.query_map_1, ash.query_map_2;
+            truncate ash.rollup_1m, ash.rollup_1h;
+            alter table ash.query_map_0 alter column id restart;
+
+            update ash.config
+            set current_slot = 0,
+                num_partitions = 3,
+                sample_interval = interval '1 second',
+                rotation_period = interval '10 minutes',
+                -- Mid-minute ring edge: the public boundary must floor this.
+                rotated_at = v_minute - interval '23 seconds'
+            where singleton;
+
+            select ash._register_wait('active', 'Lock', 'tuple') into v_wait;
+            insert into ash.query_map_0 (query_id) values (212212)
+            returning id into v_query_map_id;
+            select oid into v_datid
+            from pg_database
+            where datname = current_database();
+
+            -- A young install: eight one-second samples, all in the first
+            -- partial minute and all attributable to the same lock/query tie.
+            insert into ash.sample (
+              sample_ts, datid, active_count, data, slot
+            )
+            select ash.ts_from_timestamptz(v_oldest + i * interval '1 second'),
+                   v_datid, 1, array[-v_wait, 1, v_query_map_id]::int4[], 0
+            from generate_series(0, 7) as i;
+
+            -- README.md:193 verbatim: a young, fully-held install must not
+            -- reject the flagship query_id-by-wait drill.
+            v_message := null;
+            begin
+              select count(*), min(key), min(source), min(avg_aas),
+                     min(peak_aas), min(p99_aas), min(backend_seconds), min(pct)
+                into v_count, v_key, v_source, v_avg,
+                     v_peak, v_p99, v_seconds, v_pct
+              from ash.top(
+                'query_id',
+                since => now() - interval '5 minutes',
+                wait_event => 'Lock:tuple',
+                order_by => 'peak',
+                n => 5
+              );
+            exception when others then
+              v_message := sqlerrm;
+            end;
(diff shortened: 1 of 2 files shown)


NikolayS:

The separation this PR's title proposes is already in place — the boundary and the oldest held sample are two different functions returning two different values:
```text
select ash._raw_retention_start(), ash._raw_oldest_sample();
 2026-08-20 05:49:00+00 | 2026-08-21 05:49:34+00
```
And the drill does not error. On a fresh install carrying 40 real samples (generated under `pgbench -c 8 -j 2` load):
```text
select count(*) from ash.top('wait_event', since => ash._raw_retention_start());
 7
```
The full README investigation flow runs clean end to end — `ash.periods()`, `ash.top('wait_event_type', order_by => 'peak')`, `ash.timeline()`, `ash.top('query_id')` and `ash.summary()` all return populated, self-consistent results with no `this drill needs raw samples` error.

Full reproduction transcript: [#212](https://github.com/NikolayS/pg_ash/issues/212#issuecomment-5365756581).

This branch's merge base is `37a0e2a` ("docs: polish user-facing text for clarity and consistency", #158) — **27 commits behind `main`**. Sitting between that base and `main` is `86e0ca8` ("release: fix raw retention boundary", #189), merged **2026-07-27**.

Issue #212 was filed **2026-08-03**, a week after that merge, and this PR was opened the same day from the pre-fix base. The same pattern holds for #216, #217, #218, #219 and #220 as a group: all five share the `37a0e2a` base, and the fixes they propose all merged on 2026-07-27 in #180, #181, #182, #185 and #189. That is consistent with the audit having been run against the stale tree rather than against `main`.

### Recommendation

NikolayS:

PR #220, opened 2026-08-03
fix: keep AAS extremes honest across rollup grains (#210)

### LINKED ISSUE #210: Release blocker: rollup_1h grain collapse — top()/compare() emit hour averages in peak_aas/p99_aas, fabricating regressions

NikolayS:
Fixes #210 — the largest blocker from the pre-tag audit (#160). `ash.top()` and `ash.compare()`
report **hour-grain numbers in columns named `peak_aas` / `p99_aas`**, so `compare()` fabricates
regressions on byte-identical load.

**Draft on purpose.** This is the riskiest change in the release: it alters a user-visible contract
on eight readers and changes `ash.compare()`'s return signature. Needs REV and your approval.

## The rule implemented

**NULL `peak_aas` / `p99_aas` whenever the read grain is coarser than the requested bucket.**

Not an arbitrary pick — `ash.timeline()` already follows exactly this rule for filtered `rollup_1h`
reads, with the shipped comment *"a percentile over hour averages would masquerade as a minute
percentile"*. `aas` and `top` were violating an invariant this same file writes down two functions
away. This makes the file self-consistent.

| Sub | Fix |
|---|---|
| **B1a** | `top()` no longer emits hour averages under `peak_aas`/`p99_aas`. `avg_aas`, `backend_seconds`, `pct` stay valid. |
| **B1b** | `aas()` follows the same rule when a filter forces hour grain. |
| **B1c** | `compare()` gains `source_1` / `source_2`, and suppresses all four peak/p99 values when the two windows resolve to different grains. Average deltas remain valid. |
| **database dim** | **Fixed properly rather than NULLed** — `rollup_1h.minute_counts` is stored per `(ts, datid)`, so the minute detail already existed and simply wasn't read. Now expanded through `rollup_1h_minutes`, keeping true minute-grain peak/p99. `AAS_API.md` §2.4's justification ("the per-dimension arrays are stored per hour") was false for this dimension and is corrected. |
| **B1d** | Legacy hours (`minute_counts IS NULL`, ~98% of a mature 1.x→2.0 archive) now report `source = 'rollup_1h_flat'` instead of masquerading as measurement. |

**On B1d's marker choice:** `rollup_1h_flat` was chosen over `data_points = 0` because those hours
still hold usable average and backend-second evidence — zero data points would incorrectly signal
*no coverage*. That reasoning is sound and I'd keep it, but it's a contract decision worth your
confirmation.

## Verification

Tests were **committed before the implementation** (`98e4c2c test:` then `e135d22 fix:`) — the TDD
ordering is visible in the history, not just claimed.

Deterministic fixture: 59 minutes at AAS 1 plus one minute at AAS 10 → 4,140 backend-seconds,
avg 1.15, peak 10.00, p99 4.69. Every expected number is hand-computed.

RED against unpatched SQL, showing the actual wrong values:

    ERROR: B1a top key=CPU* source=rollup_1h avg=1.15 peak=1.15 p99=1.15 seconds=4140.00
    ERROR: B1b filtered source=rollup_1h avg=0.17 peak=0.17 p99=0.17 seconds=600.00
    ERROR: record "v_dim" has no field "source_1"
    ERROR: database spike source=rollup_1h avg=1.15 peak=1.15 p99=1.15 pct=36.51
    ERROR: B1d aas source=rollup_1h expected=60 with_data=60 avg=0.17 peak=0.17 p99=0.17

GREEN: all eight blocks pass, including **cross-reader agreement** — `aas`, `timeline`, `top`,
`summary` and `report` now assert to the *same* 1.15 / 10.00 / 4.69 / 4140.00 for one fixed window.
That agreement is the whole point of the fix, so it is asserted directly.

Also verified:
- **Docker matrix green on PG 14, 15, 16, 17, 18 and 19beta1**
- fresh install, full 1.0→2.0 chain, double re-apply
- fresh-vs-upgrade semantic schema snapshots: 288 lines each, **zero diff**
- `compare()`'s new signature identical on both paths
- zero-argument sweep of every reader (the regression shape #131 currently trips on)

I separately confirmed the riskiest mechanic by hand: a real `v2.0-beta1` install upgraded with
this branch's installer ends up with `source_1 text` / `source_2 text` present and `compare()`
callable. `CREATE OR REPLACE` cannot change a return type, so this depends on the installer's
dynamic drop block — it works, but that is the thing to re-check if anything here is rebased.

## ⚠️ Collides with #131 and #150 — neither is mechanically safe to merge

Both already collide with each other; this is a third party. **Do not resolve any of it with
`-X ours` / `-X theirs`.**

**vs #131** — overlaps `aas` `3885–3896`, `timeline` `4042–4063`, `top` `4470–4489`, catalog
`5752–5764` (all four function comments), and workflow `3837–3992`. Its chart hunk `5519–5560` is
separable. A human must rebuild its partial-hour policy around the new requested/read/effective
grain logic, **preserve the database minute-detail exception**, and keep now-relative and default
calls working. Its unconditional alignment guard must *not* be carried over — that is what breaks
zero-arg `ash.summary()`.

**vs #150** — overlaps `_grain_counts` ~`3578`, `aas` `3800–3867`, `timeline` `3950–4032`,
`_grain_by` `4179–4359`, `top` `4376–4507`, catalog `5752–5764`, blueprint `36–45`, `138`,
`174–204`, `381–400`, workflow `3712–3871`. Order matters: in `_grain_by`, keep this branch's early
database-minute expansion *then* apply #150's query-attribution branch; in `top`, apply #150's
exact-query decision *first*, then this branch's source / legacy-flat / extreme-validity logic.
`compare()`'s new signature belongs here, but its delegated reads must inherit #150's query policy
after the hand merge.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -3723,10 +3723,10 @@ jobs:
           -- p99_aas as a short (rollup_1m-backed) window containing the same spike.
           -- Invariants: (a) monotonic peak — widening the window never shrinks the
           -- peak; (b) seam consistency — identical peak/p99 from either source;
-          -- (c) p99_aas non-null over rollup_1h-backed windows. Also: rollup_hour()
-          -- computes minute_counts exactly, legacy NULL rows flat-fill to the hour
-          -- average (lower bound, never an invented spike), and wait-filtered reads
-          -- honestly stay hour-grain.
+          -- (c) p99_aas non-null over exact rollup_1h-backed windows. Also:
+          -- rollup_hour() computes minute_counts exactly, legacy NULL rows are
+          -- visibly marked as flat estimates, and wait-filtered reads NULL
+          -- minute peak/p99 when only hour-grain evidence survives.
           -- ============================================================================
           do $$
           declare
@@ -3796,35 +3796,42 @@ jobs:
             where data_points > 0;
             assert n = 6, 'timeline 10-min buckets with data over rollup_1h = 6, got ' || n;
 
-            -- wait-filtered long window has no per-minute detail and honestly stays
-            -- hour-grain: peak = IO hour average (59*20+60)/3600 = 0.34, timeline p99 null
+            -- wait-filtered long window has no per-minute detail. Its average
+            -- remains exact, but a requested minute peak/p99 is honestly NULL.
             select * into a from ash.aas(v_h - interval '2 hours', v_h + interval '1 hour',
                                          wait_event_type => 'IO');
-            assert a.source = 'rollup_1h' and a.peak_aas = 0.34,
-              format('filtered long window stays hour-grain: peak=%s (want 0.34)', a.peak_aas);
+            assert a.source = 'rollup_1h' and a.avg_aas = 0.11
+                   and a.peak_aas is null and a.p99_aas is null,
+              format('filtered long window source=%s avg=%s peak=%s p99=%s',
+                     a.source, a.avg_aas, a.peak_aas, a.p99_aas);
             select p99_aas into v_p99
             from ash.timeline(v_h - interval '2 hours', v_h + interval '1 hour', '1 hour',
                               wait_event_type => 'IO')
             where data_points > 0;
             assert v_p99 is null, 'filtered rollup_1h timeline p99 stays null, got ' || coalesce(v_p99::text, 'null');
 
-            -- legacy pre-2.0 row (minute_counts null): flat-fills to its hour average
-            -- (AAS 1.00/min) — peak keeps the real spike, p99 non-null, no fake spike
+            -- legacy pre-2.0 row (minute_counts null): the compatibility path
+            -- still flat-fills averages, but marks the whole read and NULLs the
+            -- minute extremes because the legacy hour's distribution is unknown.
             insert into ash.rollup_1h (ts, datid, samples, peak_backends, wait_counts, query_counts, minute_counts)
             values (ash.ts_from_timestamptz(v_h - interval '1 hour'), 0::oid, 3600, 9,
                     array[v_cpu, 3600]::int4[], '{}'::int8[], null);
             select * into a from ash.aas(v_h - interval '2 hours', v_h + interval '1 hour');
-            assert a.peak_aas = 11.30 and a.p99_aas = 1.00 and a.buckets_with_data = 120,
-              format('legacy mix: peak=%s p99=%s bwd=%s (want 11.30 1.00 120)',
-                     a.peak_aas, a.p99_aas, a.buckets_with_data);
+            assert a.source = 'rollup_1h_flat'
+                   and a.peak_aas is null and a.p99_aas is null
+                   and a.buckets_with_data = 120,
+              format('legacy mix: source=%s peak=%s p99=%s bwd=%s',
+                     a.source, a.peak_aas, a.p99_aas,
+                     a.buckets_with_data);
 
-            -- periods(): the long windows (1w, 1mo) carry the true minute peak and a
-            -- non-null p99 (US-6 capacity planning works past rollup_1m retention)
+            -- periods() propagates the legacy marker and does not re-label the
+            -- flat estimate as a measured minute peak/p99.
             for r in select * from ash.periods(v_h + interval '1 hour') loop
               if r.period in ('1w', '1mo') then
-                assert r.peak_aas = 11.30 and r.p99_aas is not null,
-                  format('periods %s: peak=%s p99=%s (want 11.30, non-null)',
-                         r.period, r.peak_aas, r.p99_aas);
+                assert r.source = 'rollup_1h_flat'
+                       and r.peak_aas is null and r.p99_aas is null,
+                  format('periods %s: source=%s peak=%s p99=%s',
+                         r.period, r.source, r.peak_aas, r.p99_aas);
               end if;
             end loop;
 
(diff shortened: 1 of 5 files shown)


NikolayS:

The grain provenance issue #210 reports as absent is present on `main`. `ash.top()` returns `source` and `effective_bucket`; `ash.compare()` returns them **per side** — `source_1`/`source_2` and `effective_bucket_1`/`effective_bucket_2` — which is exactly what stops two sides answered at different grains from being silently comparable. The installer also carries grain-honesty helpers absent at this branch's base: `_rollup_1h_has_flat`, `_exact_query_uses_coarser`, `_raise_tie_retention`, `_grain_counts`, `_grain_by`, `_pick_source_agg`.

Stated honestly: this is the one item of the six I could **not** drive through a full behavioral reproduction, because exercising the `rollup_1h`-backed path needs aged data spanning the `rollup_1m` retention boundary, which I could not synthesize without hand-forging `ash.sample` contents. Strong evidence, not conclusive.

Full reproduction transcript: [#210](https://github.com/NikolayS/pg_ash/issues/210#issuecomment-5365756205).

This branch's merge base is `37a0e2a` ("docs: polish user-facing text for clarity and consistency", #158) — **27 commits behind `main`**. Sitting between that base and `main` is `a1f0a30` ("release: fix rollup_1h grain honesty", #182), merged **2026-07-27**.

Issue #210 was filed **2026-08-03**, a week after that merge, and this PR was opened the same day from the pre-fix base. The same pattern holds for #216, #217, #218, #219 and #220 as a group: all five share the `37a0e2a` base, and the fixes they propose all merged on 2026-07-27 in #180, #181, #182, #185 and #189. That is consistent with the audit having been run against the stale tree rather than against `main`.

### Recommendation

NikolayS:

PR #232, opened 2026-08-21
docs: state the soft-dependency contract and cadence guidance

### LINKED ISSUE #230: Docs: state the embed-by-default contract — degraded-mode matrix and cadence guidance
## Problem

Two questions decide whether pg_ash can be embedded and enabled by default, and neither is answered anywhere in the docs.

**1. What is the real dependency surface?**

- `pg_stat_statements` is a **soft** dependency, and the shape of the degradation matters: without it pg_ash still captures query **IDs** (they come from `pg_stat_activity`); only query **text** is lost. This distinction is load-bearing, because on vendor-managed Postgres an application's own database user cannot install extensions — a hard dependency would make pg_ash uninstallable for a large share of users.
- `pg_cron` is likewise optional, and it is **not** universally available: it is a third-party extension, absent on several managed platforms, and some integrators will deliberately run without it to keep scheduling under their own control.

The README's `## Requirements` and `## Scheduling` sections cover parts of this, but nowhere state the degraded-mode contract as a single answer: what exactly still works, and what exactly is lost, for each of the four combinations.

**2. Which sampling interval should be used?**

- 1/sec is the practical target — natural rather than ideal; the true ideal would be wait-event tracing, which is not available here.
- 15s is workable and comparable to what managed platforms sample at.
- 1 minute is too coarse: spikes vanish, and a minute of trouble can average away to nothing. Schedulers capped at minute granularity walk straight into this.
- Even 1/sec cannot resolve the ordering of waves inside a multi-second incident — for instance whether lock-manager contention preceded buffer-mapping contention or followed it.

An integrator choosing a cadence today has to infer all of this.

## Scope

Docs only.

- A degraded-mode matrix: pg_cron x pg_stat_statements, four rows, each stating what works, what is lost, and what the operator must do instead.
- A cadence section: recommended 1s, acceptable 15s, why 1 minute is not acceptable, and the honest statement that sub-second incident ordering is not resolvable by sampling at all.
- The embedding story: separate schema, installable via an ordinary application migration, no extension required.

## Acceptance criteria

- [ ] Matrix present, all four combinations, no "should work" hedging.
- [ ] Explicitly states that query IDs survive without pg_stat_statements and only text is lost.
- [ ] Cadence section names 1s / 15s / 1min with the reasoning, and states the resolution limit honestly.
- [ ] pg_cron described as third-party and not universally available, with the external-scheduler path given equal standing rather than as a footnote.
- [ ] No claim about overhead that is not backed by #226.

## Test plan

Docs-only, so the guard is against drift. CI already exercises degraded modes (`devel/tests/degraded_no_cron.sql`, `devel/tests/degraded_no_pgss.sql`); this change should make the README matrix match what those tests actually assert. Where the matrix claims a capability survives degradation, there must be a corresponding assert proving it — add the missing asserts rather than weakening the matrix. Specifically: assert that with pg_stat_statements absent, query-ID attribution still returns the exact seeded query IDs and only the text column is null.

NikolayS:
Closes #230.

## Why

Two questions decide whether pg_ash can be embedded in another product and enabled by default, and the README answered neither.

**"What is the minimal set of dependencies?"** The answer that makes embedding viable is that `pg_stat_statements` is soft *in a specific way*: query **IDs** survive without it (they come from `pg_stat_activity`), only query **text** is lost. That distinction matters because on vendor-managed Postgres an application's own database user cannot install extensions. The old three-bullet Requirements list — "optional but recommended for `query_text`" — does not convey it.

**Which sampling interval?** 1s is the target, 15s is workable, 1 minute is too coarse and loses spikes outright. Schedulers capped at minute granularity walk straight into that.

## What changed

- `## Requirements` becomes a real dependency contract: a four-row matrix over pg_cron x pg_stat_statements, each row saying what works and what is lost. pg_cron is described as third-party and not universally available, with the external-scheduler path given equal standing rather than framed as a consolation prize.
- A cadence section with the 1s / 15s / 1min tradeoff, plus the limit that no interval setting fixes: 1-second sampling cannot tell you which wait came first inside a multi-second incident. That needs wait-event tracing, which pg_ash does not provide. Better to say so than to imply precision we do not have.

## Test

The matrix makes a behavioral claim, so it gets a behavioral assert. `devel/tests/degraded_no_pgss.sql` previously asserted only that `query_text` is `NULL` without pg_stat_statements — which cannot distinguish "text degraded" from "query attribution gone entirely". Added exact-value asserts pinning the surviving `query_id` on both `ash.top()` and `ash.samples()`.

Verified against real PostgreSQL 18 in Docker with pg_stat_statements absent:

- red (expected id perturbed to 888888): `ERROR: top must still attribute the exact query_id without pgss, got 999999`
- green: `All degraded-mode (no pgss) 2.0 reader tests PASSED`

No SQL changed — `sql/` and `devel/sql/` are untouched.

### DIFF
--- a/README.md
+++ b/README.md
@@ -510,9 +510,7 @@ surface from the catalog alone.
 
 ## Requirements
 
-- Postgres 14+
-- `pg_stat_statements` optional but recommended for `query_text`
-- `pg_cron` optional but recommended for built-in scheduling
+Postgres 14+. Nothing else is required.
 
 `compute_query_id` must be on for useful query attribution:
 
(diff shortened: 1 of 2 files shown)


NikolayS:

Corrected dependency and cadence documentation is being consolidated in #254, including observed runtime behavior and executed degraded-mode examples.

PR #233, opened 2026-08-21
docs: restore the LLM-analysis section, grounded in AAS and ash.report()

### LINKED ISSUE #225: Docs: restore the LLM-analysis section, grounded in AAS and ash.report()
## Problem

The README once had an LLM-analysis section; it was removed and never replaced. It should come back, but rebuilt around AAS rather than raw sample dumps.

`ash.report()` already emits exactly the structured payload an LLM wants: one JSONB document, stable keys, no scraping. What is missing is the context that makes the numbers mean anything. "AAS" is not general knowledge, and a model handed the payload without an explanation of it will confidently issue a verdict on a database it has misread.

Two adjacent needs point at the same gap:

- An embedding application wants a "copy report" affordance — dump the AAS data as JSON, paste it into whatever model the operator already uses, get a plain-language summary. The payload exists; the interpretive prompt does not.
- Anyone building a tool-side skill that processes pg_ash tables needs the same canonical description of what the fields mean, and it should live here rather than being reinvented per integration.

## Scope

Docs only. No SQL changes — `ash.report()` already emits the payload.

Add a README section after `## Machine report` containing:

1. A definition of AAS an LLM can use: the average number of sessions concurrently active or waiting over a window, the Linux load-average analogy, and the rule that it is judged **as a share of vCPU count**.
2. The interpretation ladder: top-level AAS average **and** peak -> wait event **type** (cpu / io / ipc / lock / lwlock) -> specific wait event, with query ID as a *separate* dimension rather than a fourth level.
3. An explicit warning that averages hide incidents, with a worked illustration — a 16-vCPU cluster whose weekly average AAS reads as comfortably healthy while its worst single minute is many multiples of core count. A prompt that only reads `aas_avg` will call that cluster healthy.
4. A copy-pasteable prompt that takes the `ash.report()` payload and asks for a plain-language verdict.
5. The caveats the payload already carries: `coverage` / `minutes_with_data` must be checked before trusting a verdict, and `top_queryids_available` is false without pg_stat_statements, so the model must not invent query attribution.

## Acceptance criteria

- [ ] README has the section.
- [ ] The prompt names the exact `ash.report()` keys that exist today (`aas_avg`, `aas_worst1m`, `aas_p99`, `aas_p999`, wait classes, `coverage`, `top_queryids_available`) and no invented ones.
- [ ] vCPU normalization is explicit — the report takes `vcpus =>` and the prompt must use it.
- [ ] The peak-versus-average trap is stated with a worked example.
- [ ] No claim that pg_ash calls an LLM itself; this is copy-paste guidance.

## Test plan

Docs-only, so the guard is against **drift**, not behavior: CI must assert that every `ash.report()` key named in the README prompt is actually present in a live payload on a seeded install, pinning the exact key set rather than doing a substring match. A prompt naming fields that no longer exist is worse than no prompt, because the model reports on absent data instead of failing loudly.

NikolayS:
Closes #225.

## Why

The README's LLM-analysis section was removed and never replaced. It should come back, rebuilt around AAS rather than raw sample dumps.

`ash.report()` already returns the payload an LLM wants, so the missing piece was never SQL — it was the prompt that makes the payload interpretable. "AAS" is not general knowledge, and a model handed the JSON without it will confidently issue a verdict on a database it has misread.

## What changed

A `## LLM analysis` section after `## Machine report`, containing a copy-pasteable prompt that:

- defines AAS as the Postgres load average and insists it is judged **against vCPU count** — AAS 8 on 64 cores is idle, AAS 8 on 4 cores is an incident;
- explains the five wait classes in operator terms;
- walks the document's real structure (the `aas_*` objects keyed by `total` plus the five classes, `top_events_*`, `coverage`);
- states the peak-versus-average trap with a worked illustration: a 16-vCPU cluster averaging 4.9 AAS over a week reads as healthy at 31% of capacity, while a worst single minute of 295 AAS is 1844% of it. The average is true and useless;
- forces honesty on missing data: check `top_queryids_available` before attributing queries, check `minutes_with_data` against `minutes_expected` before trusting the verdict, and never invent a query.

## Tests

A prompt that names payload fields rots silently when the contract moves, so this guards both directions:

1. **Docs guard** (existing `Guard docs against stale privilege and reader names` step): every documented field must appear *inside* the LLM analysis section, not merely somewhere in the README — the section is extracted with `sed` before grepping.
2. **Behavioral** — new `devel/tests/report_contract.sql`, run as its own CI step. Self-contained (seeds its own minimal sample) and asserts the payload **shape** the prompt depends on: the exact sorted key sets of `aas_avg` / `aas_worst1m` / `aas_p99` / `aas_p999` (`cpu,io,ipc,lock,lwlock,total`) and of `coverage` (`from,minutes_expected,minutes_with_data,raw_retention_start,source,to`), plus that `vcpus` is echoed exactly when supplied and **absent** when not — so a consumer can distinguish "no core count given" from a fabricated one. Value-level asserts stay with the seeded fixture in the reader API test; this file deliberately covers shape only.

Verified against real PostgreSQL 18 in Docker:

- red (an AAS class key set perturbed): `ERROR: report_contract: aas_avg keys must be cpu,io,ipc,lock,lwlock,total, got cpu,io,ipc,lock,lwlock,total`
- green: `ash.report payload contract PASSED`
- docs guard: green, and verified to trip when a documented field is renamed.

## A CI landmine worth knowing about

The asserts were first written inline in the existing `Test 2.0 reader API (behavioral)` step, which pushed that single step past **27,000 characters**. GitHub Actions then silently refused to materialize *any* job for the workflow on this branch: the `pull_request` run was never created, and a `workflow_dispatch` run sat `queued` with **zero jobs** and no error anywhere. The YAML parsed fine locally, including a strict duplicate-key check, which made it look like a GitHub outage rather than our change.

Moving the SQL into `devel/tests/` — the pattern the repo already uses for `query_attribution_assert.sql` and `stale_rollup_source.sql` — fixed it immediately. Worth remembering before adding more inline SQL to that step; it is the largest one in the workflow by a wide margin.

## Scope

No SQL changed — `sql/` and `devel/sql/` untouched. The CI wiring was drafted with Codex (`gpt-5.6-sol`) and reviewed here; I dropped a redundant `vcpus => 16` echo assert it added next to an existing `vcpus => 8` one, and restructured the placement as described above.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -83,6 +83,22 @@ jobs:
             fi
           done
 
+          # Keep the copyable LLM prompt aligned with ash.report's payload.
+          if ! grep -Fq "## LLM analysis" README.md; then
+            echo "README.md must retain the LLM analysis section" >&2
+            exit 1
+          fi
+          llm_analysis="$(sed -n "/^## LLM analysis$/,/^## /p" README.md)"
+          for field in \
+            "aas_avg" "aas_worst1m" "aas_p99" "aas_p999" "top_events_" \
+            "top_queryids_available" "coverage" "minutes_with_data" \
+            "minutes_expected" "raw_retention_start" "vcpus"; do
+            if ! grep -Fq "$field" README.md || ! grep -Fq "$field" <<<"$llm_analysis"; then
+              echo "README.md LLM analysis must document ash.report field $field" >&2
+              exit 1
+            fi
+          done
+
           if ! grep -Fq "Only \`ash.rebuild_partitions\` and \`ash.uninstall\` require the exact \`'yes'\` confirmation token." README.md; then
             echo "README.md must limit the exact 'yes' confirmation contract to rebuild_partitions and uninstall." >&2
             exit 1
@@ -5733,6 +5749,12 @@ jobs:
           PGPASSWORD: postgres
         run: |
           psql -h localhost -U postgres -d postgres -v ON_ERROR_STOP=1 -f devel/tests/query_attribution_assert.sql
+
+      - name: "Test 2.0: ash.report payload contract the README prompt relies on"
+        env:
+          PGPASSWORD: postgres
+        run: |
+          psql -h localhost -U postgres -d postgres -v ON_ERROR_STOP=1 -f devel/tests/report_contract.sql
 
       - name: "Test 2.0: stale and rounded-up rollup source selection"
         env:
--- a/README.md
+++ b/README.md
@@ -353,6 +353,84 @@ naming that alternate source; it does not synthesize per-minute class data.
 The payload contract is stable for the 2.0 minor line: keys may be added, not
 renamed or removed.
 
+## LLM analysis
+
+`ash.report()` is already the shape an LLM wants: one JSONB document, stable
+keys, no scraping. What a model lacks is the meaning of the numbers. "AAS" is
+not general knowledge, and a model that guesses at it will confidently call a
+sick database healthy.
+
+Pass the payload together with the prompt below.
+
+```sql
+select ash.report(
+  since => now() - interval '1 day',
+  vcpus => 16   -- required for a verdict: AAS is judged against core count
+);
+```
+
+<details>
+<summary>Prompt (copy this, then paste the JSON after it)</summary>
+
+```text
+You are analyzing a PostgreSQL database's Active Session History, produced by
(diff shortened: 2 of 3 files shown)


NikolayS:

The restored original LLM investigation workflow and corrected AAS/report semantics are being consolidated in #254, with live PostgreSQL output and read-only examples.

Issue #223, opened 2026-08-21
ash.start() misleads integrators: it registers pg_cron jobs, and external tickers still need it


NikolayS:
## Problem

`ash.start()` is named for something it does not do, and the gap costs integrators real collection time.

It starts no process. It does two things:

1. sets `sampling_enabled` in `ash.config`;
2. registers pg_cron jobs, when pg_cron is available.

Three questions that integrators hit immediately cannot be answered from the README:

| Question | Answer |
|---|---|
| What does it start? | Nothing. There is no background process. |
| Must it be re-called after a Postgres restart? | No. pg_cron jobs live in `cron.job`, an ordinary table. |
| Is it idempotent? | Yes — stated only in a `comment on function`, nowhere in the README. |

Worse, the name invites a conclusion that is wrong in the expensive direction: *"if I am not using pg_cron, I do not need to call it at all."* In fact `ash.start()` also sets `sampling_enabled`, and `ash.take_sample()` no-ops (bumping `skipped_samples`) whenever that flag is false — see `sql/ash-install.sql:963`. Anyone driving `take_sample()` from their own scheduler after an `ash.stop()` collects nothing while every call appears to succeed.

This matters now: 2.0 is intended to be the last interface change on this line, so a rename must land before the 2.0 tag or never.

## Proposed design (needs a decision on the name)

The name is a product decision, so this proposes rather than assumes. Recommendation, in preference order:

1. **`ash.schedule(every)`** — describes exactly what happens (it writes pg_cron schedules), pairs naturally with an `unschedule` inverse, and implies no daemon. It also reuses vocabulary pg_cron users already have.
2. `ash.register(every)`.

Whatever name wins, the mechanics should be:

- Rename `ash.start(interval)` and its `ash.stop()` inverse.
- Keep `ash.start()` / `ash.stop()` as thin wrappers **for the 2.0 line only**, emitting a `NOTICE` naming the replacement, so existing scripts survive the upgrade. Drop the wrappers in 2.1.
- Split the two responsibilities the current function conflates, so the external-ticking path stops depending on a function named "start": `ash.enable()` / `ash.disable()` toggle `sampling_enabled` only, and the schedule function calls `ash.enable()` internally.

## Acceptance criteria

- [ ] Name decided; this issue updated with the decision.
- [ ] New function exists under the agreed name; the old name still resolves on the 2.0 line with a NOTICE pointing at the replacement.
- [ ] `ash.status()` reports scheduler state under the new vocabulary.
[...]

PR #234, opened 2026-08-21
docs: state what ash.start() actually does, and test the trap it hides

NikolayS:
Refs #223 (does not close it — the rename is still open).

## Why

`ash.start()` is the single biggest source of confusion for anyone integrating pg_ash. Three questions cannot be answered from the README:

| Question | Answer |
|---|---|
| What does it start? | Nothing. It sets `sampling_enabled` and registers pg_cron jobs. There is no background process. |
| Must it be re-called after a Postgres restart? | No. pg_cron jobs live in `cron.job`, an ordinary table. |
| Is it idempotent? | Yes — stated only in a `comment on function`, nowhere in the README. |

And the name invites a conclusion the docs do nothing to prevent: *"if I am not using pg_cron, I do not need to call it at all."*

That is wrong in the expensive direction. `ash.stop()` clears `sampling_enabled`; `ash.take_sample()` then returns `0` and increments `skipped_samples` instead of collecting (`sql/ash-install.sql:963`). Anyone ticking `take_sample()` from their own scheduler on a stopped installation sees every call succeed and collects nothing.

## What changed

- README `## Scheduling` gains a subsection stating what `ash.start()` does and does not do, the idempotency and restart facts, and the external-ticker trap with the two fields (`sampling_enabled`, `skipped_samples`) that diagnose it.
- `devel/tests/degraded_no_cron.sql` called `ash.take_sample()` at *exactly* the point where sampling had just been disabled by `ash.stop()` — and asserted nothing about the result. The trap was live and untested. Now asserted in both directions with exact values:
  - disabled: returns `0` **and** `skipped_samples` increases by exactly 1;
  - re-enabled: `skipped_samples` does **not** move.

The return value is deliberately not asserted positive: `take_sample()` filters `pid <> pg_backend_pid()`, so an idle test database correctly captures zero backends. `skipped_samples` is the honest discriminator between "sampled and found nothing" and "refused to sample".

Verified against real PostgreSQL 18 in Docker with pg_cron absent:

- red (expected delta perturbed to +2): `ERROR: disabled take_sample() must bump skipped_samples by exactly 1, got 0 -> 1`
- green: `All degraded-mode (no pg_cron) tests PASSED`

## Deliberately not done here

The **rename** (`start` -> `schedule` / `register`). Naming the flagship lifecycle function is a product decision, it must land before the 2.0 tag since 2.0 is meant to be the last interface change, and it touches demos, blueprints, and catalog comments repo-wide. #223 carries a concrete proposal and a full test plan. This PR is the half that is unambiguous and safe to land now.

No SQL changed — `sql/` and `devel/sql/` untouched.

### DIFF
--- a/README.md
+++ b/README.md
@@ -375,6 +375,29 @@ Only `ash.rebuild_partitions` and `ash.uninstall` require the exact `'yes'` conf
 
 ## Scheduling
 
+### What `ash.start()` does, and does not do
+
+`ash.start()` starts no process. Despite the name, it does two things:
+
+1. sets `sampling_enabled` in `ash.config`;
+2. registers pg_cron jobs, when pg_cron is available.
+
+Consequences worth knowing before you wire it into an application:
+
+- **It is idempotent.** Calling it again does not duplicate jobs.
+- **You do not need to call it again after a restart.** pg_cron jobs live in
+  `cron.job`, which is an ordinary table; they survive restarts like any other
+  row. There is no in-memory state to lose.
+- **You still need it when you tick externally.** This is the trap. `ash.stop()`
+  clears `sampling_enabled`, and `ash.take_sample()` then returns `0` and
+  increments `skipped_samples` instead of collecting. An external scheduler
+  calling `ash.take_sample()` on a stopped installation will appear to run
+  perfectly and collect nothing. `ash.status()` reports both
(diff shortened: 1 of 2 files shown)


NikolayS:

The start/stop contract is being consolidated across #251, #253 and #254, including resumed cadence, job ownership and custom-command behavior.

Issue #228, opened 2026-08-21
Design: evaluate transaction control (chunked COMMIT) in the maintenance procedures


NikolayS:
## Motivation

Follow-up to #221, which adds `CALL`-able procedure forms for the sampling and rollup routines.

#221 deliberately ships those procedures as **thin wrappers with identical semantics** — the immediate goal was the routing property (`CALL` cannot be mistaken for a read by a query router). But the other reason [PostgreSQL procedures exist](https://www.postgresql.org/docs/current/plpgsql-transactions.html) is transaction control, which the function forms fundamentally cannot do, and that is worth its own evaluation.

Where it could pay off:

- **`rollup_minute(batch_limit)`** catches up to 60 minutes of grains in **one** transaction. A long catch-up after an outage means a long-lived transaction holding back `xmin` on the primary — and with `hot_standby_feedback` on, on the replicas too. Committing per grain would bound that.
- **`rollup_cleanup()`** deletes from `rollup_1m` / `rollup_1h` in one statement per table. Chunked, committed deletes are the standard way to keep retention purges from locking up or bloating in one shot.
- **`rebuild_partitions()`** does a large multi-DDL sequence atomically today.

## Why it needs design, not just implementation

The routines rely on **transaction-scoped advisory locks** for mutual exclusion:

```sql
perform pg_advisory_xact_lock(hashtext('pg_ash')::int4, hashtext('pg_ash_rollup')::int4);
```

A `COMMIT` inside the procedure **releases those locks**, so naive chunking would drop the mutual exclusion that `rollup_minute`, `rollup_hour`, `rollup_cleanup` and `rebuild_partitions`'s drain poll all depend on. Any chunking design has to either switch to session-level locks with disciplined release, or re-acquire per chunk and re-validate the state it read before the commit.

This is the area with the worst bug history in the project — #211 (permanent rotation deadlock), #162, #81 (silent raw sample loss) — so it needs a written design and a strong test plan, not an opportunistic change.

There is also a hard usage constraint: a procedure can only commit when it is **not** called from inside an outer transaction or another atomic context. `CALL` from pg_cron is fine; `CALL` from inside an application transaction is **not** — it raises `invalid_transaction_termination`. So a committing procedure changes the contract for callers, and that has to be documented and probably detected.

## Scope of this issue

Decide, with a written rationale:

1. Which routines (if any) should chunk-and-commit, and at what granularity.
2. How mutual exclusion survives the commits.
[...]

Issue #224, opened 2026-08-21
Feature: optional UNLOGGED sample partitions to cut WAL amplification


NikolayS:
## Motivation

pg_ash writes roughly 50–245 MiB/day of raw samples at 1 Hz for 100–500 active backends, and the WAL for that is larger still (full-page writes plus index maintenance). Every one of those bytes ships to **every** replica, **every** disaster-recovery site, and the WAL archive / PITR store. On a quiet instance that can be a large share of total WAL — for data whose entire purpose is a short-lived diagnostic ring buffer.

`UNLOGGED` raw sample partitions remove that WAL. The mechanics work in our favour:

- `ash.sample` is LIST-partitioned by `slot` into a **fixed ring created once at install**. `rotate()` reuses partitions via `TRUNCATE ... RESTART IDENTITY` and never creates new ones. **`TRUNCATE` preserves persistence**, so a one-time per-partition `ALTER TABLE ash.sample_N SET UNLOGGED` on the empty ring is enough — the setting survives every rotation.
- PostgreSQL **rejects `UNLOGGED` on partitioned parents** (`partitioned tables cannot be unlogged`), and persistence is not inherited by partitions anyway. Per-partition handling is the only forward-compatible form.
- `ash.rebuild_partitions()` creates partitions with `create table if not exists`, so persistence must be **re-asserted** on partitions created by a slot-count change.

The counter-argument decides the default: if the server dies or restarts uncleanly, unlogged data is truncated, so the ASH history is gone exactly when it is wanted for root-cause analysis. **Logged stays the default**; unlogged is opt-in for operators who have consciously traded post-crash forensics for WAL volume.

## Scope

- New config column `ash.config.sample_unlogged bool not null default false`.
- New admin function `ash.set_sample_persistence(mode text)` accepting `'logged'` / `'unlogged'`, which updates the config **and** converts the existing ring in place per partition. Converting a populated ring rewrites it; the function should say so.
- Both partition-creation sites honour the config: the installer and `ash.rebuild_partitions()`, so a slot-count change cannot silently drop back to logged.
- Rollup tables (`ash.rollup_1m`, `ash.rollup_1h`) **stay logged, always.** They are the long-term value and are tiny; keeping them logged means a promoted replica or a crashed primary still has history at rollup granularity even though the raw ring is gone.
- `ash.status()` surfaces the current persistence mode.
- Installer re-apply must not flip persistence back — re-apply is idempotent with respect to this setting.

[...]

Issue #222, opened 2026-08-21
Feature: guard write paths and scheduling against running on a replica


NikolayS:
## Motivation

pg_ash is primary-only by design: `ash.take_sample()` reads the local `pg_stat_activity` and inserts locally, and `ash.sample` has no host/instance column. On a physical standby the `ash` schema arrives read-only and every write path fails.

Today **nothing guards this**. `grep -n pg_is_in_recovery sql/ash-install.sql` returns nothing — not the sampler, not rotation, not the rollups, not `ash.start()`, not the installer. The consequences:

- A sampler pointed at a standby (or a primary that got demoted, or a node in maintenance mode) emits one error per second, forever. Writability has to be checked **per tick, not at boot**, because failover happens mid-life.
- After a failover, a daemon holding its own connection needs `pg_is_in_recovery()` to notice it is now talking to a standby.
- Installing on a replica looks like it might work and silently isn't a workaround.

## Scope

Add `ash._in_recovery()` and apply it with **two deliberately different behaviours**, because the callers are different:

**Scheduled routines → clean no-op, not an error.** `take_sample()`, `rotate()`, `rollup_minute()`, `rollup_hour()`, `rollup_cleanup()` return their neutral "did nothing" value and emit a `NOTICE` when in recovery. A pg_cron job left behind on a promoted-then-demoted node must not spam the log once per second, and a monitoring wrapper must not see a hard failure for a condition that is expected and transient. This also replaces an opaque `cannot execute INSERT in a read-only transaction` with a message that names the actual cause.

**Explicit operator actions → hard, clear error.** `ash.start()`, `ash.stop()`, `ash.rebuild_partitions()`, `ash.uninstall()`, state-changing `ash.set_debug_logging()`, and the installer itself raise a `read_only_sql_transaction` (`25006`) exception naming the standby condition. Someone typing these on a replica made a mistake and should be told, not silently ignored.

**Readers stay unguarded.** `ash.top()`, `ash.aas()`, `ash.report()` and friends are read-only and remain callable wherever the data is visible.

**Report the state.** `ash.status()` gains a row exposing recovery state, so an operator can see node role in the one place they already look.

[...]

PR #235, opened 2026-08-21
feat: add CALL-able maintenance procedures (#221)

### LINKED ISSUE #221: Feature: CALL-able procedure surface for sampling and rollup routines
## Motivation

Query routers, connection poolers, and application-level load balancers commonly decide where to send a statement by its **kind**, not its content: anything that looks like a `SELECT` is a read and may go to a replica. Every pg_ash maintenance entrypoint is syntactically a `SELECT`, including the ones that write:

```sql
select ash.take_sample();     -- INSERTs
select ash.rotate();          -- TRUNCATEs
select ash.rollup_minute();   -- INSERT/UPDATE
```

On a read-routing deployment these land on a read-only standby and fail with `cannot execute INSERT in a read-only transaction`. The failure mode is nasty because it is topology-dependent: it works on a single-node install and only breaks where reads are routed elsewhere.

`CALL` is unambiguously a write path — no router can mistake it for a read. It is also the semantically correct construct for a routine invoked for its side effect, and (unlike a function) a procedure is allowed to manage transactions.

## Design constraint: procedures need distinct names

A procedure cannot reuse a function's callable name. Verified on PostgreSQL 18.4:

- same name + same argument types → `ERROR: function "f" already exists with same argument types`
- same name + different argument types (procedure arg with a `DEFAULT` so `CALL f()` would parse) → creation succeeds, but **both** call forms then break: `ERROR: procedure t.f() is not unique` / `ERROR: function t.f() is not unique`

Since the existing functions are public API and 2.0 readers depend on them, the procedures get a mechanical `ash.run_<function>()` prefix. No per-name invention, and it is trivially greppable for anyone who wants a lint rule forbidding raw `ash.` SQL outside a wrapper layer.

## Scope

Add procedure forms for the maintenance/collection routines:

| existing function | new procedure |
| --- | --- |
| `ash.take_sample() -> int` | `call ash.run_take_sample()` |
| `ash.rotate() -> text` | `call ash.run_rotate()` |
| `ash.rollup_minute(batch_limit int default 60) -> int` | `call ash.run_rollup_minute(batch_limit int default 60)` |
| `ash.rollup_hour() -> int` | `call ash.run_rollup_hour()` |
| `ash.rollup_cleanup() -> text` | `call ash.run_rollup_cleanup()` |

- Procedures are thin wrappers with **identical semantics** to the functions. The function's return value is surfaced as a `NOTICE` (a procedure cannot return it), so pg_cron/`psql` logs keep the same diagnostic content.
- `ash.start()` schedules pg_cron jobs with `call ash.run_*()` instead of `select ash.*()`.
- Privileges mirror the wrapped functions exactly (revoked from `PUBLIC`, owner-only), and the procedures join `ash._admin_funcs()` so `grant_reader()` / `revoke_reader()` / the privilege-hardening tests keep covering the whole admin surface.
- The functions stay. This is additive, not a breaking rename.

## Explicitly out of scope

**Intra-procedure `COMMIT` chunking.** Transaction control is the other reason procedures exist, and chunked commits in `rollup_minute()` would cap transaction duration. But it changes atomicity around rotation and rollup — an area with a history of correctness bugs (#211, #162, #81) — and it interacts with the xact-level advisory locks those routines rely on (a `COMMIT` releases them mid-routine). That deserves its own design issue rather than riding along here.

## Acceptance criteria

- [ ] `call ash.run_take_sample();` inserts exactly one sample row, same as `select ash.take_sample()`.
- [ ] `call ash.run_rollup_minute();` / `run_rollup_hour()` / `run_rollup_cleanup()` / `run_rotate()` produce the same side effects as their function counterparts.
- [ ] `ash.run_rollup_minute(batch_limit)` honours the batch limit.
- [ ] Each procedure raises a `NOTICE` carrying the wrapped function's return value.
- [ ] `ash.start()` registers pg_cron commands containing `call ash.run_`, and those jobs actually fire and produce samples/rollups.
- [ ] `select ash.run_take_sample()` fails (it is a procedure) and `call ash.take_sample()` fails (it is a function) — the two surfaces stay distinct and unambiguous.
- [ ] Procedures are not executable by `PUBLIC`; a reader role granted via `ash.grant_reader()` cannot `CALL` them.
- [ ] README documents the procedure surface and *why* it exists.

## Test plan

Behavioural assertions in CI (exact values, not catalog existence):

1. Count `ash.sample` rows, `call ash.run_take_sample()`, assert the count increased by exactly 1 and the new row's slot matches `ash.current_slot()`.
2. Equivalence: run `ash.rollup_minute()` (function) on one window and `call ash.run_rollup_minute()` on an equivalent window over an identical fixture, and assert the resulting `ash.rollup_1m` rows match in the aggregated columns — proving the procedure is not a different code path.
3. Assert `prokind = 'p'` for the five new routines **and** assert the negative resolution cases above actually error.
4. Assert `ash.start()`'s scheduled `cron.job.command` values contain `call ash.run_`, then assert a job fires and lands rows.
5. Privilege test: a non-owner reader role gets `42501` on every `CALL`.

Run locally against the development installer before pushing, per CLAUDE.md.

NikolayS:
Closes #221.

## Why

Query routers, connection poolers, and application-level load balancers commonly decide where to send a statement by its **kind**, not its content. Every pg_ash writer is syntactically a `SELECT`:

```sql
select ash.take_sample();   -- INSERTs
select ash.rotate();        -- TRUNCATEs
select ash.rollup_minute(); -- INSERT/UPDATE
```

so on a read-routing deployment they land on a read-only standby and fail with `cannot execute INSERT in a read-only transaction`. It works on a single-node install and breaks only where reads are routed elsewhere, which makes it a nasty topology-dependent failure.

`CALL` is unambiguously a write — no router can mistake it for a read — and a procedure is the semantically correct construct for a routine invoked for its side effect.

## What

| Function | Procedure form |
| --- | --- |
| `ash.take_sample()` | `call ash.run_take_sample()` |
| `ash.rotate()` | `call ash.run_rotate()` |
| `ash.rollup_minute(batch_limit)` | `call ash.run_rollup_minute(batch_limit)` |
| `ash.rollup_hour()` | `call ash.run_rollup_hour()` |
| `ash.rollup_cleanup()` | `call ash.run_rollup_cleanup()` |

Thin wrappers, identical semantics, wrapped return value surfaced as a `NOTICE`. The existing functions are untouched — this is additive.

**Why the `run_` prefix.** A procedure cannot share a callable name with a function. Verified on PG 18.4: same name + same argtypes is rejected outright; same name + different argtypes creates fine but then makes *both* `CALL` and `SELECT` fail with `is not unique`. So distinct names are forced, and a mechanical prefix beats per-name invention — it also makes a "no raw `ash.` SQL outside the wrapper" lint rule trivial for callers who want one.

**Privileges.** Procedures default to EXECUTE-for-PUBLIC, so this would otherwise have been a privilege hole. The hardening block now covers `prokind = 'p'` and uses `ON ROUTINE` — `ON FUNCTION` *rejects* a procedure with `is not a function`. The five names join `ash._admin_funcs()`; `grant_reader`/`revoke_reader` keep their `prokind = 'f'` filter, so the reader bundle is unchanged.

**Scheduling + a bug found along the way.** `ash.start()` now schedules the `CALL` forms. Its "already exists" branches for `ash_sampler` and `ash_rotation` previously re-synced only the *schedule*, never the command — so an upgraded installation would have silently kept running the legacy `SELECT` writer forever. Both now pass `command :=` as well. A new installer re-apply migration rewrites the five known legacy commands in place, matching on **exact** command text so an operator's customised command survives.

## Deliberately out of scope

Intra-procedure `COMMIT` chunking, tracked as #228. These routines rely on transaction-scoped advisory locks that a `COMMIT` would release, and this is the area with the worst bug history in the project (#211, #162, #81). It needs a written design, not a ride-along.

## Manual verification (PG 18.4 + pg_cron, Docker)

- Legacy jobs migrate on installer re-apply, schedules preserved (including `1 seconds`), operator's custom job untouched:
  ```
  ash_custom_probe | * * * * * | select ash.rollup_minute() /* operator custom */
  ash_rollup_1m    | * * * * * | call ash.run_rollup_minute()
  ash_sampler      | 1 seconds | set statement_timeout = '500ms'; call ash.run_take_sample()
  ```
- pg_cron actually fires the procedure at 1 Hz: 20 rows in `ash.sample` after 20 s, `cron.job_run_details` reports 26 runs, all `succeeded` with `return_message = CALL`.
- Reader role: `42501` on all five `CALL`s, while `ash.periods()` still works.
- Mis-resolution raises `42809` both ways (`select ash.run_take_sample()` → "is a procedure"; `call ash.take_sample()` → "is not a procedure").

Note: my original acceptance criteria said `42883` for that last case. That was wrong — the real SQLSTATE is `42809` (`wrong_object_type`), caught by running it, and the tests assert the correct value.

## Tests

`devel/tests/procedures_call_surface.sql` asserts exact values and real side effects: a tagged workload backend so `take_sample` has something to sample and the row delta is exactly 1; function-vs-procedure aggregate equivalence on an identical fixture (proving the procedure is not a divergent code path); `batch_limit = 1` advancing exactly one grain with the third grain left untouched; exact `rollup_1h` aggregates including `minute_counts`; retention post-conditions for cleanup; `rotate` advancing the slot by one and truncating the expected `sample_N`/`query_map_N`; both mis-resolution SQLSTATEs; reader denial plus an assertion that no procedure is executable by PUBLIC. The pg_cron steps assert exact `cron.job.command` strings and that a fired job produced rows, and there is a legacy-command migration test.

`devel/tests/features_lifecycle.sql` pins the exact schedule/command/database/active state of all five cron jobs; that pin moves to the `CALL` forms.

## Schedulers: both driven paths are covered

pg_ash has exactly **two** ways to be driven, and only one of them is an extension:

| path | how a job is registered | covered by |
| --- | --- | --- |
| **pg_cron** | `ash.start()` calls `cron.schedule()` with a command string | exact `cron.job.command` asserts + a fired job that lands rows |
| **external scheduler** (system cron, systemd timer, psql loop) | nothing to register — `ash.start()` **prints** the jobs for the operator | new asserts in `devel/tests/degraded_no_cron.sql` + guidance greps |

pg_cron is the only scheduler **extension** pg_ash supports: `grep -rioE "pg_timetable|pgagent|pg_cron"` over the tree returns 444 hits, all pg_cron, and `git log --all -S"pg_timetable"` / `-S"pgagent"` return nothing — no such integration has ever existed here.

pg_cron executes `CALL` without issue (verified live: `cron.job_run_details.return_message = "CALL"`). For the external path there is no catalog to inspect, so the **printed guidance is the registration surface** and is now asserted like one — five positive greps pinning the `CALL` forms and one negative grep proving the legacy `select ash.take_sample()` recommendation is gone. The cron-off axis also now runs the procedures for real against a tagged fixture backend, asserting exact row counts.

Replica behaviour is identical under both, because the guard lives in the routines rather than the scheduler: `ash.start()` raises `25006` on a standby so nothing is ever registered, a job left over from before a demotion no-ops with a NOTICE, and sampling resumes after promotion. That is #239's test, and its seam step runs on **every** matrix cell including cron-off.

## Red/green evidence

Run against real PostgreSQL 18.4 in Docker.

**RED — `devel/tests/procedures_call_surface.sql` against the pre-#221 installer (`sql/ash-install.sql`):**

```
psql:devel/tests/procedures_call_surface.sql:53: ERROR:  procedure ash.run_take_sample() does not exist
LINE 1: call ash.run_take_sample()
HINT:  No procedure matches the given name and argument types.
CONTEXT:  PL/pgSQL function inline_code_block line 8 at CALL
```

**RED — value-level, not just existence.** The scheduled command assertion against baseline `ash.start('2 seconds')`:

```
 ash_sampler   | set statement_timeout = '500ms'; select ash.take_sample()

ERROR:  ash_sampler command expected the CALL form, got
        'set statement_timeout = ''500ms''; select ash.take_sample()'
```

**RED — external path, pg_cron dropped:**

```
ERROR:  procedure ash.run_take_sample() does not exist
CONTEXT:  PL/pgSQL function inline_code_block line 64 at CALL
MISSING: "call ash.run_take_sample()" (for per-second, use a loop)
MISSING: Also schedule CALL ash.run_rotate() at the rotation_period
legacy SELECT form still recommended
```

**GREEN — same files, this branch:**

```
NOTICE:  Issue #221 CALL-able procedure tests PASSED
NOTICE:  All degraded-mode (no pg_cron) tests PASSED
  OK  "call ash.run_take_sample()" (for per-second, use a loop)
  OK  CALL ash.run_take_sample() \watch 1
  OK  execute "CALL ash.run_take_sample()" in a loop
  OK  Also schedule CALL ash.run_rotate() at the rotation_period
  OK  Schedule rollup: CALL ash.run_rollup_minute() every minute
  OK  legacy SELECT form absent
```

Honest note on ordering: the commits do not show test-then-implementation. I wrote the tests alongside the implementation and validated behaviourally, then captured the RED above afterwards by running each test file against the pre-change installer. The red output is real and reproducible, but the commit order does not demonstrate test-first, and I am not going to claim otherwise.

## Real delta

GitHub shows this PR against `main`, so its diff currently also contains the ~7900-line baseline copy (#245) and the parent PRs in the stack. **This PR's own change is:**

```
7 files changed, 1664 insertions(+), 83 deletions(-)
```

```bash
git diff origin/main...feat/call-procedures-221
```

`base` is deliberately left at `main` rather than retargeted to `chore/devel-baseline`. `test.yml` and `demo.yml` trigger on `pull_request.branches: [main, 'release/**']`, so a PR based on a `chore/` or `feat/` branch would get **no CI run at all** — the same silent outage as #237, where a PR shows green checks while the suite never ran. Once the parents merge, this diff sheds them automatically with no retargeting needed.

## Process notes

- `sql/` is frozen between releases, so all SQL is staged in `devel/sql/ash-install.sql`. The **first commit is a byte-identical copy** of `sql/ash-install.sql` opening the post-2.0-beta1 cycle — review from the second commit onward.
- One of four stacked PRs. Merge order: **#238 (CI headroom) → this → #239 (#222) → #240 (#224)**. This PR's diff therefore also contains #238's commits.
- All four target `main` deliberately: the workflows trigger only on PRs targeting `main` (`on.pull_request.branches: [main]`), so a non-`main` base silently skips the entire test suite — the same class of invisible CI outage as #237.
- Not merged, per CLAUDE.md — REV review and owner approval still outstanding.


NikolayS:
## Review verdict: **PASS after fixes** — 2 majors found and fixed

A five-role REV review (security, bug-hunter, test-analyzer, guidelines, docs) ran against this branch, verified against real PostgreSQL rather than by reading. It returned **FAIL** with two majors. Both were genuine, both are now fixed with red/green evidence, and the fixes are in `fix: address review findings on the CALL surface (#221)`.

### Major 1 — the legacy-command migration missed the v1.0/v1.1 uppercase forms

`sql/ash-1.0.sql` and `sql/ash-1.1.sql` scheduled `'SELECT ash.take_sample()'` and `'SELECT ash.rotate()'` in **uppercase**. `ash-1.1-to-1.2.sql` normalized only the sampler, and no released migration has ever rewritten the rotation command — so any installation that started at 1.0 or 1.1 still carries the uppercase text. The map joined on exact lowercase strings and skipped them entirely. That is precisely the population this feature exists for.

**RED** (installer re-applied with v1.0-era jobs seeded):
```
ash_rotation  | SELECT ash.rotate()
ash_sampler   | SELECT ash.take_sample()
```
**GREEN** (uppercase rows added):
```
ash_rotation  | call ash.run_rotate()
ash_sampler   | set statement_timeout = '500ms'; call ash.run_take_sample()
ash_rollup_gc | select ash.rollup_cleanup() /* site-local */    <- customised, untouched
```

### Major 2 — `ash.start()`'s sampler lookup was cluster-wide

`cron.job` is cluster-wide. The rotation lookup beside it has always filtered on `database` and `username`; the sampler lookup did not. Previously that only reset another database's schedule — but adding `command :=` turned it into **overwriting another database's command** while this database still ended up with no sampler at all. Filtered both, plus `ash.stop()`'s two lookups, which were capable of unscheduling another database's jobs.

### Also fixed
- The migration loop now downgrades a `cron.schedule()` failure to a warning. It fires for essentially every upgrading installation, and pg_cron revokes parts of its API from PUBLIC, so a restricted installing role could otherwise turn a routine re-apply into a **failed upgrade**.
- Test-analyzer found three assertions that could pass while the behaviour was broken: the migration test only ever seeded the exact lowercase strings (join condition never exercised on its false branch, custom-command preservation never proven), the reader-denial block had no positive control (`42501` is also what you get if `grant_reader()` silently did nothing), and the external-path rollup checks asserted only a negative invariant that holds trivially if the procedures do nothing. All three strengthened; the rollup checks now seed a complete minute two hours back and assert the exact grain.
- Docs: README claimed "each side-effecting collection or maintenance function" has a procedure form — false, since `start`/`stop`/`rebuild_partitions`/`uninstall` are side-effecting and deliberately have none. RELEASE_NOTES now discloses that `ash.start()` re-syncs the sampler and rotation commands.

### Findings accepted but not actioned
- The per-call `raise notice` in each procedure: confirmed harmless (pg_cron's notice receiver is an empty stub, and bgworker mode drops NOTICE at default `log_min_messages`), so no change.
- The ACL snapshot/restore block excludes `prokind='p'`: a no-op today because procedures are never dropped by the installer. Left alone rather than adding untested churn.

<details>

- The cycle-opening commit is byte-identical to `sql/ash-install.sql`, verified by `diff` rather than trust.
- Security: **no findings**. All five procedures are `postgres=X/postgres` on fresh install *and* re-apply; no PUBLIC grant, no NULL ACL. `grant_reader()` grants zero procedures. A hypothetical future procedure omitted from `_admin_funcs()` still lands revoked-from-PUBLIC via the second loop — safe default. `ON ROUTINE` is correct for `prokind in ('f','a','p')`; `format('%I(%s)')` is not injectable.
- pg_cron's `ScheduleCronJob` `ON CONFLICT … DO UPDATE` does not touch `active`, `nodename`, or `nodeport`, so the migration's claim to preserve cadence, jobid, active state and the socket-mode `nodename=''` fix is correct.
- The migration is idempotent by construction; ordering against the pre-existing `ash_rollup_1h` schedule migration is correct in either direction.
- `batch_limit` default and `search_path` match the wrapped functions exactly; `assert <null>` fails in PL/pgSQL, so the `cron.job` command asserts cannot pass on a missing job.
- Guidelines: dev-cycle staging correct, `sql/` untouched, version not bumped, Conventional Commits, no external attribution anywhere in the diff.

**Note on the CI rollup artifact:** this repo's PRs always carry one `SKIPPED` (the path-gated duplicate "capture + reel") and previously a `NEUTRAL` CodeQL row. A gate that requires `success == total` reads those as failures. Judge by the `test (…)` matrix and `ci-required`, not by a raw rollup.
</details>

NikolayS:
## samorev gate: **1 real finding, 1 false positive, 1 fair docs point**

Run against head `41bc05e05310205fd1e1ab8221e7aa0a33e25f5f`. This is a second, independent gate on top of the five-role review already in this thread — I read that review first, and I call out below where the two disagree and who is right.

**CI is green.** 16 success, **0 failure, 0 pending**. The single non-`success` row is `capture + reel (real PostgreSQL)` with conclusion `skipped` — the path-gated duplicate invocation. samorev's gate requires `success == total` and therefore reports `ci_status=unknown` and raises it as CRITICAL. **That CRITICAL is the known rollup artifact, not a finding about this PR**, exactly as the existing review's closing note says. Every PR in this repository trips it.

```text
```

## Adjudication of the substantive findings

### HIGH "explicit EXECUTE grants on procedures are destroyed on every re-apply" — **false positive**

samorev is right that the #107 ACL snapshot filters `prokind in ('f', 'a')` and so never captures procedures. It is wrong about the consequence, and the existing review's decision to leave it alone was correct.

The snapshot exists to survive a **drop**. All five procedures are created with `create or replace procedure` (`devel/sql/ash-install.sql:3782, 3797, 3812, 3827, 3842`) and the installer never drops them — and `CREATE OR REPLACE` preserves the existing ACL. Measured on PostgreSQL 18.4 rather than asserted:

```text
A. explicit grant made                        | t
B. after CREATE OR REPLACE (public revoked)   | t     <- ACL survives
C. after DROP + CREATE (public revoked)       | f     <- ACL lost
```

(The `public` revoke in B and C matters: without it the default PUBLIC grant makes every row `t` and the test proves nothing.)

So an operator's `grant execute on procedure ash.run_take_sample() to scheduler_role` survives a re-apply today. No action needed.

One caveat worth recording rather than acting on: that safety is **contingent on procedures never being dropped**. The day a procedure's signature changes, a drop becomes necessary and this gap becomes real and silent. The existing review's "no-op today … rather than untested churn" reasoning holds, but it is worth a comment at the snapshot block noting the assumption.

### MEDIUM "`ash.start()` clobbers customised sampler/rotation commands" — **real, and worth fixing**

Confirmed. The already-exists branches pass `command :=` unconditionally:

```sql
perform cron.alter_job(
  job_id := v_sampler_job,
  schedule := v_schedule,
  command := 'set statement_timeout = ''500ms''; call ash.run_take_sample()'
);
...
status := 'already exists — schedule updated to %s; command re-synced'
```

and likewise `command := 'call ash.run_rotate()'` for rotation.

What makes this more than a nitpick is that it contradicts the PR's own thesis in two places:

1. **The migration in this same PR goes out of its way to preserve custom text** — that is Major 1's headline evidence, `select ash.rollup_cleanup() /* site-local */  <- customised, untouched`. The hourly-rollup block even carries the comment *"Reusing the stored command also preserves custom command text"*. So the PR preserves customisation in one path and destroys it in another.
2. The release note's remedy — "apply your customisation after the last `ash.start()`" — is not a stable contract, because `start()` is re-entrant by design and runbooks routinely suggest calling it.

samorev's proposed fix is the right shape: in the already-exists branch, rewrite the command only when it matches one of the known legacy/canonical strings (reuse the migration's exact-match list), otherwise `raise notice` that a custom command was preserved. That would make the two paths agree. Worth a test: customise `ash_sampler`, call `ash.start()`, assert the custom command survives.

### LOW "per-call NOTICE is undocumented" — **fair, docs only**

The existing review established the NOTICE is *harmless* (pg_cron's notice receiver is an empty stub; bgworker drops it at default `log_min_messages`). samorev is making a different point: the README's documented external loop runs `call ash.run_take_sample()` once a second, and each call emits a NOTICE to stderr. Harmless is not the same as unsurprising. One sentence in the Scheduling section giving `set client_min_messages = warning;` would close it.

## Verdict

CI green. The gate's CRITICAL is the rollup artifact. Of the three substantive findings, one is a false positive I disproved empirically, one (`start()` clobbering custom commands) is real and I recommend fixing before merge, and one is a small docs gap.

<details>

## samorev Code Review Report

- **PR:** NikolayS/pg_ash#235 - feat: add CALL-able maintenance procedures (#221)
- **Author:** @NikolayS
- **AI-Assisted:** Yes

| Pipeline | Coverage |
|----------|----------|
| unknown | Not reported |

---

### BLOCKING ISSUES (4)

**HIGH** [bugs] Explicit EXECUTE grants on the new procedures are destroyed on every installer re-apply. The #107 ACL-preservation snapshot at the top of the installer filters `prokind in ('f', 'a')`, so a grant on `ash.run_take_sample()` etc. (exactly the "explicit minimal grants" the README tells operators to make for their scheduler role) is never captured and never restored after the drop/recreate. A dedicated scheduler/maintenance role silently loses access on the next upgrade — the precise failure #107 was meant to close, now reopened for the surface that is *designed* to be driven by a non-superuser role.
> `where nsp.nspname = 'ash' and proc.prokind in ('f', 'a') and acl.privilege_type = 'EXECUTE'` (devel/sql/ash-install.sql, ACL snapshot block) — procedures (`prokind = 'p'`) excluded; the restore block (content omitted for length) would also need `GRANT ... ON ROUTINE`/`ON PROCEDURE` since `ON FUNCTION` rejects procedures, as the PR itself notes.
> **Fix:** Extend the snapshot to `prokind in ('f', 'a', 'p')` and make the restore emit `grant execute on routine ash.%I(%s)` (or branch on prokind). Add a re-apply test: grant EXECUTE on `ash.run_take_sample()` to a role, re-run the installer, assert `has_function_privilege(role, 'ash.run_take_sample()', 'execute')`.

**MEDIUM** [bugs] Contradictory customisation guarantees. The new legacy-command migration deliberately matches exact text "so an operator's customised command survives", but the same PR makes `ash.start()` unconditionally overwrite `command` for `ash_sampler` and `ash_rotation` in the "already exists" branch. Any operator who customised those two jobs (e.g. a different `statement_timeout`, an added `set search_path`, a wrapper) has it clobbered on the next `ash.start()` — which `ash.status()`/runbooks routinely suggest calling. The release note documents this as "apply your customisation after the last `ash.start()`", which is not a stable contract since `start()` is re-entrant by design.
> PR description: "Both now pass `command :=` as well"; RELEASE_NOTES.md: "`ash.start()` itself re-syncs the `ash_sampler` and `ash_rotation` commands to the canonical values, so a customised command should be applied after the last `ash.start()` call".
> **Fix:** In the "already exists" branch, only rewrite the command when it equals one of the known legacy/canonical strings (reuse the same exact-match list as the migration); otherwise `raise notice` that a custom command was preserved. Add a test: customise `ash_sampler`, call `ash.start()` again, assert the custom command survives.

**LOW** [docs] Each procedure surfaces the wrapped return value as a `NOTICE`. In the documented external loop (`psql -qAtX ... -c "call ash.run_take_sample()"` every second) this emits a NOTICE line to stderr once per second, and under pg_cron bgworker mode it goes to the server log if `log_min_messages` is lowered. README/RELEASE_NOTES do not mention this or how to silence it (`-c "set client_min_messages = warning; call ..."`).
> PR description: "wrapped return value surfaced as a `NOTICE`"; README external loop example gives no `client_min_messages` guidance.
> **Fix:** Document the NOTICE and the silencing option in the Scheduling section, or emit at `DEBUG1` and leave the function form for interactive use where the return value is visible anyway.

---

### Summary

| CI/Pipeline | 1 | 0 | 0 |
| Security | 0 | 0 | 0 |
| Bugs | 0 | 2 | 0 |
| Tests | 0 | 0 | 0 |
| Guidelines | 0 | 0 | 0 |
| Docs | 0 | 1 | 0 |
| Metadata | 0 | 0 | 0 |

Note:

<details>

```text
provider=github
kind=pr
project=NikolayS/pg_ash
number=235
target=github:NikolayS/pg_ash#235
state=OPEN
draft=false
diff_lines=9463
diff_added=9214
diff_removed=45
diff_bytes=363801
comments_count=1
commits_count=8
ci_status=unknown
blocking=true
samorev_review=1
review_round=1/10
escalated=false
posted_by=local
no_comment=true
live_posting=not-run
defender_dropped=1
```

**Contested / dropped**

- [tests] No test covers grant preservation for the procedures across installer re-apply, and no test covers `ash.start()` being re-run against a customised `ash_sampler`/`ash_rotation` job. `procedures_call_surface.sql` asserts reader denial and "no PUBLIC execute" but not that an explicitly-granted scheduler role keeps access after re-apply — the exact scenario the README now recommends. — The PR already exercises both scenarios — the installer's #107 ACL snapshot/restore block is unchanged and (per the snapshot filter `prokind in ('f','a')`) admin routines are intentionally never restored by design, and the workflow diff adds an explicit `ash.start()` re-run against drifted `ash_sampler`/`ash_rotation` jobs asserting exact "command re-synced" statuses and command text, so the requested coverage is either present or contrary to documented hardening policy.

</details>

---

</details>

NikolayS:
## Fresh review round 2: **PASS**

A second independent five-role REV review ran against the post-fix head, verified against real PostgreSQL 18 + pg_cron 1.6. **No blockers, no majors.** It confirmed both earlier fixes are correct and complete, and checked the uppercase map by enumerating `sql/` rather than trusting my commit message:

> The complete historical set of command strings for the five `ash_*` job names is exactly seven … All seven are in the map. There are no semicolon, whitespace, or `statement_timeout` variants anywhere in the released tree.

It also confirmed the filters introduce **no regression**: pg_cron's `cron_job_policy` RLS already filtered by `username` for non-superusers, so their behaviour is byte-identical, and pg_cron only permits the extension in `cron.database_name`, so `database` always matches.

### Two vacuous assertions it found — both fixed

Fixed in `test: make the external-path cleanup and boundary checks able to fail`.

- **The cleanup check could not fail.** Nothing in the fixture was expired, so `run_rollup_cleanup()` reported `deleted 0 minute rows` and the assertion passed against a procedure that does nothing. Now seeds a grain past `rollup_1m_retention_days` and requires it deleted while the two-hour-old grain survives. Proven RED-able by stubbing out `rollup_cleanup()`:
  ```
  ERROR: external path: run_rollup_cleanup left 1 expired minute grain(s)
  ```
- **The "open minute is never folded" invariant could not fail either.** `run_rollup_minute()`'s default `batch_limit` of 60 stops a full hour short of now from a two-hour-old watermark, so the current minute was never even a candidate. The watermark is now advanced to just behind now before the invariant is asserted, making it the boundary check it claims to be.

### Findings acknowledged, not actioned
- **`ash.status()`'s cron listing is unfiltered**, so a superuser sees jobs it does not own. Real and worth fixing, but pre-existing and not introduced here — better as its own change than widened into this diff.
- **A second superuser running `ash.start()` creates a duplicate job set.** The reviewer's own analysis notes this already affected four of the five jobs before this PR, so the filter makes the sampler *consistent* rather than introducing it. A warning on a foreign-owned same-named job is a good idea for a follow-up.
- **The rationale in one commit message overstated reachability** — pg_cron refuses `CREATE EXTENSION` outside `cron.database_name`, so the two-databases scenario needs `cron.schedule_in_database()`. The filter is still correct, and the reviewer confirmed it removes a *real* latent bug: pre-fix, `ash.stop()` could match a foreign-owned job and then raise an unhandled `could not find valid entry for job`.
- Round 2 not asserting `schedule`/`active` (round 1 does), and the README caveat living only in RELEASE_NOTES — both minor, noted for follow-up.

<details>
<summary>Reviewer's clean list</summary>

- Uppercase map completeness verified by enumerating every `cron.schedule` call and `command` assignment across all 16 files under `sql/`.
- Every `cron.job` lookup in the installer is now filtered except `status()` (reported above).
- The exception handler is scoped to the single `cron.schedule()` call; the warning format string has exactly three `%` placeholders for three arguments, executed against PG18 to confirm rendering. `exception when others` does not catch `query_canceled`/`assert_failure`.
- Round 2 is genuinely behavioural and RED-able: removing either uppercase map row makes the corresponding assert fail. It cleans up after itself.
- The positive control is load-bearing: `perform ash.periods()` runs outside any exception block, so a `grant_reader()` no-op aborts the block rather than letting the five denials pass vacuously.
- The degraded rollup fixture is deterministic at every clock position — the reviewer worked through the `v_minute = H0` and minute-of-hour 59 boundary cases.
- Security and guidelines: **no findings**.

**CI rollup artifact:** one `SKIPPED` row (the path-gated duplicate "capture + reel") is normal here; judge by the `test (…)` matrix and `ci-required`.
</details>

NikolayS:
## samorev re-run on `78c67a6` — **previous finding verified fixed**

> Supersedes my report of 2026-08-21, which was against head `41bc05e`. That head is gone; treat this as the current one.

**CI green:** 16 success, 0 failure, 0 pending. One `skipped` row (`capture + reel`, the path-gated duplicate) still makes samorev's `success == total` rule report `ci_status=unknown` and raise CRITICAL. Same artifact as always, not a finding.

## Verification of the earlier findings

### MEDIUM `ash.start()` clobbers customised commands — **fixed, and fixed properly**

This was the one real finding last time. `73c2be9` + `ee594e5` + `78c67a6` close it, and they took the better of the two available routes rather than the cheap one.

The already-exists branch no longer rewrites unconditionally. It resolves the stored command through a shared map and only re-syncs a *recognised* one:

```sql
select cron_job.command into v_stored_command from cron.job where jobid = v_sampler_job;
v_canonical_command := ash._canonical_job_command('ash_sampler', v_stored_command);

if v_canonical_command is null then
  perform cron.alter_job(job_id := v_sampler_job, schedule := v_schedule);   -- schedule only
  raise notice 'ash.start: ash_sampler has a customised command; preserving it. ...';
  status := format('already exists — schedule updated to %s; custom command preserved', v_schedule);
else
  perform cron.alter_job(job_id := v_sampler_job, schedule := v_schedule, command := v_canonical_command);
  status := format('already exists — schedule updated to %s; command re-synced', v_schedule);
end if;
```

Three things worth calling out, because they are what make this a real fix rather than a patch:

1. **The contradiction is actually resolved, not papered over.** `_canonical_job_command()` reads `ash._job_command_map()` — *the same map the installer migration uses* (`devel/sql/ash-install.sql:1976` defines it, `:2032` and `:8182` consume it). The two paths can no longer disagree about what counts as a customisation, which was the substance of the objection.
2. **The preservation NOTICE earns its place.** It does not merely say "preserved" — it warns that a surviving `select ash.take_sample()` command may be routed to a replica by a read-routing proxy now that pg_ash writes through `CALL`, and prints the recommended replacement. That is a hazard I had not raised.
3. **The schedule repair still happens for a customised job**, and there is an assertion pinning exactly that. That is the obvious over-correction — skipping the whole `alter_job` — and it is explicitly guarded against.

The test asserts both directions rather than only the new one: recognised legacy commands **are** re-synced, unrecognised ones are preserved with the matching status string, and the schedule is still repaired. That is the shape I asked for.

### HIGH procedure ACLs on re-apply — still a **false positive**, unchanged and correctly left alone

Re-confirmed on the new head: all five procedures remain `create or replace procedure` and are never dropped, so the ACL survives. My measured evidence from last time still applies (explicit grant → `t`; after `CREATE OR REPLACE` with PUBLIC revoked → `t`; after `DROP`+`CREATE` → `f`). No change was needed and none was made. Correct call.

### LOW per-call NOTICE documentation

Not addressed, and still fair. Unchanged in severity: docs-only.

## New this round

**LOW [tests] — exact-count sampling assertion said to be racy in `degraded_no_cron.sql`.** samorev's stated mechanism does not hold. Its premise is that `procedures_call_surface.sql` sets `include_bg_workers = false` while `degraded_no_cron.sql` does not, leaving the latter exposed to autovacuum. But the column is declared

```sql
include_bg_workers  bool not null default false,
```

so the default is already `false` and both tests sample under identical effective settings. The explicit set in the other file is belt-and-braces, not a behavioural difference. `degraded_no_cron.sql` additionally asserts its fixture precondition (`exactly one tagged active backend`) immediately before sampling, and `ash.sample` stores one row per database rather than per backend.

Residual exposure is a *client* backend active in a different database between the precondition check and the sample — theoretical in CI. Optional hardening: set `include_bg_workers = false` explicitly so the test does not silently depend on a default that could change. Not blocking.

## Verdict

The one real finding from my previous report is fixed, verified on the new head, and the fix is better than what I proposed. Nothing new blocks. Remaining items are one docs nicety and one optional test hardening.

<details>

## samorev Code Review Report

- **PR:** NikolayS/pg_ash#235 - feat: add CALL-able maintenance procedures (#221)
- **Author:** @NikolayS
- **AI-Assisted:** Yes

| Pipeline | Coverage |
|----------|----------|
| unknown | Not reported |

---

### BLOCKING ISSUES (2)

**LOW** [tests] Exact-count sampling assertions are racy against transient background activity. Both new test paths assert that a sample pass writes exactly 1 row on the premise that the tagged fixture backend is the only active session, but the fixture check only proves the tagged backend exists (`count(*) = 1` filtered by `application_name`), not that nothing else is active. An autovacuum worker (or any other momentarily active session) at the sampling instant adds a row and fails `v_after = 1` / `count(*) = 1` — unless `include_bg_workers = false` is known to exclude `autovacuum worker` backend_type specifically, which the diff doesn't show. This would surface as a rare, hard-to-reproduce CI flake.
> `call ash.run_take_sample(); select count(*) into v_after from ash.sample; assert v_after = 1, ...expected exactly 1 sample row...` (degraded_no_cron.sql), and the equivalent `count(*) = 1 and min(slot) = v_slot` assert in procedures_call_surface.sql.
> **Fix:** Either confirm the sampler's backend filter excludes all non-client backend_types (autovacuum worker, parallel workers, walsender) and note that in the test comment, or make the assertion tolerant-but-precise: assert exactly one row whose sampled backend matches the tagged fixture, ignoring rows from other backend types.

---

### Summary

| CI/Pipeline | 1 | 0 | 0 |
| Security | 0 | 0 | 0 |
| Bugs | 0 | 0 | 0 |
| Tests | 0 | 1 | 0 |
| Guidelines | 0 | 0 | 0 |
| Docs | 0 | 0 | 0 |
| Metadata | 0 | 0 | 0 |

Note:

<details>

```text
provider=github
kind=pr
project=NikolayS/pg_ash
number=235
target=github:NikolayS/pg_ash#235
state=OPEN
draft=false
diff_lines=2166
diff_added=1686
diff_removed=81
diff_bytes=83183
comments_count=3
commits_count=10
ci_status=unknown
blocking=true
samorev_review=1
review_round=2/10
escalated=false
posted_by=local
no_comment=true
live_posting=not-run
defender_dropped=2
```

**Contested / dropped**

- [bugs] Cross-database safety is confirmed only for the `ash_sampler` lookup in `ash.start()`; the installer re-apply migration and the other job lookups are trimmed from the diff, and if any of them match `cron.job` rows cluster-wide by `(jobname, command)` alone, they would rewrite another database's pg_ash job to `call ash.run_...()` — breaking that database if it still runs a pre-#221 pg_ash without the procedures. The diff's own comment documents exactly this failure class ("cron.job is cluster-wide, so an unfiltered jobname lookup finds another database's sampler") but only shows the sampler fix; the rotation lookup is said to have always filtered, while the `ash_rollup_1m`/`ash_rollup_1h`/`ash_rollup_gc` branches and the re-apply migration block are content omitted for length. Manual verification in the PR covers a single-database cluster only. — The diff visibly refutes the concern — all three rollup lookups and the installer re-apply migration filter on `database = current_database() and username = current_user`, and CI now tests the migration with those filters asserted.
- [tests] The customised-command preservation contract — the central promise of `_job_command_map()` ("A stored command absent from this map is operator-customised and must never be rewritten", also promised in README and RELEASE_NOTES) — appears to be verified only manually. The PR's test list mentions a "legacy-command migration test" (rewrite happens) but the manual-verification section is the only place the "operator's custom job untouched" behaviour is shown, and no automated assertion for it is visible in the included test content (procedures_call_surface.sql is trimmed, so it may exist there — if so, ignore this). This is the single most regression-prone behaviour in the change: one wrong `where` clause silently clobbers operator customisations on the next `ash.start()` or installer re-apply. — The behaviour is already automatically asserted in the diff — test.yml Round 2 proves a customised command (`select ash.rollup_cleanup() /* site-local tweak */`) survives installer re-apply byte-identically, and Round 3 plus Case 2 assert ash.start() preserves customised commands (including the `select 1` case) with the "custom command preserved" status — so the finding's own escape hatch ("if so, ignore this") applies.

</details>

---

</details>

Issue #229, opened 2026-08-21
Feature: periodic AAS digest to the Postgres log, for when the database itself is unreachable


NikolayS:
## Problem

pg_ash stores its history *inside* the database it observes. When that database is down or unreachable, the data describing why it went down is unreachable too — precisely when it is needed. There is also no path to the numbers that does not involve querying the instance.

A cheap mitigation: periodically write a coarse AAS digest to the **Postgres log file**. The log is already collected, rotated, and shipped by every serious deployment, so the digest inherits an entire retention and transport story for free, and it survives the instance that produced it.

The digest is deliberately coarse. That tradeoff is worth making: partial visibility after an incident beats none.

## Proposed design

New admin function, e.g.:

```sql
ash.log_digest(since => now() - interval '1 hour', vcpus => null)
```

- Emits one `RAISE LOG` line per call with a compact AAS digest: window bounds, `aas_avg`, `aas_worst1m`, `aas_p99`, the wait-class breakdown, and the top wait event.
- Single line, stable prefix (e.g. `pg_ash digest:`), payload as JSON so the line is both greppable and parseable. A multi-line dump is unusable in a log pipeline.
- Reuses `ash.report()` rather than reimplementing the math, so the digest and the JSONB payload can never disagree.
- Honest about coverage: if the window has no rollup coverage, log that fact rather than a fabricated zero. A digest reporting `aas_avg = 0` for an uncovered window is worse than one that logs nothing.
- `vcpus` optional; when supplied, include the percent-of-vCPU normalization that makes the number judgeable without external context.
- Schedulable as an additional optional pg_cron job (hourly by default, opt-in), and callable manually or from an external scheduler for the no-pg_cron path.

Open design questions:

- Default on or off? Writing to the server log on every install is a visible side effect, so opt-in is the safer default — but an opt-in fallback is one the incident victim will not have enabled.
- `LOG` versus `WARNING` severity. The digest must land above typical `log_min_messages` without polluting; `LOG` is semantically correct but would **not** be written under the default `log_min_messages = warning`. This needs deciding deliberately rather than by accident.

## Acceptance criteria

- [ ] Function exists, emits exactly one log line per call.
- [ ] The line parses as JSON after a fixed prefix.
- [ ] Numbers match `ash.report()` for the same window and `vcpus`, exactly.
- [ ] Uncovered windows produce an explicit "no coverage" digest, never a fabricated zero.
[...]

Issue #227, opened 2026-08-21
Design: replica coverage via instance-labelled samples and a client-driven insert path


NikolayS:
## Motivation

Design issue — **recording the ask, not proposing to build it now.**

pg_ash is primary-only: `ash.take_sample()` reads the local `pg_stat_activity` and inserts into a local table, and `ash.sample` carries no host or instance identity. #222 makes that limitation honest (recovery guards, documented blind spot). This issue tracks the harder question of actually *closing* it.

On a deployment that routes reads to replicas, pg_ash sees writes, post-write sticky reads and default-consistency background jobs, but misses the bulk of ordinary read `SELECT`s. The reported workload mix is write-skewed, and the most common incident class on such installs — slow reads and replica saturation — is precisely what pg_ash cannot see.

The shape of a fix: an instance label on samples plus a client-driven insert path, so one sampler can poll each replica's `pg_stat_activity` and push rows into the primary's store. That is the only route to replica coverage that does not require pg_ash to write on a standby.

## What this would require

- **Schema**: a node/instance identity column on `ash.sample` (and a matching dimension in `ash.rollup_1m` / `ash.rollup_1h`, or the rollups stop being comparable across nodes). Node identity needs to survive failover sensibly — `system_identifier` plus a human label, rather than hostname alone.
- **A client-driven insert path**: today the only entrypoint takes its own snapshot. Replica coverage needs an entrypoint that *accepts* an already-captured `pg_stat_activity` snapshot from another node and stores it against that node's label, with the same wait-event interning and packing the local path uses.
- **Reader semantics**: every reader (`aas`, `top`, `timeline`, `report`, `compare`, `chart`) currently answers "what was this database doing". With multiple nodes they need a node filter and a defensible default — aggregate across the cluster, or the primary only. Getting this wrong silently doubles AAS.
- **Trust boundary**: a remote insert path lets a caller write arbitrary sample rows. It needs its own privilege, and the packed-array validation (`ash._sample_data_is_valid`) becomes a security control rather than a sanity check.
- **Sampler cadence skew**: per-node sampling intervals must be recorded per node, or AAS across nodes is computed from mismatched denominators (cf. #137).

## Why it is deferred

This touches the storage format, the rollup dimensions, every reader's semantics, and the privilege model at once — it is a major-version-sized change, not an increment. #222 plus documentation covers the near-term need to not mislead anyone.

## Open questions

- Do cross-node rollups aggregate, or stay per-node with aggregation at read time?
[...]

Issue #226, opened 2026-08-21
Benchmark pg_ash overhead at scale, including an already-saturated instance


NikolayS:
## Problem

Whether pg_ash can be enabled **by default** rather than opt-in is the question that decides most embedding scenarios. Opt-in observability is absent exactly when it is needed, because nobody turns it on before their first incident.

The current answer is a reasoned estimate, not a measurement:

- pg_ash adds **one transaction** on top of whatever throughput the database already serves. Below roughly 1000 TPS a database is barely loaded, and adding one is noise.
- Sample cardinality tracks the number of *active sessions*, not database size. At 50-100 active sessions and 1 sample/sec the compressed volume is on the order of kilobytes per second — megabytes per day, not gigabytes.

That reasoning is plausible and probably right, and it is still not something this project should assert in user-facing documentation unmeasured.

There is also one case the estimate explicitly does not cover: a database that is **already in trouble** — few cores, many active sessions — where the marginal cost of sampling is least predictable and most consequential. Flame graphs and a controlled benchmark can answer it; intuition cannot.

A second, distinct overhead axis has never been quantified at all: samples land in a **persistent regular table**, so they generate WAL and flow into backups and replication. That is a different cost from CPU and deserves its own number.

## Scope

A measurement task, not a code change. Filed so the resulting numbers become citable in the README instead of recalled from memory.

Matrix:

- Sampling cadence: 1s (default), 15s, off (control).
- Active sessions: ~10, ~50, ~100 concurrent.
- The saturation case: constrained vCPU (e.g. 8 cores) driven to ~100 active sessions.
- With and without pg_stat_statements loaded.

Metrics per cell:

- TPS and latency percentiles of the *workload*, pg_ash on versus off — the number that decides ship-by-default.
- CPU cost attributable to sampling (flame graph).
- WAL bytes/sec generated by pg_ash, and the resulting backup and replication delta.
- On-disk growth per day for raw samples and each rollup tier.
- `skipped_samples` — under saturation, does the sampler start losing ticks, and at what point? A sampler that silently degrades is worse than one that costs a little more.

## Acceptance criteria

- [ ] Results published in the repo (a `benchmarks/` writeup) with exact hardware, PG version, workload, and pg_ash config per cell.
- [ ] A stated overhead budget: measured worst-case cost as a percentage of workload TPS, and the conditions under which it holds.
- [ ] README gets a short, sourced "Overhead" statement replacing any unmeasured claim, linking to the writeup.
[...]

PR #236, opened 2026-08-21
docs: record pg_ash product goals

NikolayS:
## What

Adds `docs/GOALS.md`: what pg_ash is for, who it is for, and which constraints are non-negotiable.

## Why

Several of the project's defining decisions have no written rationale, so each design argument re-derives them:

- the anti-extension design is what makes embedding in another product possible at all — a vendor cannot ask each customer's DBA for privileges;
- "can this be **on by default**?" is the question that decides an embedding, because opt-in observability is absent exactly when it is needed;
- AAS is a machine-readable health metric, not merely a dashboard input, and averages alone lie (a 16-vCPU cluster averaging 4.9 AAS whose worst minute is 295);
- pg_stat_statements being *soft in a specific way* — query IDs survive, only text is lost — is load-bearing, not a detail;
- pg_cron is third-party and not universally available, so the external-scheduler path has to be first-class.

## Honesty

The page states the overhead budget as an **estimate** and says explicitly that it should not go into user-facing docs until measured (#226). It closes with a gaps table naming every place the product does not currently meet its own goals, each linked to its tracking issue (#221, #222, #223, #224, #225, #226, #227, #228, #229, #230).

## Scope

Docs only — one new file, no SQL, no changes to `sql/` or `devel/sql/`, no CI behavior change.

### DIFF
--- a/docs/GOALS.md
+++ b/docs/GOALS.md
@@ -1,0 +1,145 @@
+# pg_ash goals
+
+What pg_ash is for, who it is for, and which constraints are
+non-negotiable. This is the "why" behind the design decisions;
+`docs/RELEASE_PROCESS.md` covers the "how" of shipping them.
+
+## The problem
+
+When a self-managed Postgres instance has an incident, the operator usually has
+nothing to look at. Managed platforms answer this with a proprietary active
+session dashboard, but that answer does not travel. Self-hosted, on-prem, and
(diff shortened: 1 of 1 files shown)


NikolayS:

The product goals are being carried into #254 and reviewed alongside the actual supported behavior.

Issue #237, opened 2026-08-21
CI: test.yml is at 99.5% of the 512,000-byte workflow size limit, and crossing it silently disables the entire test suite


NikolayS:
## What happened

While adding tests for #221 I pushed a branch and every check reported green — but the **`Test pg_ash` workflow never ran at all**. No error, no annotation, no failed run. `gh pr checks` listed CodeQL and the demo workflow as passing, and nothing indicated that the entire test suite had silently stopped executing for that branch.

The cause: **GitHub Actions refuses to create a run for a workflow file larger than 512,000 bytes**, and does so silently.

```
main's .github/workflows/test.yml   509,209 bytes   (runs)
same file + the #221 tests inline   534,559 bytes   (never runs)
                        ceiling     512,000 bytes
```

The limit is **512,000 bytes (decimal), not 512 KiB.** I first assumed 524,288, trimmed the file to 518,629, and the workflow still did not run — which is how the decimal value got pinned down.

`main` is currently at **99.5% of the limit — 2,667 bytes of headroom.** Any test addition at all crosses it.

This is a dangerous failure mode for this project specifically, because CLAUDE.md makes CI green a merge precondition and a release-gate input. A workflow that does not run looks exactly like a workflow with nothing to report: PR checks are green, the release gate sees no failures, and the actual test suite has not executed. Every guarantee downstream of "CI is green" is void without anyone noticing.

## Immediate mitigation (already applied in #235)

Moved the new test body to `devel/tests/procedures_call_surface.sql` and invoked it with `psql --file` from a short workflow step. That is the pattern the repo already uses for `cron_path.sql`, `features.sql`, `degraded_no_pgss.sql`, `query_attribution_assert.sql`, `stale_rollup_source.sql` and `schema_snapshot.sql` — this change just applies it consistently. `test.yml` came back to 393,715 bytes and the workflow started running again.

PR #238 also adds a size guard, because a reprieve without one just defers the next silent outage.

## Proposed fix

1. **Add a CI guard that fails loudly when `test.yml` approaches the ceiling.** This is the important part: the guard must exist so the failure can never again be silent. A check in the `docs-lint` job, failing at (say) 480 KiB with a message pointing at `devel/tests/`, costs nothing and converts an invisible outage into an obvious red X. It cannot run in the workflow it guards once that workflow is too big to start — so it should live in a separate, small workflow file, or in the `demo` workflow, which is nowhere near the limit.
[...]

PR #239, opened 2026-08-21
feat: guard write paths and scheduling against running on a replica (#222)

### LINKED ISSUE #222: Feature: guard write paths and scheduling against running on a replica

NikolayS:
Closes #222.

## Why

pg_ash writes only on a primary, but nothing enforced that — `grep pg_is_in_recovery sql/ash-install.sql` returned nothing.

I built a real streaming standby and measured what actually happens. It is worse than "it errors":

| call on a standby | before |
| --- | --- |
| `take_sample()` | `ERROR: cannot execute INSERT in a read-only transaction` — naming `ash.query_map_0`, not the cause |
| `rotate()` | **returns success**: `skipped: rotated too recently at ...` |
| `rollup_minute()` | **returns `0`** silently |
| `rollup_hour()` | **returns `0`** silently |
| `rollup_cleanup()` | `ERROR: cannot execute DELETE in a read-only transaction` |
| `start()` | errors partway through, after touching pg_cron |
| `status()` | reports **`sampling_enabled = true`** |

Three silent paths that tell the caller everything is fine, and a `status()` that answers "is ASH running here?" with a confident yes. That is the bug; the error wording is secondary. Full transcript in [the issue](https://github.com/NikolayS/pg_ash/issues/222#issuecomment-5365332387).

## What

One predicate, `ash._in_recovery()`, applied two ways because the callers differ:

**Scheduled routines no-op cleanly.** `take_sample` / `rotate` / `rollup_minute` / `rollup_hour` / `rollup_cleanup` emit one actionable NOTICE and return their neutral value. A pg_cron job left behind on a demoted node must not spam an error every second, and a monitoring wrapper must not see a hard failure for an expected, transient condition.

**Explicit operator actions raise.** `start` / `stop` / `rebuild_partitions` / `uninstall` and state-changing `set_debug_logging` raise SQLSTATE `25006` **before touching any state**. `set_debug_logging(NULL)` is a read and still works. The installer refuses on a standby too, before its first write.

**`status()` tells the truth.** New `in_recovery` row, plus a NOTICE in recovery explaining that `sampling_enabled` is the primary's replicated configuration, not evidence of local sampling. Existing row values are unchanged. Readers are deliberately unguarded.

**The blind spot is documented.** README now states plainly that where reads are routed to replicas, pg_ash sees writes, sticky post-write reads and default-consistency background jobs, but not the ordinary read queries the replicas serve — so the workload mix reads write-skewed, and slow reads / replica saturation, the most common incident class on such deployments, is exactly what pg_ash cannot show. Better learned from the README than from a misleading chart. Per-replica coverage is tracked as #227.

## Tests — two parts, neither sufficient alone

Every guard calls `ash._in_recovery()` rather than the catalog function, which gives the tests a seam:

- **`#222: recovery guard coverage on every PostgreSQL version`** redefines that one function inside a transaction and runs `devel/tests/replica_guards_assert.sql` — every guard branch, on every supported major. `devel/tests/replica_guards_primary.sql` then runs after the rollback and asserts the override did not leak and that the primary still samples.
- **`#222: recovery guards on a real streaming standby`** runs *the same assertion file* against a genuine physical replica: a second postmaster inside the CI service container, built with `pg_basebackup` over the unix socket. It additionally asserts that the standby calls left the **primary's** `ash.sample` untouched, and that the installer exits non-zero with `25006` there.

The standby needs no server config change: the official postgres image already ships `local replication all trust`, and `wal_level = replica` / `max_wal_senders = 10` are defaults. Clearing `shared_preload_libraries` keeps pg_cron from starting on the standby. Gated to one matrix cell, since the seam step covers the majors.

Assertions check exact values and that all three tables are **completely unchanged** across every no-op — the real contract, not just "returned a neutral value".

## Manual verification (PG 18.4, real `pg_basebackup` standby)

```
_in_recovery()          = t
take_sample()           = 0
rotate()                = skipped: server is in recovery (standby); pg_ash writes only on a primary
rollup_minute()         = 0
rollup_hour()           = 0
rollup_cleanup()        = skipped: server is in recovery (standby); ...
start()                 -> 25006      stop()              -> 25006
rebuild_partitions      -> 25006      uninstall           -> 25006
set_debug_logging(true) -> 25006      set_debug_logging(null) works (read)
reader periods()        works
status(): in_recovery | true
installer on standby: ERROR: cannot install pg_ash on a standby ...
primary afterwards: in_recovery=false, take_sample=1
```

Both test files were then run as CI runs them — against the real standby, and via the seam plus rollback on the primary — and all three passed.

## Red/green evidence

Run on a **real streaming standby** (pg_basebackup, PostgreSQL 18.4), against the pre-#222 installer — which contains zero recovery checks of any kind (`grep -c "pg_is_in_recovery\|_in_recovery"` → `0`).

**RED — the assertion file itself:**

```
psql:/tmp/rg.sql:131: ERROR:  function ash._in_recovery() does not exist
CONTEXT:  PL/pgSQL function inline_code_block line 19 at ASSERT
```

**RED — the behaviour that actually matters, on the same standby.** Three of these do not error at all, which is the real defect:

```
NOTICE:  rotate()        -> skipped: rotated too recently at 2026-08-21 06:17:46+00
NOTICE:  rollup_minute() -> 0
NOTICE:  rollup_hour()   -> 0
NOTICE:  status.sampling_enabled -> true
```

A caller polling a standby is told rotation succeeded, the rollups had nothing to do, and sampling is enabled.

**GREEN — same assertion file, same real standby, this branch:**

```
NOTICE:  Issue #222 recovery-guard assertions PASSED
```

Honest note on ordering: the commits do not show test-then-implementation. I wrote the tests alongside the implementation and validated behaviourally on a real standby, then captured the RED above afterwards against the pre-change installer. The red output is real and reproducible, but the commit order does not demonstrate test-first, and I am not going to claim otherwise.

## Scheduler coverage

The guard lives in the routines, not the scheduler, so both driven paths behave identically: `ash.start()` raises `25006` on a standby under **either** pg_cron or the external-scheduler path, so nothing is ever registered; a job left over from before a demotion no-ops with a NOTICE; sampling resumes after promotion. The seam step deliberately carries **no** `matrix.cron` gate, so it runs on every cell including cron-off. (pg_cron is the only scheduler extension pg_ash supports — see #235.)

## Diff contents

#245 has merged, so the baseline copy is gone. GitHub shows this PR against `main`, so its diff still includes #235 (which merges first). **This PR's own change is:**

```
9 files changed, 712 insertions(+), 9 deletions(-)
```

```bash
git diff feat/call-procedures-221...feat/replica-guards-222
```

`base` is deliberately left at `main` rather than retargeted to `feat/call-procedures-221`. `test.yml` and `demo.yml` trigger on `pull_request.branches: [main, 'release/**']`, so a PR based on a `chore/` or `feat/` branch would get **no CI run at all** — the same silent outage as #237, where a PR shows green checks while the suite never ran. Once the parents merge, this diff sheds them automatically with no retargeting needed.

## Stacking

Third of four. Merge order: **#238 (CI headroom) → #235 (#221) → this → #240 (#224)**. This PR's diff therefore also contains the earlier PRs' commits; review the commits titled `feat: guard write paths...`, `test: account for ash._in_recovery()...` and `ci: tear the #222 fixture backend down...`.

All four target `main` deliberately. The workflows trigger only on PRs targeting `main` (`on.pull_request.branches: [main]`), so a non-`main` base silently skips the entire test suite — the same class of invisible CI outage as #237.

Not merged, per CLAUDE.md — REV review and owner approval still outstanding.


NikolayS:
## Review verdict: **PASS after fixes** — 2 majors found and fixed

A five-role REV review ran against this branch. It returned **FAIL** with two majors, both of the same shape: **the test passed with the guard deleted**. The production SQL was found correct; the failures were test-integrity ones. Both are fixed in `fix: make the recovery-guard assertions actually discriminating (#222)`.

### Major 1 — the `rollup_minute` / `rollup_hour` guard assertions were vacuous

Both routines return `0` whenever there is simply nothing to fold, and the seam step runs immediately after a fresh install, when `ash.sample` and `ash.rollup_1m` are empty. So `assert rollup_minute() = 0` was measuring an empty database, not the guard. Delete the guard and it still passed.

Fixed by seeding real pending work first — raw samples two hours back with the watermarks rewound — so an unguarded rollup would write rows. The assert file now also **checks that precondition and fails loudly** if the caller did not seed, so it cannot silently go vacuous again. The seam seeds inside its transaction; the standby step seeds on the primary before `pg_basebackup` so the work replicates.

Proven discriminating by restoring the **real** unguarded `rollup_minute`/`rollup_hour` bodies from the parent branch:
```
RED:   ERROR:  rollup_minute() on a standby must return 0, got 60
GREEN: NOTICE: Issue #222 recovery-guard assertions PASSED
```

### Major 2 — the installer standby-refusal check was fully vacuous

It grepped only for `25006`. With the refusal block deleted, the installer still dies at its very first statement — `create temp table _ash_install_func_acl` — with exactly that SQLSTATE. It was the **only** coverage of the installer guard. Now the guard's own message is pinned as well.

The same collision applies to the five administrative entrypoints: on a real standby an unguarded write raises `25006` from PostgreSQL itself. The assertions now pin the guard's wording alongside the SQLSTATE, which makes them discriminating in both the seam and the real-replica context, and pins the operator-facing message.

### Also fixed
- Standby teardown was best-effort and unverified. It now polls until the second postmaster on `:5433` is really down, fails loudly if not, and reclaims the basebackup directory. A silent failure would have left a live postmaster and walsender running for the remaining ~8000 lines of the job — the same leak class the fixture-backend teardown already hard-fails on.

### Findings accepted but not actioned, with reasons
- **`_in_recovery()` marked `stable` while `pg_is_in_recovery()` is volatile.** Correct observation. Left as-is: the `SET search_path` clause blocks inlining and per-statement re-evaluation is what the guards want. Worth revisiting if that clause is ever dropped.
- **`grant_reader`/`revoke_reader` have no standby guard.** True, and out of scope here — they are privilege helpers, not the scheduled/administrative surface this PR covers. Better as its own change than widened silently.
- **Readers on a standby are only spot-checked.** Now materially addressed by the stacked #240, which asserts four readers on a real standby including with an unlogged ring.

<details>

- Guard ordering verified at **all ten call sites**: each is the first statement in its block, before any config read, any write, any advisory lock, and before `rebuild_partitions`/`uninstall`'s `confirm` check.
- `set_debug_logging(NULL)` correctly remains a read.
- Installer refusal sits inside `begin;` and before the ACL snapshot.
- `ash.status()`: no existing metric's value, name, or computation changed; all existing status assertions are name-keyed, so the added row is safe.
- **Seam override cannot leak** — the reviewer tried to break it and could not: transactional DDL restores OID/prosrc/volatility/ACLs on rollback; if the assert file errors, `ON_ERROR_STOP` closes the session and aborts the transaction anyway; the override is invisible to other sessions; and non-leakage is explicitly re-asserted afterwards from a separate connection.
- Reader-bundle pin 44 → 45 correct and complete; a tree-wide grep found no other hardcoded pin. `_in_recovery()` **must** be in the reader bundle because `ash.status()` calls it.
- Security: **no findings**. SECURITY INVOKER, pinned `search_path`, no dynamic SQL, and `pg_is_in_recovery()` is already PUBLIC so the reader grant leaks nothing.
- The standby step is genuinely gated to an existing matrix cell and asserts `pg_is_in_recovery()` before running, so it cannot silently test the primary.

**Note on the CI rollup artifact:** this repo's PRs always carry one `SKIPPED` row (the path-gated duplicate "capture + reel"). A gate requiring `success == total` misreads that as a failure. Judge by the `test (…)` matrix and `ci-required`.
</details>

NikolayS:
## samorev gate: **1 real finding — one half of the recovery assertion is still vacuous**

Run against head `37563e24532152f68da70bddaaa6900461d47fa2`. Second, independent gate on top of the five-role review already in this thread, which I read first.

**CI is green.** 16 success, **0 failure, 0 pending**. The single non-`success` row is `capture + reel (real PostgreSQL)` with conclusion `skipped` — the path-gated duplicate invocation. samorev requires `success == total`, so it reports `ci_status=unknown` and raises CRITICAL. **That CRITICAL is the known rollup artifact, not a finding about this PR**, exactly as the existing review's closing note says.

```text
```

## The real finding, and I think it is rated too low

samorev filed this as LOW. Given that this PR's entire thesis is *"the test passed with the guard deleted"*, an assertion in the same family that **still** passes with the guard deleted deserves more weight than LOW. It is the same defect Major 1 set out to fix, surviving in the half of the claim that was not proven.

**`assert ash.rollup_hour() = 0` remains vacuous.** Confirmed by reading the fixture:

`devel/tests/replica_guards_seed_pending.sql` writes into `ash.sample` only, and sets the watermarks:

```sql
insert into ash.sample (sample_ts, datid, active_count, data, slot) ...
update ash.config
set last_rollup_1m_ts = v_minute - 60,
    last_rollup_1h_ts = null
```

`devel/tests/replica_guards_assert.sql` then guards with two preconditions — but **both of them are about `rollup_minute()`**:

```sql
assert (select count(*) > 0 from ash.sample where sample_ts < ...),
  'fixture: no completed-minute samples — the rollup guard asserts would be vacuous';
assert (select last_rollup_1m_ts is null or last_rollup_1m_ts < (select max(sample_ts) from ash.sample) ...),
  'fixture: rollup watermark is already caught up — the rollup guard asserts would be vacuous';
```

One checks that `ash.sample` has completed-minute rows; the other checks the **1m** watermark is behind. Neither establishes pending work for `rollup_hour()`.

`rollup_hour()` folds `ash.rollup_1m` → `ash.rollup_1h`. On the standby `ash.rollup_1m` is **empty** — the seed never writes it, and `rollup_minute()` is itself guarded so it never runs to populate it. So an *unguarded* `rollup_hour()` finds nothing to fold and returns `0` anyway. Delete the guard and the assertion still passes.

That reading is consistent with the PR's own evidence: the RED/GREEN proof quotes

```text
RED:   ERROR:  rollup_minute() on a standby must return 0, got 60
```

— the `rollup_minute` arm. There is no corresponding `rollup_hour` RED, because there could not be one.

The file's comment says the preconditions exist so the asserts cannot "silently go vacuous again". They do that job for `rollup_minute()` and only appear to do it for `rollup_hour()`, which is the more dangerous state to be in — the guard rail reads as covering both.

**Fix:** seed a completed-hour row into `ash.rollup_1m` as well (`degraded_no_cron.sql` does this by calling `rollup_minute()` then `rollup_hour()` on the primary), and add a third precondition asserting `ash.rollup_1m` has rows older than the current hour with `last_rollup_1h_ts` behind them. Then prove it discriminating the same way Major 1 was proven: restore the unguarded `rollup_hour` body and confirm RED.

## Everything else

No other substantive findings. I re-checked the items the existing review listed as accepted-but-not-actioned (`_in_recovery()` volatility, no standby guard on `grant_reader`/`revoke_reader`, reader spot-checks) and agree with each disposition as recorded.

## Verdict

CI green, gate CRITICAL is the artifact. One real test-integrity finding that I recommend fixing before merge, because it is precisely the class of defect this PR exists to eliminate.

<details>

## samorev Code Review Report

- **PR:** NikolayS/pg_ash#239 - feat: guard write paths and scheduling against running on a replica (#222)
- **Author:** @NikolayS
- **AI-Assisted:** Yes

| Pipeline | Coverage |
|----------|----------|
| unknown | Not reported |

---

### BLOCKING ISSUES (2)

**LOW** [tests] The seed file sets `last_rollup_1h_ts = null` and `last_rollup_1m_ts = v_minute - 60`, but the assertion file's vacuity check for `rollup_hour()` only verifies the 1m watermark; an unguarded `rollup_hour()` with no `rollup_1m` rows present (the seed writes only `ash.sample`) legitimately returns 0, so "rollup_hour returns 0 in recovery" is still vacuous — exactly the failure mode the seed comments warn about.
> seed inserts into `ash.sample` only; assertion precondition checks only `last_rollup_1m_ts`.
> **Fix:** Seed a completed-hour row in `ash.rollup_1m` as well (as `degraded_no_cron.sql` does via `rollup_minute()` then `rollup_hour()`), and add a precondition asserting `rollup_1m` has rows older than the current hour with `last_rollup_1h_ts` behind them.

---

### Summary

| CI/Pipeline | 1 | 0 | 0 |
| Security | 0 | 0 | 0 |
| Bugs | 0 | 0 | 0 |
| Tests | 0 | 1 | 0 |
| Guidelines | 0 | 0 | 0 |
| Docs | 0 | 0 | 0 |
| Metadata | 0 | 0 | 0 |

Note:

<details>

```text
provider=github
kind=pr
project=NikolayS/pg_ash
number=239
target=github:NikolayS/pg_ash#239
state=OPEN
draft=false
diff_lines=10129
diff_added=9824
diff_removed=49
diff_bytes=391975
comments_count=1
commits_count=12
ci_status=unknown
blocking=true
samorev_review=1
review_round=1/10
escalated=false
posted_by=local
no_comment=true
live_posting=not-run
defender_dropped=5
```

**Contested / dropped**

- [bugs] The real-standby CI step clears `shared_preload_libraries` to keep pg_cron from starting on the standby, but the `ash.status()` path (and `start()`/`stop()` guards) are only asserted in the seam step where pg_cron *is* loaded. On a genuine standby with pg_cron loaded (the realistic production case: replicas share the primary's `postgresql.conf`), `start()` raising `25006` "before touching pg_cron" is not actually exercised against a live `cron` schema — the PR description claims the earlier bug was "errors partway through, after touching pg_cron", and that regression is only covered via the seam. — The assertion file already pins the pg_ash guard's own message (not just 25006) on the real standby, and pg_cron's presence is irrelevant to start()'s guard which fires before any cron access; the "touched pg_cron" regression can't reoccur when the guard runs first, so this is a non-actionable test-design nitpick.
- [tests] The seam override of `ash._in_recovery()` is done with `create or replace function` inside a transaction, and `replica_guards_primary.sql` checks only that `_in_recovery()` returns false afterwards. If the installer defines `_in_recovery()` with specific attributes (e.g. `security definer`, `set search_path`, `stable`), a leaked `create or replace` with different attributes would still return `false` but change the function's properties; conversely the seam's replacement does not assert the original attributes were restored. Additionally, any guard that is `immutable`/inlined or cached in a plan would not observe the override — the test would then silently test nothing for that guard. — The seam override runs inside an explicit transaction that is rolled back, so CREATE OR REPLACE cannot leak attributes or body; the installer defines `_in_recovery()` as `stable` (not immutable) with `set search_path`, so plan caching/inlining cannot hide the override from plpgsql callers, and `replica_guards_primary.sql`'s `not _in_recovery()` plus a successful primary `take_sample()` already verify the rollback — comparing `pg_get_functiondef` output adds brittleness without catching a real failure mode.
- [bugs] The installer's early standby check uses `raise exception` inside a `do` block after `begin;` — correct — but the hint/message says "reaches replicas through streaming replication" while logical replicas and `pg_is_in_recovery()` = false cases (e.g. logical subscriber) are not addressed; more importantly, on a standby the failed `do` block leaves psql in an aborted transaction, and because `ON_ERROR_STOP` is on, psql exits non-zero without the trailing `rollback`. For `\i` use inside a caller's existing transaction (documented in the header), the caller's outer work is now in an aborted state with no explicit rollback. — The standby-refusal-inside-`begin;` behaviour is identical to every other installer failure mode (the header already documents that the script owns its transaction and ON_ERROR_STOP aborts it), a `\i` caller's outer work is equally aborted by any later error, psql exits non-zero with the guard's 25006 which the CI test pins, and the logical-replica tangent is out of scope for #222's physical-standby guard.
- [tests] `features_privileges.sql` bumps the reader-function count 44 → 45, but no new reader function is introduced by #222 (only `_in_recovery()`, which should be internal, not in the `grant_reader()` bundle). If the extra grant is `ash._in_recovery()`, the underscore-prefixed helper is being exposed to readers; if it comes from #221's stacked commits, the change belongs to that PR and makes the stacked diff misleading. — The 45th grant is `ash._in_recovery()`, and it must be in the reader bundle: `ash.status()` (a reader function, SECURITY INVOKER) calls it, so excluding it as the fix proposes would break `status()` for pg_monitor/reader roles; underscore helpers readers depend on (`_active_slots`, `_pgss_schema`, `_grain_counts`, …) are already granted the same way, the function merely wraps the PUBLIC-callable `pg_is_in_recovery()`, and the hunk correctly belongs to #222 since that PR introduces the function.
- [docs] README says scheduled routines return "`0` or explicit recovery-skip text" but RELEASE_NOTES says "return a neutral value"; neither documents that `ash.status()` `sampling_enabled` on a standby reflects the primary's config, nor that `take_sample()` returning `0` on a standby is indistinguishable from "no active sessions" for monitoring wrappers that alert on zero samples — the only differentiator is the NOTICE, which most drivers discard. — The README already explicitly directs readers to `ash.status()` → `in_recovery` and states `sampling_enabled` is the primary's replicated config, the README/RELEASE_NOTES wording difference is not a contradiction ("0 or explicit skip text" is a concrete instance of "neutral value"), and the `0`-is-ambiguous point is a LOW-severity doc nitpick about an intentional design choice (scheduled no-ops must not alarm monitoring wrappers) rather than an actionable defect.

</details>

---

</details>

NikolayS:
## samorev re-run on `8b9fa1c` — **previous finding verified fixed, and fixed more thoroughly than I asked**

> Supersedes my report of 2026-08-21, which was against head `37563e2`.

**CI green:** 16 success, 0 failure, 0 pending. The one `skipped` row (`capture + reel`, path-gated duplicate) is what drives samorev's `ci_status=unknown` CRITICAL. Artifact, not a finding.

## Verification of the earlier finding

### `assert ash.rollup_hour() = 0` was still vacuous — **fixed**, by `dc176dd`

I flagged that the seed wrote only into `ash.sample` while both preconditions guarded `rollup_minute()`, leaving `rollup_hour()` with nothing to fold and therefore returning `0` with or without the guard.

The fix goes past my suggestion, and the extra step is the part I had not worked out. Seeding `ash.rollup_1m` alone would **still** have left the assertion vacuous, because `rollup_hour()` only seals an hour once `last_rollup_1m_ts` is past the end of that hour. The seed now sets up all three conditions:

```text
- rollup_1m grains 3 hours back   -> pending hour work
- last_rollup_1m_ts 1 hour back   -> behind the samples, past the hour
- last_rollup_1h_ts null          -> the old hour has never been sealed
```

and two new preconditions were added alongside the existing pair, so the fixture cannot silently regress:

```sql
-- rollup_hour() reads ash.rollup_1m, not ash.sample. Without pending MINUTE ...
assert (select ... from ash.rollup_1m ...), '... assert would be vacuous';
assert (select last_rollup_1m_ts >= (select min(ts) + 3600 from ash.rollup_1m) ...);
```

That second precondition — watermark past the end of the seeded hour — is the one that actually makes the work sealable, and it is the difference between a fixture that looks right and one that is.

**Proven discriminating the same way Major 1 was**, which is what I wanted to see and what was missing before:

```text
RED:   ERROR: rollup_hour() on a standby must return 0, got 2
GREEN: NOTICE: Issue #222 recovery-guard assertions PASSED
```

There is now a genuine RED for the `rollup_hour` arm. My objection is fully answered.

## New this round

**MEDIUM [tests] — `degraded_no_cron.sql` exact-count sampling race.** Does not hold as stated. samorev's premise is that this file lacks the `include_bg_workers = false` that `procedures_call_surface.sql` sets, but the column is `bool not null default false`, so both sample under identical effective settings; the explicit set elsewhere is belt-and-braces. The file also asserts its fixture precondition immediately before sampling, and `ash.sample` stores one row per database, not per backend. Optional hardening only: set it explicitly so the test does not depend on an unstated default.

**LOW [docs] — the guards' "actionable NOTICE" is invisible in the scenario it exists for.** This one is fair and I would take it. A pg_cron job left behind on a demoted node runs with no interactive client, so a `NOTICE` explaining why the routine no-opped goes nowhere useful. It is the same family as the NOTICE-visibility point on #235 and is worth one coherent decision across both PRs rather than two separate ones — the honest options are to document where the message actually surfaces, or to raise the level for the unattended path.

## Verdict

The finding from my previous report is fixed, verified, and the fix is stronger than the one I proposed. Nothing new blocks. One fair docs point about NOTICE visibility, best decided jointly with the equivalent on #235.

<details>

## samorev Code Review Report

- **PR:** NikolayS/pg_ash#239 - feat: guard write paths and scheduling against running on a replica (#222)
- **Author:** @NikolayS
- **AI-Assisted:** Yes

| Pipeline | Coverage |
|----------|----------|
| unknown | Not reported |

---

### BLOCKING ISSUES (3)

**MEDIUM** [tests] `degraded_no_cron.sql` asserts run_take_sample writes exactly 1 sample row, but unlike `procedures_call_surface.sql` (which sets `include_bg_workers = false`) the visible config update only sets `sampling_enabled = true`. If `include_bg_workers` is true at that point (left over from an earlier test in the chain or by default), any background worker active at the sampling instant (autovacuum worker, parallel worker) makes `v_after = 1` fail intermittently — a flaky exact-count assertion.
> `truncate table ash.sample; update ash.config set sampling_enabled = true where singleton; call ash.run_take_sample(); ... assert v_after = 1, format('external path: run_take_sample expected exactly 1 sample row, got %s', v_after);`
> **Fix:** Pin the sampler config the same way `procedures_call_surface.sql` does: `update ash.config set sampling_enabled = true, include_bg_workers = false where singleton;` before the exact-count assertions.

**LOW** [docs] The "actionable NOTICE" emitted by the no-op guards is invisible in exactly the scenario they are designed for. A pg_cron job left behind on a demoted node runs with no interactive client, and with the default `log_min_messages = warning` a `raise notice` is written nowhere — not to the server log and not to `cron.job_run_details` output in a form an operator will see. In practice the demoted-node no-op is fully silent, while README and RELEASE_NOTES both claim an "actionable notice" will be emitted.
> `raise notice 'ash.take_sample: server is in recovery (standby); pg_ash writes only on a primary'; return 0;` (guard in `ash.take_sample`), vs. README: "emit an actionable NOTICE and return their neutral value".
> **Fix:** Either document that the notice is only visible interactively / when `log_min_messages` ≤ notice (and that the demoted-node case is silent by design), or raise the first occurrence per session at WARNING level (which is logged by default) and NOTICE thereafter to keep the no-spam property.

---

### Summary

| CI/Pipeline | 1 | 0 | 0 |
| Security | 0 | 0 | 0 |
| Bugs | 0 | 0 | 0 |
| Tests | 0 | 1 | 0 |
| Guidelines | 0 | 0 | 0 |
| Docs | 0 | 1 | 0 |
| Metadata | 0 | 0 | 0 |

Note:

<details>

```text
provider=github
kind=pr
project=NikolayS/pg_ash
number=239
target=github:NikolayS/pg_ash#239
state=OPEN
draft=false
diff_lines=3028
diff_added=2393
diff_removed=85
diff_bytes=118355
comments_count=2
commits_count=16
ci_status=unknown
blocking=true
samorev_review=1
review_round=2/10
escalated=false
posted_by=local
no_comment=true
live_posting=not-run
defender_dropped=2
```

**Contested / dropped**

- [bugs] `ash._in_recovery()` is declared `stable`, but the function it wraps, `pg_is_in_recovery()`, is volatile in `pg_proc`. The mislabeled volatility permits the planner to evaluate the wrapper once per statement and reuse the result, so a promotion (or the seam override becoming visible) mid-statement is not observed; it also invites SQL-function inlining under the wrong contract. Practically this is an edge case, but it is a semantic mislabel on the single predicate every guard depends on. — Every guard invokes the wrapper as its own statement at plpgsql function entry, so `stable` (once-per-statement) evaluation is behaviorally identical to `volatile` there; a mid-statement promotion race is not handled anywhere in the design regardless, and the seam override works either way — this is a non-actionable semantic nitpick with no observable failure mode.
- [tests] `replica_guards_cleanup.sql` sets `rotated_at = pg_catalog.clock_timestamp()` while restoring "pristine post-install state". Any later step in the same CI job that exercises `ash.rotate()` will now hit the "skipped: rotated too recently" branch for the whole rotation period, so a rotation test placed after the #222 steps silently tests the skip path instead of rotation — the same cross-step state coupling this file exists to prevent. — A fresh install stamps `rotated_at` at install time anyway, so `clock_timestamp()` matches pristine post-install state; the concern is hypothetical — no later step in the job tests rotation without seeding, and the existing rotation test (procedures_call_surface.sql) already backdates `rotated_at` explicitly before calling `run_rotate()`, which is the correct pattern for any future rotation test regardless of this file.

</details>

---

</details>

PR #240, opened 2026-08-21
feat: optional UNLOGGED sample partitions to cut WAL amplification (#224)

### LINKED ISSUE #224: Feature: optional UNLOGGED sample partitions to cut WAL amplification

NikolayS:
Closes #224.

## Why

pg_ash writes ~50–245 MiB/day of raw samples at 1 Hz for 100–500 active backends, and the WAL behind that is larger still (full-page writes plus index maintenance). Every byte ships to **every** replica, **every** Geo secondary, and the WAL archive — for data whose entire purpose is a short-lived diagnostic ring buffer.

## What

```sql
select ash.set_sample_persistence('unlogged');   -- or 'logged'
```

**The default stays logged, deliberately.** If the server dies or restarts uncleanly, unlogged data is truncated — so the ASH history is gone exactly when it is wanted for root-cause analysis. That objection is the reason this was not added before, and it decides the default rather than the feature: the option ships with the durable default intact, for operators who knowingly treat pg_ash as best-effort diagnostics. The README states the trade-off without softening it.

**Only the raw ring changes.** `ash.rollup_1m` / `ash.rollup_1h` always stay logged — they are the long-term value and are tiny, so a crashed primary or a promoted replica still has history at rollup granularity even though the raw ring is gone.

## Mechanics (all verified on PG 18.4, not assumed)

- **`TRUNCATE` preserves `relpersistence`**, so `rotate()` cannot silently undo the setting. This is what makes a per-partition setting viable at all.
- **A partitioned parent cannot be `UNLOGGED`** (`not supported for partitioned tables`), and persistence is not inherited — so only the numeric `sample_N` children are ever touched.
- New partitions are created with the right persistence **up front**, avoiding a rewrite. But `create table if not exists` silently skips an existing table *without changing it*, so creation alone is not enough. `ash._apply_sample_persistence()` reconciles every child afterwards, and **both** creation sites — the installer and `rebuild_partitions()` — call it, so a slot-count change cannot drift back to logged.
- Reconciliation alters only partitions whose `relpersistence` differs, so a no-op call is genuinely a no-op rather than a full ring rewrite.
- Invalid mode raises `22023` **before** touching config; a call on a standby raises `25006` like the other administrative entrypoints (#222). `status()` reports `sample_unlogged`. Both new functions are admin-only.

## Documented trade-offs (all in README, none softened)

- A crash or immediate shutdown **truncates every unlogged partition** — the raw ring is empty exactly when a post-mortem wants it. A clean restart preserves the data. This is why the default is logged.
- A promoted replica starts with an empty sample ring; logged rollups survive.
- Unlogged tables are not readable on standbys at all.
- **Backups do not shrink**: `pg_dump` without `--no-unlogged-table-data` dumps unlogged contents.
- Reducing sampling frequency is the alternative lever for the same problem and costs no durability — named alongside.

The feature-comparison table's "History survives restart: Yes" is now qualified as "Yes (logged default)", since it would otherwise be untrue for opt-in users.

## Manual verification (PG 18.4)

```
default                     : 3 partitions, all 'p'
set_sample_persistence('  UNLOGGED  ') -> "3 partitions converted"
                              sample_0/1/2 = 'u'; sample, rollup_1m, rollup_1h = 'p'
4 rotations across the ring : still 3 x 'u'
rebuild_partitions(6,'yes') : 6 x 'u'
repeat call                 : "0 partitions converted"   (idempotent)
set_sample_persistence('bogus') -> 22023, sample_unlogged still = t
status(): sample_unlogged   | true
set_sample_persistence('logged') -> "6 partitions converted", 6 x 'p'
```

`devel/tests/sample_persistence.sql` passed in full against real Postgres, including the **true installer re-apply** step that proves an operator's unlogged choice survives re-running the installer.

## Tests

Exact partition counts by `relpersistence` through: fresh install, conversion, rotation across the **entire** ring, rebuild to a larger *and* a smaller slot count, round trip back to logged, idempotence reporting zero conversions, the invalid-mode negative case leaving config unchanged, and a genuine installer re-apply. The installer path is passed in as `:install_path` from `ash_sql_chain.py fresh-install-path` rather than hardcoded, so the test keeps working after a release stamp moves the installer.

`devel/tests/config_ordinal_assert.sql` gains the new column in its canonical ordering pin, which is the mechanism that catches accidental `ash.config` layout drift.

## Red/green evidence

Run against real PostgreSQL 18.4, with the pre-#224 installer (the #222 branch tip) as the baseline.

**RED — `devel/tests/sample_persistence.sql`:**

```
psql:devel/tests/sample_persistence.sql:71: ERROR:  column "sample_unlogged" does not exist
CONTEXT:  PL/pgSQL function inline_code_block line 40 at ASSERT
```

**RED — the value-level gap: no persistence control exists at all.**

```
 logged | unlogged
--------+----------
      3 |        0

has_config_column -> f
```

**GREEN — same file, this branch:**

```
NOTICE:  Issue #224 sample persistence assertions PASSED
```

That green run includes the true installer re-apply step, proving an operator's unlogged choice survives re-running the installer.

Honest note on ordering: the commits do not show test-then-implementation. I wrote the tests alongside the implementation and validated behaviourally, then captured the RED above afterwards against the pre-change installer. The red output is real and reproducible, but the commit order does not demonstrate test-first, and I am not going to claim otherwise.

## Diff contents

#245 has merged, so the baseline copy is gone. GitHub shows this PR against `main`, so its diff still includes #235 and #239 (which merge first). **This PR's own change is:**

```
7 files changed, 1485 insertions(+), 48 deletions(-)
```

```bash
git diff feat/replica-guards-222...feat/unlogged-samples-224
```

`base` is deliberately left at `main` rather than retargeted to `feat/replica-guards-222`. `test.yml` and `demo.yml` trigger on `pull_request.branches: [main, 'release/**']`, so a PR based on a `chore/` or `feat/` branch would get **no CI run at all** — the same silent outage as #237, where a PR shows green checks while the suite never ran. Once the parents merge, this diff sheds them automatically with no retargeting needed.

## Stacking

Last of four. Merge order: **#238 (CI) → #235 (#221) → #239 (#222) → this**. All four target `main` so CI actually runs — the workflows only trigger on PRs targeting `main`, so a non-`main` base silently skips the entire suite. Review the top commit for this change; the earlier commits are the parent PRs.

Not merged, per CLAUDE.md — REV review and owner approval still outstanding.


NikolayS:
## Review verdict: **PASS after fixes** — 2 majors found, including a real user-facing defect

A five-role REV review ran against this branch and returned **FAIL**. One of the two majors was a genuine defect in the feature, not a test problem — I reproduced it independently on a real `pg_basebackup` standby before fixing it.

### Major 1 — opting into an unlogged ring silently broke **every reader on every replica**

PostgreSQL refuses to read an `UNLOGGED` relation during recovery, and pg_ash's readers probe the raw ring before they can decide to answer from a rollup. Reproduced on a real standby:

```
select count(*) from ash.status()   -> ERROR: cannot access temporary or unlogged relations during recovery
select count(*) from ash.periods()  -> ERROR: ... (55000)
select count(*) from ash.aas()      -> ERROR: ...
select count(*) from ash.report()   -> ERROR: ...
```

So a WAL optimisation applied on the *primary* killed standby monitoring with an error naming neither pg_ash nor the setting — including `ash.status()`, the one place an operator is told to look, and the row that reports node role. It also made the standby paragraph the parent PR added to the README **false**: logged rollup history was not still readable.

Fixed with `ash._raw_ring_readable()` — false only when in recovery **and** the ring is unlogged — consulted at the two choke points (`ash._raw_oldest_sample()`, which `_pick_source` and `_raw_retention_start` both route through, and `ash.status()`'s three direct scans). Readers now fall through to the always-logged rollups, and `status()` reports raw counters as **unknown rather than zero**, so "cannot read" is never mistaken for "no samples".

**GREEN on the same real standby with an unlogged ring:**
```
ash.status()  -> 36 rows     ash.periods() -> 6
ash.aas()     -> 1           ash.report()  -> 1     ash.top() -> 0
ash.status() still reports in_recovery = true
```
Primary behaviour is identical in both modes.

**One trap worth recording for future guards:** writing this as `select case when readable then (select … from ash.sample) end` in a `language sql` function does **not** work. The uncorrelated sub-select is planned as an InitPlan and evaluated at executor startup regardless of the `CASE`, so it still opens the relation and still raises 55000. I measured that failing before switching to PL/pgSQL, where an `IF` genuinely does not execute the query.

Now pinned in CI: the real-standby step flips the primary to unlogged, waits for the persistence change to replay, asserts all four readers still answer and that `status()` still reports node role, then flips back.

### Major 2 — the `ash.config` OID-stability assertion had gone vacuous

Switching the re-apply pass to `reapply-chain` (currently just the development installer) meant the replayed payload only does `create table if not exists` / `add column if not exists`, so it **cannot** change the OID — the assert could no longer fail. Previously it replayed the released 1.5-to-2.0 migration, whose normalization block *does* rebuild `ash.config`, and the assert was proving its early-return path.

My earlier commit message claimed "the idempotence property being asserted is unchanged". **That was wrong** — one of the two properties had been dropped. Restored by splitting the payload: apply the released migration alone, snapshot the OID, re-apply it against the schema it shipped for (where it genuinely *is* re-apply-safe) and assert no rebuild — all before the overlay stages the new column, after which the migration correctly refuses to run at all. Verified end to end: `released migration re-apply left ash.config in place`.

### Also fixed
- `test -s` on the discovered re-apply chain would have **hard-failed a release stamp**, since `reapply-chain` is legitimately empty once `devel/sql/` is removed. Now skips with a message, matching the pinned v1.5 step.
- `set_sample_persistence(null)` was untested. NULL is a distinct branch — `lower(btrim(null))` is NULL, so `not in (…)` alone yields NULL rather than false, and the `is null` disjunct is what actually rejects it. Deleting it would have gone unnoticed.
- Every reconcile assertion was all-N or zero, so a helper that only ever converted the whole ring — or nothing — would pass. Now drifts exactly one partition and requires exactly one conversion, plus a decoy `ash.sample_backup` pinning the "only numeric children" claim.
- Dropped the `if not exists` that had crept into `rebuild_partitions()`'s creation loop. Step 6 drops every partition first, so a surviving name can only be a leftover non-partition table; skipping it silently would leave the slot unattached and surface much later as `no partition of relation "sample" found for row`.

### Findings accepted but not actioned
- **Lock/rewrite cost of conversion is under-documented.** Fair — `ALTER TABLE … SET LOGGED/UNLOGGED` takes ACCESS EXCLUSIVE and rewrites each partition. The README says conversion rewrites; the blocking implication deserves more, but that is a docs improvement rather than a correctness fix and is better done deliberately than squeezed in here.
- Stale "release-wrapper re-apply" wording in some pre-existing assert messages — cosmetic, and touching unrelated assertion text would enlarge this diff for no behavioural gain.

<details>

- Column-order convergence verified by installing fresh **and** running the full 1.0→head chain on PG 14: `ash.config` attnum order byte-identical, `sample_unlogged` last in both.
- The `add column if not exists` block precedes the partition-creation block, so the upgrade path reads a column that exists.
- Privileges: both new functions confirmed admin-only against PUBLIC and against a live `grant_reader()`-processed role.
- Blast radius: the partitioned parent, `rollup_1m` and `rollup_1h` stay `p` through set/rotate/rebuild/re-apply; a decoy table is never touched.
- PostgreSQL semantics confirmed on **PG 14**, the oldest supported version: `create unlogged table … partition of` a logged parent, and per-partition `set logged/unlogged`, both work.
- Persistence retention verified across 3→5 and 5→3 slot changes.
- The CI change was confirmed **forced, not gratuitous** — the reviewer reproduced the exact `cannot normalize ash.config` failure independently.
- `\i :install_path` genuinely re-runs the installer; the "re-apply preserves the mode" step is not a stub. No wall-clock dependence found.
- Security: **no findings**. Neither function is SECURITY DEFINER, `search_path` pinned on both, the only dynamic SQL passes an identifier through `%I` with a two-valued internal literal.

**Note on the CI rollup artifact:** this repo's PRs always carry one `SKIPPED` row (the path-gated duplicate "capture + reel"). A gate requiring `success == total` misreads that as a failure. Judge by the `test (…)` matrix and `ci-required`.
</details>

NikolayS:
## samorev gate: **1 false positive (disproved), 1 real concern already on the record**

Run against head `5f6252e317a1765aabb456717386bf4c64d05c4d`. Second, independent gate on top of the five-role review already in this thread, which I read first.

**CI is green.** 16 success, **0 failure, 0 pending**. The single non-`success` row is `capture + reel (real PostgreSQL)` with conclusion `skipped` — the path-gated duplicate invocation. samorev requires `success == total`, so it reports `ci_status=unknown` and raises CRITICAL. **That CRITICAL is the known rollup artifact, not a finding about this PR.**

```text
```

## Adjudication

### LOW "reader-grant count 44 → 46 contradicts *both new functions are admin-only*" — **false positive**

samorev suspected either muddled diff attribution or two admin functions leaking into the reader bundle. Neither. This PR adds **three** functions, not two:

```text
ash._apply_sample_persistence
ash._raw_ring_readable
ash.set_sample_persistence
```

`set_sample_persistence` and `_apply_sample_persistence` are both listed in `_admin_funcs()` and correctly excluded from the reader grant. `_raw_ring_readable` is **not** admin-only, and must not be: readers call it — `ash.status()` and `ash._raw_oldest_sample()` consult it at `devel/sql/ash-install.sql:3138, 3147, 4343`. A reader that cannot execute it cannot answer at all. That is the identical rationale that put `_in_recovery()` in the bundle in the stacked #222.

So the delta decomposes cleanly: **44 → 45** from #222's `_in_recovery()`, **45 → 46** from this PR's `_raw_ring_readable()`. I verified each number directly against `devel/tests/features_privileges.sql` on `origin/main` (44), the #222 head (45) and this head (46). The PR's "both new functions are admin-only" claim is true as written — it refers to the two *administrative* functions — though "both" reads as "all the new ones" and is worth rewording to name the third.

**samorev's secondary suggestion is worth taking even though its premise was wrong:** assert the explicit function list rather than a bare count. A count assertion cannot distinguish "the right function was added to the bundle" from "an admin function leaked in and a reader function fell out" — it is precisely the pass-while-broken shape the project's testing rules warn about, and it is what made this finding ambiguous enough to file in the first place.

### HIGH "conversion takes ACCESS EXCLUSIVE and can collide with the 1 Hz sampler" — **real, and already on the record**

The mechanism is right: `ALTER TABLE … SET LOGGED/UNLOGGED` takes `ACCESS EXCLUSIVE` and rewrites each partition, and the `ash_sampler` job runs at 1 Hz under a 500 ms `statement_timeout`. During conversion of a populated ring the sampler will block and then time out, producing a burst of failed samples and `insert_errors` — with nothing pausing sampling and nothing bounding how long `_apply_sample_persistence()` waits.

The existing review already logged this as accepted-but-not-actioned ("lock/rewrite cost of conversion is under-documented … better done deliberately"). I agree it is not a correctness bug and does not need to block. But samorev adds one concrete item that review did not: a **`lock_timeout` inside `_apply_sample_persistence()`**, so a long-running reader cannot stall the ALTER — and therefore the sampler — indefinitely. That is a bounded, mechanical hardening rather than a docs change, and it is cheaper to add now than after someone hits it on a busy ring.

Recommended, in descending order of value:
1. `lock_timeout` around the ALTER loop, so the failure mode is a clean error rather than an unbounded stall.
2. README guidance to convert via `ash.stop()` → convert → `ash.start()`, or at minimum a statement that conversion briefly blocks sampling.

## Note on the standby defect Major 1 fixed

Not a finding — worth amplifying. The `InitPlan` trap recorded in the existing review (an uncorrelated sub-select inside a `CASE` in a `language sql` function is evaluated at executor startup regardless of the `CASE`, so it still opens the unlogged relation and still raises `55000`) is a genuinely non-obvious PostgreSQL behaviour, and the fact that it was *measured* rather than reasoned about is the right standard. Anyone adding a future "don't touch this relation" guard should read that paragraph before writing it in SQL.

## Verdict

CI green, gate CRITICAL is the artifact. The LOW is a false positive I disproved, though its suggested improvement (assert the list, not the count) is worth doing. The HIGH is real but non-blocking and already acknowledged; I would take the `lock_timeout` before merge and leave the docs expansion to a follow-up.

<details>

## samorev Code Review Report

- **PR:** NikolayS/pg_ash#240 - feat: optional UNLOGGED sample partitions to cut WAL amplification (#224)
- **Author:** @NikolayS
- **AI-Assisted:** Yes

| Pipeline | Coverage |
|----------|----------|
| unknown | Not reported |

---

### BLOCKING ISSUES (3)

**HIGH** [bugs] `ash.set_sample_persistence('unlogged')` on a populated ring issues `ALTER TABLE ... SET UNLOGGED` per partition, which takes `ACCESS EXCLUSIVE` and rewrites each partition. If the 1 Hz `ash_sampler` cron job fires concurrently, it blocks (or the ALTER blocks behind it), and the 500ms `statement_timeout` on the sampler job means a burst of failed samples / `insert_errors` during conversion. The returned message warns about rewrite but nothing pauses sampling or documents the lock interaction.
> `'sample persistence: unlogged; %s partitions converted. Converting populated partitions rewrites them; ...'` (devel/tests/sample_persistence.sql, expected message) — no `lock_timeout` or sampling pause around the ALTER loop is visible.
> **Fix:** Set a `lock_timeout` inside `_apply_sample_persistence()` (as `rotate()` presumably does for TRUNCATE) so a long-running reader cannot stall the sampler indefinitely, and document in README that conversion should be run while sampling is stopped (`ash.stop()` → convert → `ash.start()`) or at least that it briefly blocks the sampler.

**LOW** [tests] `features_privileges.sql` bumps the reader-grant function count from 44 to 46, but this PR adds only admin-only functions (`set_sample_persistence`, `_apply_sample_persistence`) which are supposedly excluded from `grant_reader()`. Either the +2 comes from the stacked #221/#222 commits (and the diff attribution is muddled), or the two new functions are being granted to readers, contradicting the PR's "Both new functions are admin-only" and the `_admin_funcs()` assertion.
> `and v_direct_functions = 46` (was 44) while the PR claims both new functions are admin-only.
> **Fix:** Confirm which two functions account for the delta; if they are the #224 functions, remove them from the reader grant set. Consider asserting the explicit function list rather than a count so this class of drift is self-explaining.

---

### Summary

| CI/Pipeline | 1 | 0 | 0 |
| Security | 0 | 0 | 0 |
| Bugs | 0 | 1 | 0 |
| Tests | 0 | 1 | 0 |
| Guidelines | 0 | 0 | 0 |
| Docs | 0 | 0 | 0 |
| Metadata | 0 | 0 | 0 |

Note:

<details>

```text
provider=github
kind=pr
project=NikolayS/pg_ash
number=240
target=github:NikolayS/pg_ash#240
state=OPEN
draft=false
diff_lines=11273
diff_added=10890
diff_removed=57
diff_bytes=435459
comments_count=1
commits_count=18
ci_status=unknown
blocking=true
samorev_review=1
review_round=1/10
escalated=false
posted_by=local
no_comment=true
live_posting=not-run
defender_dropped=5
```

**Contested / dropped**

- [bugs] `rotate()` on an unlogged partition after an unclean restart: PG truncates unlogged relations at crash recovery, but `ash.config` watermarks (`last_rollup_1m_ts`, rotation state) are logged and survive. `rollup_minute()` will then see a watermark pointing into a now-empty ring — likely benign (returns 0) but "endangered sample" rollup in `rotate()` may assume the slot has rows to fold before truncating. No test exercises the post-crash state (empty `sample_N`, stale config), and the README documents the data loss but not whether rollup/rotation remain consistent afterwards. — Post-crash state is benign by construction — rotate() counts endangered rows (0 after truncation → skips rollup), rollup_minute() upserts from an empty ring and returns 0 gaps, and a pg_ctl -m immediate step is out of scope for a service-container CI matrix; the finding is speculative test-coverage, not a demonstrated bug.
- [bugs] Mode string is trimmed and case-folded (`'  UnLoGgEd  '` accepted) but the description says invalid mode raises `22023`; a NULL argument is not covered. If the function is not `STRICT`, `lower(trim(NULL))` yields NULL and `NULL in ('logged','unlogged')` is NULL (not false), so a plain `if mode not in (...)` guard would not raise and the function could proceed with `sample_unlogged = NULL`, which `status()` and `_apply_sample_persistence()` then treat inconsistently. — The function already guards `if v_mode is null or v_mode not in (...)` raising 22023, and `$extra_negatives$` in sample_persistence.sql explicitly asserts `set_sample_persistence(null)` -> 22023 with config unchanged, so the finding is already handled.
- [bugs] Migration path is missing. `sample_unlogged` is added to `ash.config` and pinned in `config_ordinal_assert.sql`, but only `devel/sql/ash-install.sql` is in the changed files — no `sql/migrations/ash-1.5-to-2.0.sql` (or 2.0 beta delta) change. Existing 2.0 beta 1 installs upgraded via the migration chain will lack the column, and `status()` / `rotate()` / `rebuild_partitions()` that reference `sample_unlogged` will fail with `42703`. — The project's upgrade chain deliberately applies the development installer as an overlay after the immutable released migrations (ash_sql_chain.py full-upgrade-chain / reapply-chain, explained at length in the CI diff), and that installer already carries `add column if not exists sample_unlogged`, with CI's config-ordinal test exercising exactly that chain; editing released sql/migrations/ mid-cycle contradicts the documented release-stamp process.
- [docs] README claims "With an unlogged raw ring the samples themselves are physically unreadable during recovery, so readers on a standby answer from the always-logged rollups instead, and `ash.status()` reports the raw-sample counters as unknown rather than zero." This is a significant behavioural claim (reader fallback and a distinct "unknown" status state) that is not covered by any assertion in `sample_persistence.sql`, `replica_guards_assert.sql`, or the real-standby CI step, and the PR description's verification list does not mention it. — The real-standby CI step already asserts exactly this claim — it converts the ring to unlogged, confirms `relpersistence = 'u'` on the replica, runs `ash.status()`, `periods()`, `aas()`, and `report()` successfully against the standby, and pins `in_recovery = true`; the "unknown" counters are implemented via `_raw_ring_readable()` nulling `samples_total`/`samples_in_current_slot`, so the README is accurate and the only residual gap (asserting the literal NULL value) is a minor test-strengthening nit, not a docs defect.
- [guidelines] The diff bundles three features (#221, #222, #224) plus a new-file view of an 8.4k-line installer into one review unit, and the description asks reviewers to "review the top commit" only. Per CLAUDE.md, any post-review commit on the parent PRs invalidates this review; a reviewer checking only the top commit cannot verify the earlier commits in this stack have been re-reviewed after their own fix-commits. — The finding targets PR stacking process, not a code defect; the description already commits to the correct merge order and to re-review per CLAUDE.md, and the bundled diff is an unavoidable artifact of stacked PRs targeting main for CI, not something fixable in this change.

</details>

---

</details>

NikolayS:
## Fresh review round 2: **FAIL → fixed** — the first fix was partial, and its NULL was lying

A second independent five-role REV review ran against the post-fix head, on a **real `pg_basebackup` standby**. It returned FAIL with four majors. It was right on all four; my earlier fix was incomplete and in one respect made things worse. Fixed in `fix: complete the unlogged-standby guard and stop it lying (#224)`.

### Major 1 — the guarded call-site set was incomplete

`_pick_source` was genuinely fixed, but four readers bypass source selection entirely: `ash.samples()` and `ash.decode_sample()`/`decode_sample_at()` read the ring unconditionally, and `aas`/`timeline`/`top` **force** `v_source := 'raw'` whenever `query_id` is given, because an exact per-query drill cannot be answered from a rollup. All still raised a bare `55000`.

### Major 2 — the NULL I introduced was read as "your data is gone forever"

This is the one that matters most. `NULL` from `_raw_oldest_sample()` reached `_raise_tie_retention`, which reported:

> the requested window is entirely outside raw retention (no raw samples exist). Exact query attribution is **unrecoverable** for that window — narrowing it will not help.

Every part of that is false on a standby: the samples exist and the drill works on the primary. **Trading a confusing-but-honest error for a confident wrong one is a regression, not a fix.**

Both fixed with `ash._require_raw_ring(what)`, which raises a pg_ash-authored `55000` naming the cause *and* the remedy, called at all six sites. Because it fires before source selection, the tie-retention path is now unreachable in recovery, so the false claim is gone.

### Major 3 — `ash.status()` was still half-lying

`samples_in_current_slot` and `samples_total` became NULL, but `last_sample_ts` and `raw_retention_start` still said **"no samples"**, and the `time_since_last_sample` row silently disappeared. Combined with rollup-backed `periods()` showing zeros, a standby looked exactly like a dead sampler — the precise misreading the fix existed to prevent. All five raw rows now report `unknown (unlogged ring, in recovery)`, with a stable row shape so scrapers see a changed value rather than a missing metric.

### Minor 4 — the predicate trusted intent over physical truth

`_raw_ring_readable()` read `ash.config.sample_unlogged` rather than `pg_class.relpersistence`. The two can drift — precisely what `_apply_sample_persistence()` exists to repair — and during that drift the guard would return "readable" against unlogged partitions and put the `55000` straight back. Now derived from `pg_inherits`/`pg_class`: same cost, cannot drift.

### Verified on a real standby with an unlogged ring
```
rollup-backed:  status()=37 rows  periods()=6  aas()=1  report()=1
raw-only:       samples(), decode_sample_at(), aas(query_id:=),
                timeline(query_id:=), top(query_id:=)
                -> "... needs the raw sample ring, which is UNLOGGED ..."
status() raw rows -> all five "unknown (unlogged ring, in recovery)"
```
Primary unaffected in both persistence modes.

### Test findings, also fixed
- The CI standby step pinned only the readers that already worked. It now also pins the ones that must **raise**, asserts the word `unrecoverable` never appears, pins all five status markers, and **fails loudly if the persistence change never replays** — the wait loop previously fell through, which would have run every assertion against a still-logged ring and passed trivially.
- The drift assertion re-invoked the mutating helper inside `format()`, repairing the ring before building the message, so it always misreported `0 converted` and did DDL on the failure path.
- The decoy was a standalone table, already excluded by the `inhparent` predicate — it passed even with the documented `'^sample_[0-9]+$'` filter deleted. It is now a real partition (`sample_spare`).

README now names which readers degrade to rollups and which cannot answer, and the trade-off bullet cross-references it.

Reader-bundle pin 46 → 47 for `_require_raw_ring`. **Note:** the 45 → 46 move was caught by CI rather than by me, which is that pin doing its job — I checked the count locally before pushing this time.

<details>
<summary>Reviewer's clean list</summary>

- Normalization idempotence: genuinely non-vacuous and not skippable — deleting the migration's early-return makes the second apply rebuild `ash.config` and the OID assert fire. Still correct post-release-stamp.
- Full `sample_persistence.sql` run: rc=0, both notices, ring handed to cleanup in the expected state, `set_sample_persistence(null)` → `22023` with config unchanged.
- Reader-bundle pin correct and complete; the bundle is enumerated dynamically, so the new function is in it by construction. No other pinned count in the tree.
- The plpgsql conversion of `_raw_oldest_sample()` has no planner or hot-path implication — it contained `min()` and was already non-inlinable.
- `report()` specifically probed for the InitPlan trap this fix warns about: it returns cleanly on the standby.
- Primary confirmed byte-identical in both persistence modes.
- Security: **no findings**.

**CI rollup artifact:** one `SKIPPED` row (the path-gated duplicate "capture + reel") is normal here; judge by the `test (…)` matrix and `ci-required`.
</details>

NikolayS:
## samorev re-run on `aec32f5` — **both previous items addressed**; one new finding is real

> Supersedes my report of 2026-08-21, which was against head `5f6252e`.

**CI green:** 16 success, 0 failure, 0 pending. One `skipped` row (`capture + reel`, path-gated duplicate) drives samorev's `ci_status=unknown` CRITICAL. Artifact, not a finding.

## Verification of the earlier items

### HIGH conversion takes ACCESS EXCLUSIVE and can stall the 1 Hz sampler — **addressed** by `aec32f5`

I recommended a `lock_timeout` inside `_apply_sample_persistence()` so a long-running reader cannot stall the ALTER and therefore the sampler indefinitely. That is exactly what landed:

```sql
set lock_timeout = '5s'
```

and the catalog comment now states the reasoning, including the property that makes the bound safe rather than merely shorter:

> Runs with `lock_timeout = 5s`: the ALTER takes ACCESS EXCLUSIVE and rewrites the partition, so it fails fast rather than stalling the sampler; **reconciliation is idempotent so the next call finishes a timed-out conversion.**

The idempotence argument is the part that matters — a bounded wait is only an improvement if a timeout leaves the ring in a resumable state, and it does. The failure mode is now a clean error plus a partially-converted ring that the next call completes, instead of an unbounded stall with a burst of `insert_errors`. Good.

### LOW reader-grant count 44 → 46 — **was a false positive; the suggestion behind it was taken anyway**

To restate for the record: the count was always correct. This PR adds three functions, not two — `set_sample_persistence` and `_apply_sample_persistence` are both in `_admin_funcs()` and correctly excluded, while `_raw_ring_readable()` must be reader-granted because `ash.status()` and `_raw_oldest_sample()` call it, exactly as `_in_recovery()` must be.

But the underlying complaint — that a bare count cannot distinguish "the right function was added" from "an admin function leaked in and a reader function fell out" — was worth acting on, and `aec32f5` acts on it. `features_privileges.sql` now pins the explicit signature array:

```sql
assert v_reader_signatures = array[
  ...
  '_in_recovery()',
  ...
  '_raw_ring_readable()',
  ...
```

alongside the count (now 47, reflecting the stacked #221 work). That converts an assertion that could pass while broken into one that names what it expects. This is the change I would have most wanted from the whole round, and it came out of a finding whose premise was wrong — worth noting for how these get triaged.

## New this round — one real, two not blocking

### MEDIUM [bugs] `replica_guards_cleanup.sql` does not restore the state it claims to — **real**

Confirmed by reading both files. The seed registers a synthetic wait event:

```sql
select ash._register_wait('active', 'Guard', 'PendingWork') into v_wait_id;   -- seed_pending.sql:40
```

which inserts into `ash.wait_event_map`. The cleanup, headed *"Restore the pristine post-install state"*, truncates `ash.sample`, `ash.rollup_1m`, `ash.rollup_1h` and the `query_map_N` partitions — **but not `ash.wait_event_map`**. So `Guard:PendingWork` survives into every subsequent step of the job.

Low blast radius, but the contract in the comment is not met, and a leftover dictionary row is exactly the sort of thing that later surfaces as a confusing schema-snapshot or count mismatch in an unrelated step. Either delete the seeded wait id in the cleanup, or narrow the comment to say what it actually restores. The first is better — a cleanup that is trusted to be complete should be complete.

### LOW [bugs] exact-count sampling race in `degraded_no_cron.sql` — does not hold

Same item samorev raised on #235 and #239. Its premise is that this file lacks the `include_bg_workers = false` that `procedures_call_surface.sql` sets, but the column is `bool not null default false`, so both sample under identical effective settings. The file also asserts its fixture precondition immediately before sampling, and `ash.sample` stores one row per database rather than per backend. Optional hardening only.

### LOW [docs] `ash.set_sample_persistence(mode)` carries no caution in the README functions table

Fair. `ash.rebuild_partitions(n, 'yes')` is annotated "destructive for raw samples" in that table; switching to `unlogged` means the ring is emptied by crash recovery and is not available on a standby, which is a comparable thing for an operator to know at a glance. Docs-only.

## Verdict

Both items from my previous report are addressed, and the reader-bundle change is a genuine improvement to test quality rather than a minimal compliance edit. One new real finding — the cleanup not undoing `_register_wait` — which is small and worth fixing before merge. The other two new items are one non-issue and one docs nicety.

<details>

## samorev Code Review Report

- **PR:** NikolayS/pg_ash#240 - feat: optional UNLOGGED sample partitions to cut WAL amplification (#224)
- **Author:** @NikolayS
- **AI-Assisted:** Yes

| Pipeline | Coverage |
|----------|----------|
| unknown | Not reported |

---

### BLOCKING ISSUES (4)

**MEDIUM** [bugs] `replica_guards_cleanup.sql` claims to "Restore the pristine post-install state" but does not undo everything the seed created: `replica_guards_seed_pending.sql` calls `ash._register_wait('active', 'Guard', 'PendingWork')`, which registers a synthetic entry in the wait dictionary, and the cleanup never removes it (it only truncates sample/rollup/query_map tables and rewinds watermarks). Any later step in the same job that pins exact dictionary contents, dictionary size, or `register_wait_cap_hits`-adjacent state will see the leftover 'Guard'/'PendingWork' entry — the exact cross-step pollution the file's own comment warns about.
> replica_guards_seed_pending.sql: `select ash._register_wait('active', 'Guard', 'PendingWork') into v_wait_id;` vs. replica_guards_cleanup.sql, which truncates only `ash.sample`, `ash.rollup_1m`, `ash.rollup_1h`, and `ash.query_map_N`.
> **Fix:** In the cleanup script, delete the seeded row from the wait dictionary (or truncate/re-seed the dictionary to its post-install contents) and reset any counters `_register_wait` bumps, so the "pristine post-install state" claim actually holds.

**LOW** [bugs] In `degraded_no_cron.sql`, the "exactly 1 sample row" assertion is race-prone: unlike `procedures_call_surface.sql`, this test does not visibly set `include_bg_workers = false` before sampling, so an autovacuum worker (or any transient backend in another database) active at the sampling instant produces a second per-database row and a spurious failure. The fixture assertion only pins the count of backends tagged `pg_ash_cronless_workload`, not the total set of sampleable backends.
> `truncate table ash.sample; update ash.config set sampling_enabled = true where singleton; call ash.run_take_sample(); ... assert v_after = 1` — compare `procedures_call_surface.sql`, which sets `include_bg_workers = false` for the same style of exact-count assertion.
> **Fix:** Add `include_bg_workers = false` to the config update in this test (and restore prior config state afterwards), or relax the assertion to count only rows for the current database.

**LOW** [docs] The README functions table annotates `ash.rebuild_partitions(n, 'yes')` as "destructive for raw samples" but the new `ash.set_sample_persistence(mode)` row carries no caution, even though switching to `unlogged` makes the entire raw ring crash-volatile and the conversion itself rewrites every populated partition under ACCESS EXCLUSIVE locks (which can stall a 1 Hz sampler). The trade-off is well documented in the dedicated section, but a reader scanning only the table — the place they will copy the call from — gets no signal that this call has operational consequences.
> `| \`ash.set_sample_persistence(mode)\` | Set raw partitions to \`logged\` or \`unlogged\` |` alongside `| \`ash.rebuild_partitions(n, 'yes')\` | Recreate raw partitions; destructive for raw samples |`
> **Fix:** Extend the table cell, e.g. "Set raw partitions to `logged` or `unlogged`; `unlogged` trades crash survival of raw samples for less WAL — see the persistence section."

---

### Summary

| CI/Pipeline | 1 | 0 | 0 |
| Security | 0 | 0 | 0 |
| Bugs | 0 | 2 | 0 |
| Tests | 0 | 0 | 0 |
| Guidelines | 0 | 0 | 0 |
| Docs | 0 | 1 | 0 |
| Metadata | 0 | 0 | 0 |

Note:

<details>

```text
provider=github
kind=pr
project=NikolayS/pg_ash
number=240
target=github:NikolayS/pg_ash#240
state=OPEN
draft=false
diff_lines=4833
diff_added=3867
diff_removed=122
diff_bytes=197831
comments_count=3
commits_count=24
ci_status=unknown
blocking=true
samorev_review=1
review_round=2/10
escalated=false
posted_by=local
no_comment=true
live_posting=not-run
defender_dropped=2
```

**Contested / dropped**

- [tests] The README documents an entirely new behavioral surface for standbys with an unlogged ring (readers `ash.samples()`/`ash.decode_sample()`/exact `query_id` drills raise SQLSTATE `55000`; rollup-backed readers degrade; `ash.status()` reports raw rows as `unknown (unlogged ring, in recovery)`), but no test exercises a standby while the ring is unlogged. The real-standby workflow step belongs to #222 and runs with the default logged ring, and `sample_persistence.sql` runs entirely on a primary — the PR's own scenario enumeration (fresh install, conversion, rotation, rebuild, round trip, idempotence, invalid mode, installer re-apply) contains no standby case. The seam trick used for #222 (override `ash._in_recovery()`) cannot cover the physical-unreadability path either, since on a primary the unlogged table remains readable. — The finding's premise is factually wrong — the "#222: recovery guards on a real streaming standby" step in this diff already flips the ring to unlogged, waits for replay, and asserts rollup-backed readers degrade, raw-only readers fail with the pg_ash message, and every raw status() metric shows the exact `unknown (unlogged ring, in recovery)` marker.
- [bugs] In `procedures_call_surface.sql`, the `raise exception '... unexpectedly resolved'` sits inside the same `begin ... exception when others` block, so if the mis-resolution ever *does* resolve, the sentinel exception is caught by the block's own handler and surfaces as a confusing `assert` failure quoting SQLSTATE `P0001` and the sentinel text, rather than the intended clear error. The test still fails (no false pass), but the diagnostic is misleading. — The mis-diagnosis only occurs in a practically unreachable path (a SELECT of a procedure resolving), the test still fails rather than falsely passing, and the assert's failure message quotes the sentinel text "unexpectedly resolved" verbatim via sqlerrm — so the diagnostic actually names the real cause, making this a non-actionable nitpick.

</details>

---

</details>

PR #242, opened 2026-08-21
ci: run tests on release-branch PRs, and make CI requirable (#241)

### LINKED ISSUE #241: ci: PRs into release branches run no tests, and no status check is required to merge into main
Two independent holes let a change reach `main` with no CI evidence behind it. Both are process-level, both are currently live.

## 1. `test.yml` never runs for release-branch PRs

`.github/workflows/test.yml` filters both triggers to `branches: [main]`. A PR targeting a release line — `release/2.0`, for example — therefore never starts the PG 14–19 matrix at all.

Observed on the open PR into `release/2.0`: `statusCheckRollup` is empty. Zero checks, ever.

`demo.yml` is unaffected — it uses unfiltered `push:`/`pull_request:`.

CodeQL is configured through code-scanning **default setup**, which covers the default branch only and has no branch knob, so release-branch PRs get no `CodeQL` check either. That is a separate, lower-severity gap (advanced setup would fix it) and is documented rather than fixed here.

## 2. The `main-protected` ruleset requires no status checks

Ruleset `13093534` has `deletion`, `non_fast_forward` and `pull_request` rules — and no `required_status_checks` rule. Nothing at the ruleset level requires a check to have *run*, let alone passed. A PR whose workflows never started is mergeable into `main` with no evidence:

- a fork PR sitting in "workflow approval required"
- a branch the trigger filters excluded (hole 1)
- a workflow file over the 512,000-byte limit, which GitHub refuses to run silently (#237)

CLAUDE.md makes green CI a merge precondition and a release-gate input. Right now that is enforced only by whoever is looking at the PR page.

## Fix

- Extend `test.yml`'s `push` and `pull_request` triggers to `release/**`.
- Add a `ci-required` aggregator job so a single stable check name covers the whole matrix (the matrix contexts embed PG versions and are renamed at every bump, so requiring them by name rots).
- Document the exact ruleset payload and `gh api -X PUT` command for the required-status-checks rule. Applying it is an owner-level change and is not done in the PR.

NikolayS:
Closes #241.

Two holes that both end the same way: a change reaching `main` with no CI evidence behind it.

## Hole 1 — release-branch PRs run no tests

`.github/workflows/test.yml` filtered both triggers to `branches: [main]`, so a PR targeting `release/2.0` never started the PG 14–19 matrix. The open PR into `release/2.0` has an empty `statusCheckRollup` — zero checks, ever.

```diff
   push:
-    branches: [main]
+    branches: [main, 'release/**']
     tags: ['v*']
   pull_request:
-    branches: [main]
+    branches: [main, 'release/**']
```

`demo.yml` needed no change — it already uses unfiltered `push:`/`pull_request:`, which is why the demo checks *do* appear on release-branch PRs. Those two were the only workflow files; CodeQL is code-scanning **default setup**, covering the default branch only, with no branch knob. Release-branch PRs therefore still get no `CodeQL` check. Fixing that means moving to advanced setup (a committed `codeql.yml`); I judged that out of scope and documented it instead, because it changes how code scanning is configured repo-wide.

## Hole 2 — nothing is required to merge into `main`

Ruleset `main-protected` (`13093534`) has `deletion`, `non_fast_forward` and `pull_request` rules and **no `required_status_checks` rule**. Nothing requires a check to have *run*. A PR whose workflows never started is mergeable with no evidence: a fork PR pending workflow approval, a branch the trigger filters excluded, or a workflow file over the 512,000-byte limit that GitHub silently refuses to run (#237).

Changing the ruleset is an owner-level action, so this PR does not touch it. It ships the payload instead: `docs/ci/main-protected-ruleset.json`, which reproduces the three existing rules verbatim and adds the fourth.

```bash
gh api -X PUT repos/NikolayS/pg_ash/rulesets/13093534 \
  --input docs/ci/main-protected-ruleset.json

# verify
gh api repos/NikolayS/pg_ash/rulesets/13093534 \
  --jq '.rules[] | select(.type=="required_status_checks")'
```

Contexts required, and why those four:

| Context | App id | Why |
| --- | --- | --- |
| `ci-required` | 15368 | Stable rollup of `docs-lint` + the whole PG matrix |
| `renderer (fixtures, no database)` | 15368 | Fast demo gate, every PR, stable name |
| `capture + reel (real PostgreSQL)` | 15368 | Real-database demo gate, every PR |
| `CodeQL` | 57789 | Default setup covers the default branch, which is all this ruleset targets |

Not required: the individual `test (N, on)` contexts (see below), `Analyze (actions)`/`Analyze (python)` (rolled up by `CodeQL`), and `size` (only exists once #238 lands — add it then).

`strict_required_status_checks_policy: false` is recommended in the payload. `true` forces every branch up to date with `main` before merge; with a dozen PRs open at once each merge invalidates the rest and forces a full matrix re-run, serialising the queue for little benefit — the matrix is a schema/behaviour test, not a semantic-conflict detector.

## The `ci-required` aggregator — yes, worth it

I did implement the aggregator (`if: always()`, `needs: [docs-lint, test]`, fails on any result other than `success`/`skipped`), for two reasons.

The matrix contexts embed PostgreSQL versions — `test (14, on)`, `test (19beta2, on)`. Requiring them by name means the ruleset silently stops covering a version the moment the matrix changes, or blocks every PR on a context that no longer exists. One stable name covers whatever the matrix contains next year.

It also turns the #237 failure mode into something visible. A required check whose workflow never ran stays **pending**, and the PR is blocked — as opposed to today, where absent checks are indistinguishable from nothing to report.

Two behaviours worth stating explicitly, since they decide whether a required check is real:

- A job skipped by a **job-level `if:`** still reports a check run with conclusion `skipped`, and GitHub treats that as satisfying a required check. So `capture + reel (real PostgreSQL)` being required does not by itself prove it ran — its `if:` does include `github.event_name == 'pull_request'`, so on PRs it does run. A workflow skipped **entirely** (branch/path filter, or a file over the size limit) reports nothing at all, and the required check stays pending. That asymmetry is the whole reason this works.
- Fork PRs need workflow approval; until a maintainer approves, nothing reports, the required checks stay pending, and the PR cannot be merged. That is precisely the case that is mergeable today.

## Size

`test.yml` grows 509,333 → 510,165 bytes, leaving 1,835 bytes under the 512,000 limit. Thin, deliberately: I kept the edit to the trigger block plus an appended job so it does not conflict with #238, which cuts the file to ~394,000. **#238 should land before any further test additions.** I did not consider putting the aggregator in a separate workflow file — `needs:` cannot cross workflows.

## Verification

- `actionlint` on both workflows: no new findings (the four pre-existing `SC2016`/`SC2295` info notes are untouched lines).
- Both files parse; `test.yml` jobs are `docs-lint`, `test`, `ci-required`; triggers confirmed as `[main, release/**]`.
- Ruleset JSON parses and reproduces the three current rules field-for-field.
- Trigger change materialising runs is confirmed by this PR's own checks.

Not merged, per CLAUDE.md — REV review and owner approval still outstanding. Applying the ruleset payload is an owner action.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -2,10 +2,10 @@ name: Test pg_ash
 
 on:
   push:
-    branches: [main]
+    branches: [main, 'release/**']
     tags: ['v*']
   pull_request:
-    branches: [main]
+    branches: [main, 'release/**']
   workflow_dispatch:
     inputs:
       release_tag:
@@ -11420,3 +11420,25 @@ jobs:
           CRON: ${{ matrix.cron }}
         run: |
           echo "All tests passed for PostgreSQL $PG_MAJOR (pg_cron=$CRON)"
+
+  # One stable check name for the branch ruleset's required status checks.
+  # The matrix contexts embed PG versions ("test (19beta2, on)") and change
+  # every release, so requiring them by name would rot. Fails if any needed
+  # job failed or was cancelled; a job that never ran leaves this pending.
+  ci-required:
+    if: always()
+    needs: [docs-lint, test]
+    runs-on: ubuntu-latest
+    steps:
+      - name: Assert every required job succeeded
+        env:
+          RESULTS: ${{ join(needs.*.result, ' ') }}
+        run: |
+          set -euo pipefail
+          echo "needed job results: ${RESULTS}"
+          for r in ${RESULTS}; do
+            case "${r}" in
+              success|skipped) ;;
+              *) echo "a required job reported: ${r}" >&2; exit 1 ;;
+            esac
+          done
--- a/docs/CI.md
+++ b/docs/CI.md
@@ -1,0 +1,108 @@
+# CI and branch protection
+
+This file records which workflows run where, and the exact branch-ruleset
+configuration that makes "CI is green" a real merge precondition rather than
+an honour-system one.
+
+## Workflows and their triggers
+
+| Workflow | File | Triggers |
+| --- | --- | --- |
+| `Test pg_ash` | `.github/workflows/test.yml` | push to `main` and `release/**`, tags `v*`, PRs into `main` and `release/**`, `workflow_dispatch` |
+| `demo` | `.github/workflows/demo.yml` | every push, every PR, nightly cron, `workflow_dispatch` |
+| CodeQL | code scanning **default setup** (no file in this repo) | push to the default branch and PRs into it, weekly |
+
+`test.yml` originally filtered both `push` and `pull_request` to `branches:
+[main]`. Release-branch PRs — the `release/2.0` line, for example — therefore
+merged with the full PG 14–19 matrix never having run once. `demo.yml` has
+always used unfiltered `push:`/`pull_request:`, so it needed no change.
+
+CodeQL is configured through code-scanning *default setup*, which covers the
+default branch only and has no branch knob. A PR into a `release/**` branch
+will not get a `CodeQL` check run. That matters for the required-checks list
+below: if the ruleset is ever extended beyond `~DEFAULT_BRANCH`, requiring
+`CodeQL` would leave release-branch PRs blocked on a check that can never
+arrive. Switching to advanced setup (a committed `codeql.yml`) is the way to
+get CodeQL onto release branches, and is out of scope here.
+
+## The `ci-required` aggregator
+
+`test.yml` ends with a `ci-required` job: `if: always()`, `needs: [docs-lint,
+test]`, failing if any needed job reports anything other than `success` or
+`skipped`.
+
+It exists because the matrix contexts embed PostgreSQL versions — `test (14,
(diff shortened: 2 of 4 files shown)


NikolayS:
Two verification notes.

**The `release/**` glob is confirmed live, not just parsed.** I pushed this branch to a throwaway `release/ci-trigger-probe` ref: `Test pg_ash` started for it on the `push` event, which it would not have done before this change. Probe branch deleted afterwards.

**Merging this does not retroactively fix the open release-branch PR.** For `pull_request` events GitHub runs the workflow from the merge commit of head into base. `release/2.0` and the PR branch targeting it were both cut before this change, so their merged `test.yml` still carries `branches: [main]` and still produces zero checks. Bringing `main` into `release/2.0` (or cherry-picking this commit onto it) is what makes that PR start running the matrix — worth doing before anything else merges into that line.

Same caveat applies to any future release branch cut from a `main` that predates this commit.

PR #244, opened 2026-08-21
fix(ci): end a run block at a dedented comment, and gate the parser (#243)

### LINKED ISSUE #243: CI tooling: ci_step_script.py cannot parse test.yml, and its test suite never runs in CI
## What is broken

`devel/scripts/ci_step_script.py` can no longer parse `.github/workflows/test.yml`, so its own test suite fails on a pristine `main`:

```
$ python3 devel/scripts/test_ci_step_script.py
ci_step_script.WorkflowError: line 8830: run block content must be indented more than the run key
FAILED (errors=1)
```

That makes the "extract and run one CI step locally" workflow unusable — which is the tool's whole purpose, and the fastest way to reproduce a matrix failure without pushing.

## Cause

The literal-block reader does not treat a **dedented comment line** as the end of a `run: |` block. It sees the comment as block content, finds it indented *less* than the `run` key, and raises.

Nothing was wrong with the workflow. The trigger is simply that `test.yml`'s last step in the `test` job is now followed by a comment at job indentation, introduced with the `ci-required` job. Before that, the final step ran to end-of-file and the reader never had to terminate a block scalar on a comment.

Minimal repro — the only difference between these two is one comment line:

```python
tmpl = """jobs:
  test:
    steps:
      - name: Last step
        run: |
          echo hi
%s
  other-job:
    runs-on: ubuntu-latest
"""
```

```
WITH dedented comment    -> ERROR: line 5: run block content must be indented more than the run key
WITHOUT comment          -> parsed OK: ['Last step']
```

A comment at lower indentation legitimately terminates a block scalar in YAML, so the reader should stop there rather than treat it as content.

## Why this reached main

**`test_ci_step_script.py` is not run in CI.** The `docs-lint` job runs `test_release_stamp.py` and nothing else:

```
$ grep -rn "test_ci_step_script\|test_release_stamp" .github/workflows/
.github/workflows/test.yml:31:  run: python3 devel/scripts/test_release_stamp.py -v
```

So the parser's 7-test suite has no gate behind it and can break silently — the same shape of problem as #237, where a check that never runs is indistinguishable from a check with nothing to report.

## Acceptance criteria

- [ ] `_deindent_literal_block` (or its caller) ends a block scalar at a dedented comment line instead of raising.
- [ ] A regression test with the minimal repro above: a `run: |` step followed by a comment at job indentation parses, and the step body is exactly `echo hi\n` — comment text must not leak into the body.
- [ ] Blank lines inside a block scalar still work, and a genuinely under-indented *content* line still raises (do not fix this by loosening the check unconditionally).
- [ ] `python3 devel/scripts/test_ci_step_script.py` passes against the real `.github/workflows/test.yml`.
- [ ] `test_ci_step_script.py` runs in CI, in a job that does not depend on the workflow it parses being parseable — otherwise a future break is silent again.

## Test plan

Red first: the repro above fails on current `main` with the exact error quoted. Then fix, and assert both the parse and the exact extracted body, plus the two negative cases (under-indented content still raises; blank lines preserved).

NikolayS:
Closes #243.

## The bug

`devel/scripts/ci_step_script.py` could no longer parse `.github/workflows/test.yml`, so its own suite failed on a pristine `main`:

```text
$ python3 devel/scripts/test_ci_step_script.py
ci_step_script.WorkflowError: line 8830: run block content must be indented more than the run key
Ran 7 tests — FAILED (errors=1)
```

Nothing is wrong with the workflow. When scanning for the end of a `run: |` block, the reader used `_indent()`, which returns `None` for **comment** lines as well as blank ones. A comment at job indentation was skipped rather than treated as a terminator, so the block ran on past it into the next job, the comment's own indentation became the computed minimum, and the guard fired.

Any step that is last in its job trips this the moment another job follows it — which is exactly what the `ci-required` job in #242 introduced.

## The fix

New `_block_scalar_indent()`, which returns `None` **only** for blank lines, used for the terminator scan.

The distinction it draws is the one YAML actually makes: a `#` line indented at or beyond the block's content is ordinary content — a shell comment, which is extremely common in this repo's steps — while one dedented back to the parent's level ends the scalar. The old code could not tell those apart because it discarded the indentation of every comment.

Behaviour is unchanged for every non-comment input. I verified that by running the pre-fix and post-fix parsers over the same fixtures and diffing the results, rather than assuming it.

## Red / green

Four tests fail against the unfixed parser, including the real-workflow integration test:

```text
ERROR: test_dedented_comment_terminates_run_block
ERROR: test_dedented_comment_and_no_comment_agree
ERROR: test_blank_lines_inside_a_block_are_preserved
ERROR: test_current_workflow_integration
Ran 13 tests — FAILED (errors=4)
```

All 13 pass after it:

```text
Ran 13 tests in 0.040s
OK
```

End to end against the real workflow, showing the comment does not leak into the body:

```text
$ python3 devel/scripts/ci_step_script.py step "Final summary"
echo "All tests passed for PostgreSQL $PG_MAJOR (pg_cron=$CRON)"
```

## Coverage added

- the issue's minimal repro, asserting the body is **exactly** `echo hi\n`
- with-comment and without-comment parses must produce identical bodies, since the comment is the only difference
- shell comments at **and beyond** content indentation stay in the body — the over-correction guard, since terminating on any comment at all would silently truncate most real steps here
- blank lines inside a block still preserved
- tabs still rejected on the termination path
- a characterization test pinning that an under-indented content line is not newly tolerated

## Gating it so this cannot regress silently

`docs-lint` now runs `test_ci_step_script.py`. It belongs there rather than in `test` deliberately: this suite parses `test.yml`, and a job inside that workflow's own matrix cannot be trusted to report on the parseability of the file it was scheduled from. `docs-lint` needs no database and runs on every event, so a break is loud and immediate.

Cost to `test.yml` is **8 lines** (8853 → 8861), five of which are the comment explaining the placement — worth being mindful of given the 512,000-byte workflow ceiling that #237 dealt with.

## One correction to the acceptance criteria

The issue asks that "a genuinely under-indented *content* line still raises". **Measured against the pre-fix parser, it never did.** A non-comment line at or below the parent indentation has always simply ended the scalar, and the `content_indent <= parent_indent` guard was only ever reachable through the comment path — that is, through the bug itself. The `indentation < content_indent` guard is likewise unreachable, since `content_indent` is the minimum over the same set of lines.

```text
                                          pre-fix              post-fix
under-indented content (9 vs run key 8)   OK ' echo one\necho two\n'   identical
way-under content (7)                     OK 'echo one\n'              identical
block content at step indent              OK ''                        identical
```

So rather than invent a new error to satisfy the wording, I pinned the existing behaviour in a characterization test with the reasoning written down, so it cannot be quietly relaxed later. Making the parser genuinely strict here would be a deliberate behaviour change that could reject workflows it accepts today, and belongs in its own PR if wanted.

Every other acceptance criterion is met.

## Verification

```text
python3 devel/scripts/test_ci_step_script.py   ->  13 tests, OK
python3 devel/scripts/test_release_stamp.py    ->  17 tests, OK
yaml.safe_load(test.yml)                       ->  jobs: docs-lint, test, ci-required
```

Not merging — flagging for review.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -30,6 +30,14 @@ jobs:
       - name: Test release stamp checker
         run: python3 devel/scripts/test_release_stamp.py -v
 
+      # Lives here, not in `test`, on purpose: this suite parses test.yml, and
+      # a job inside that workflow's matrix cannot be trusted to report on the
+      # parseability of the file it was scheduled from. docs-lint needs no
+      # database and runs on every event, so a break is loud and immediate.
+      # Without this the suite had no gate at all and #243 reached main.
+      - name: Test CI step extractor
+        run: python3 devel/scripts/test_ci_step_script.py -v
+
       - name: Verify release tag matches payload stamp
         if: >-
           github.ref_type == 'tag' ||
--- a/devel/scripts/ci_step_script.py
+++ b/devel/scripts/ci_step_script.py
@@ -59,6 +59,26 @@ def _indent(line: str, *, line_number: int) -> int | None:
     return len(leading)
 
 
+def _block_scalar_indent(line: str, *, line_number: int) -> int | None:
+    """Indentation for block-scalar termination; None only for blank lines.
+
+    Unlike :func:`_indent`, a comment line reports its real indentation. Inside
+    a literal block a ``#`` line indented at or beyond the content is ordinary
+    content (a shell comment), but one indented back to the parent's level ends
+    the scalar, exactly as YAML specifies. Treating it as content instead makes
+    the block appear to start at the comment's indentation.
+    """
+    text = _without_eol(line)
+    leading = text[: len(text) - len(text.lstrip(" \t"))]
+    if "\t" in leading:
+        raise WorkflowError(
+            f"line {line_number}: tabs are not valid YAML indentation"
+        )
+    if not text.strip():
+        return None
+    return len(leading)
+
+
 def _is_mapping_key(line: str, indent: int, key: str) -> bool:
     text = _without_eol(line)
     prefix = " " * indent + key + ":"
@@ -192,7 +212,7 @@ def _deindent_literal_block(
 ) -> str:
     block_end = end
     for idx in range(start, end):
-        indentation = _indent(lines[idx], line_number=idx + 1)
+        indentation = _block_scalar_indent(lines[idx], line_number=idx + 1)
         if indentation is not None and indentation <= parent_indent:
             block_end = idx
             break
--- a/devel/scripts/test_ci_step_script.py
+++ b/devel/scripts/test_ci_step_script.py
@@ -154,6 +154,123 @@ class WorkflowParserTests(unittest.TestCase):
             b"second\0echo done\n\0",
         )
 
+    # ---- issue #243: a dedented comment must end a run block ----
+
+    # The workflow is fine; the parser was not. Before the fix the reader
+    # skipped comment lines when looking for the end of a `run: |` block, so a
+    # comment at job indentation was swallowed as block content, dragged the
+    # computed content indentation down to its own, and raised. Any step that
+    # is the last one in its job trips this the moment another job follows.
+    DEDENTED_COMMENT = """\
+jobs:
+  test:
+    steps:
+      - name: Last step
+        run: |
+          echo hi
+%s
+  other-job:
+    runs-on: ubuntu-latest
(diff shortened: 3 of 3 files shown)


NikolayS:

This parser fix is carried into #252, with additional fail-closed parsing and CI gate coverage. The integrated candidate will rerun the complete gate before an RC.

Issue #248, opened 2026-09-04
Release: v2.0 RC implementation sprints and verification tracker


NikolayS:
Prepare and fully verify **v2.0-rc1** from the composed v2 candidate. The owner explicitly authorizes RC tags only. **Do not create the final v2.0 tag or publish a final 2.0 release without explicit owner approval.** The owner will test the RC fully before that decision.

Audit base: `2c3286ac2551a89ab6bf6e2c982b6988578d5640`. All 12 open preparation PRs were reviewed using pinned Tanya301/samorev plus independent code/data/docs/migration agents. Green individual PRs are not proof of an integrated release candidate.

## Sprint 1 — release blockers and dependable CI

- [ ] CI: land #244 extractor fix; reject skipped mandatory dependencies; include devel/sql in real-demo triggers; require workflow size across release branches; enforce active required checks.
- [ ] #137: prevent new mixed-history reweighting with exact-value red/green tests; document coarse-cadence minute-bucket aliasing and keep weighted-time design open.
- [ ] #203: compose lifecycle atomicity with #235 CALL/custom-command behavior and #239 recovery guards; reject cross-database named-job collisions.
- [ ] #202: compose migration transaction ownership + #205 token hardening with #240 sample_unlogged normalization; test temporary promoted public-wrapper layout before stamping.

## Sprint 2 — product/API/documentation

- [ ] Compose #235 → #239 → #240 without losing reviewed behavior.
- [ ] Finalize lifecycle/cadence/diagnostic compatibility and privilege contracts; no unreviewed status renames hidden in copy changes.
- [ ] Correct #232/#233/#234/#236 and salvage #201 prose; restore original LLM investigation using v2 calls and real output.
- [ ] Execute examples, regenerate demos, manually inspect README/visuals.
- [ ] Measure overhead including saturated workload (#226), or remove unsupported enable-by-default claims.

## Sprint 3 — integrated verification and RC

- [ ] samorev on every finalized PR head, independent full-context review of truncated diffs, findings adjudicated and fixed.
- [ ] Full Docker PG14–18 + supported PG19 prerelease; fresh/regression, complete upgrade/reapply/schema equivalence, all new/existing features, four dependency modes, privileges, real standby/promotion, logged/unlogged crash/restart, docs/demo reproduction.
- [ ] Test each actual prior release/prerelease origin, including alpha1..alpha5 and beta1, against promoted canonical migration and root wrapper.
[...]

PR #251, opened 2026-09-04
fix: make lifecycle changes atomic without stealing custom cron jobs

### LINKED ISSUE #203: ash.stop() suppresses rollup unschedule failures, leaving jobs after rebuild/uninstall

### LINKED ISSUE #250: fix: reject cross-database managed pg_cron job-name collisions
The v2 feature stack scopes pg_cron lookups to the installation database, but pg_cron's named-job uniqueness is per owner across databases. If the same owner already has an `ash_sampler`/rotation/rollup name targeting another database, `start()` can reach `cron.schedule()` with the colliding name; `stop()` also still unschedules rollup names without database qualification.

Required v2 contract: detect all five same-owner foreign-database names before changing configuration or jobs, raise an actionable error naming the job and database, and leave the entire call unchanged. Do not silently steal or retarget another database's job. Explicit start should reactivate same-database managed jobs while preserving customized commands.

Related integration work: #203/#204 lifecycle rollback and #235/#239/#240 feature stack. Tracked in #248. A replacement implementation and regression will link here.

NikolayS:
Fixes #203. Fixes #250. Contributes to #248. Supersedes the lifecycle implementation in #204 after its unique changes have been accounted for.

When a managed job cannot be unscheduled, stop/start/rebuild/uninstall now raise and roll back configuration, earlier job changes, and DDL together. Lifecycle calls require the ash schema owner; SET ROLE is the explicit remedy for another administrator. Actual removals return real job IDs; absent jobs remain idempotent.

This composes the fix with #235/#239/#240: recovery guards and sample persistence survive; recognized commands migrate to CALL and customized commands are preserved. Explicit start reactivates local jobs. Same-owner managed job names targeting another database are rejected before changes instead of being stolen. Non-superuser owners retain administrator-configured pg_cron connection defaults when restricted settings are unreadable.

Validation: externalized #203 regression fails against the previous #240 payload and passes against this implementation on repository-built PostgreSQL 18 Docker, including locked-row rollback, real returned IDs, missing jobs, custom commands/reactivation, cross-database collision rejection, cross-superuser refusal, visible foreign-owner refusal, and non-superuser schema-owner start/stop. Shell parse and diff checks pass. Test is wired for every cron-on CI major and the local cron-path gate. Current-base extractor still needs #244; the #244 extractor successfully selects the new step. Full combined release gate remains outstanding.

Stack: this PR targets main so workflows execute; it is based on #240 and therefore includes #235/#239/#240 until those land. Own commit: 70bf5b5. Review that incremental delta. Draft pending samorev, hosted checks, full combined verification and owner approval. No release identity changes, merge or tag.


NikolayS:
Independent review of lifecycle head 9a51895 and integrated source eea1ffc: no blocking finding.

Real PostgreSQL18 evidence: full test_lifecycle_atomicity.sh passed, including foreign owner/cross-database collision and atomic rollback controls. An additional LOGIN schema-owner reproduction called start twice, deactivated all managed jobs, preserved custom command text through reactivation, and verified a cron sampler actually fired successfully and inserted raw rows. This is runtime evidence beyond a NOLOGIN/SET ROLE catalog-only check.

Local evidence: /tmp/pg-ash-251-independent.log and /tmp/pg-ash-251-owner.log. Final frozen-SHA release gate is pending; this does not authorize merge or release.

NikolayS:
Samorev first-round review adjudication (pinned Tanya301/samorev 24924b4; actual tool report retained at `/tmp/pg-ash-lifecycle-samorev.md`). This is not an exact-head passing review.

Fixed in9a51895: ordinary schema-owner second start no longer requires privileged cron.alter_job; inactive jobs are reactivated with checked removal/recreation while preserving custom command text; start rejects visible foreign-owner managed jobs before creating duplicates; README/release notes describe current_user owner/SET ROLE and fail-closed contract; failure cleanup removes the regression role/jobs. Meaningful real PG18 red/green fixtures reproduce the original permission/duplication defects. Independent review also used a LOGIN non-superuser owner and verified that reactivated cron jobs actually fired and wrote samples.

The cross-database teardown finding is an intentional policy choice: lifecycle operations refuse ambiguous same-owner managed names across databases before mutation. The release plan explicitly requires an actionable collision instead of adopting or modifying another database's named jobs. This restriction, including stop/rebuild/uninstall, is documented and regression-tested. It is not being represented as a generic pg_cron limitation.

Follow-up b441e3d corrects the old #162 foreign-owner rotation fixture to expect the new rejection, assert no changes, remove its own injected collision, then retry successfully. Exact-head hosted CI is still required; draft status and the full integrated release gate remain open. Latest samorev attempts are blocked by the Claude session usage limit until18:30PDT September4; no failed provider call is treated as a pass.

Issue #137, opened 2026-07-10
Design blocker: persist sampling cadence for trustworthy AAS


NikolayS:
## Summary

The 2.0 AAS readers multiply every historical appearance count by the current singleton `ash.config.sample_interval`. Sampling cadence is not persisted with raw samples or rollups.

Changing cadence therefore rescales history, and cadences coarser than one minute can fabricate one-minute peaks.

This is an architectural trust blocker for stable 2.0, not a one-line reader fix.

Verified on `origin/main` at `5aac2e3` with Postgres 18.3.

## Cadence-change reproduction

Two true one-backend minutes:

```text
minute A collected at 1 second: 60 appearances = true AAS 1
minute B collected at 5 seconds: 12 appearances = true AAS 1
current config.sample_interval = 5 seconds
```

Observed:

```text
timeline = 5.00, 1.00
aas avg = 3.00
peak = 5.00
backend_seconds = 360
```

Expected:

```text
timeline = 1.00, 1.00
aas avg = 1.00
peak = 1.00
backend_seconds = 120
```

## Coarse-cadence reproduction

With `sample_interval = 2 minutes`, one backend observed once over a two-minute window:

```text
aas avg = 1.00
aas peak = 2.00
timeline = [2.00, no-data]
report aas_worst1m.total = 2.00
```

The peak exceeds the one observed backend because the full two-minute weight is assigned to one minute.

## Root cause

`ash._sample_interval_secs()` reads the current config value. It is used by `aas`, `timeline`, `top`, `report`, and `chart`. Neither `ash.sample` nor the rollup rows preserve the cadence or weighted backend time that produced their counts.

## Decision required before stable 2.0

Preferred direction:

- persist cadence or weighted backend-time with raw samples;
- aggregate weighted backend-time into minute/hour rollups;
- define upgrade semantics for existing rows that lack cadence metadata; and
- keep peak/p99 grain honest across cadence transitions.

Minimum containment if the storage migration is deferred:

- prevent `ash.start(new_interval)` from changing cadence while retained history exists, or explicitly discard/segment history;
- reject cadences incompatible with fixed one-minute report semantics; and
- document the constraint in catalog comments and README.

## Regression requirements

- a 1-second → 5-second transition must preserve both minutes at true AAS 1;
- historical values must not change when the current config changes;
- cadence greater than one minute must not produce a one-minute peak above observed concurrency.

NikolayS:
[...]

PR #253, opened 2026-09-04
fix: protect retained history cadence

NikolayS:
Changing `sample_interval` currently rescales every retained sample and rollup. A minute collected at 1 second becomes AAS 5 after `ash.start('5 seconds')`; a 2-minute sampling interval can fabricate a one-minute peak above observed concurrency. This adds bounded containment for #137 while weighted cadence storage remains future work.

- Accept 1–60 whole seconds. Refuse `ash.start(interval)` and direct config changes while any raw, minute, or hour history remains, without deleting history.
- Make omitted `ash.start()` intervals resume the configured cadence; fresh installations still default to 1 second. Explicit NULL remains an error.
- Serialize changes with the sampler and all three history tiers using nonblocking locks, and reject changed cadence at repeatable-read/serializable isolation rather than trust a stale snapshot.
- Preserve unsupported legacy config/data on upgrade and re-apply. Sampling skips and AAS weighting refuses unsupported cadence with recovery guidance. Earlier mixed-cadence history cannot be reconstructed or certified as accurate.

Validation: the new regression failed on the parent because a direct config change with raw history succeeded; it passes after this change. Native PostgreSQL 18 and Docker PostgreSQL 18 with real pg_cron/pg_stat_statements pass exact-value cadence and concurrency tests. Full public feature suites pass both with cron+pgss and without either extension. Released-baseline upgrade plus installer re-apply preserves legacy config/data and exercises the guarded reader/sampler behavior. Release-stamp tests pass (17). CI and samorev are pending on this draft.

Stack: based on #240, which includes #235 and #239. The PR targets main so repository PR workflows run; parent changes remain visible until the stack merges. Own implementation begins at commit 36a6d57. Related lifecycle/docs branches must retain the custom-command, owner/atomicity, and new cadence contracts when integrated. No released SQL or version stamp changed.

Relates to #137 and v2 readiness tracker #248. This is containment, not persisted cadence or sampler heartbeat storage. No merge, tag, or final v2 release is requested by this draft.


NikolayS:
Samorev review ran against `36a6d57` using Tanya301/samorev pinned at `24924b4`. One substantive finding was confirmed and fixed in `ba6deab`: the new cadence update ran before the pg_cron <1.5 error-row return. It now runs after the version check, before job changes. A new regression fails on `36a6d57` with `old pg_cron rejection changed config` and passes on the fix with real pg_cron/PostgreSQL 18.

All seven matrix cells and `ci-required` passed on the first head; the follow-up has a fresh CI run. Draft status is intentional pending the v2 integration/release gate. The full-stack diff exceeds samorev's 40k review input budget, so this report is not proof of complete diff coverage. An untruncated scoped runtime review hit its first 180s timeout and is being retried; no PASS is claimed for that timeout.

The README/release notes now explicitly retain the coarse subminute limitation: exact 59-second sampling can put two observations in one minute. This PR contains historical reweighting and >60-second inputs; it does not claim weighted-time storage or retroactive repair of old data. #137 remains open.

<details><summary>Raw samorev report before the fix</summary>

## samorev Code Review Report

- **PR:** NikolayS/pg_ash#253 - fix: protect retained history cadence
- **Author:** @NikolayS
- **AI-Assisted:** Yes

| Pipeline | Coverage |
|----------|----------|
| PENDING | Not reported |

---

### BLOCKING ISSUES (3)

**HIGH** `CI/Pipeline` - Pipeline status is pending
> Provider CI reported status `pending`.
> **Fix:** Wait for CI to finish and rerun review.

**MEDIUM** [bugs] `ash.start()` now persists the new `sample_interval` before the pg_cron version check, so the "pg_cron too old for sub-minute scheduling" path returns an `error` row via a normal `return` while the cadence change has already been applied. Before this PR the config update ran only after all scheduling succeeded, so every error-row return left config untouched. The new comment claims "Any later scheduling failure still rolls this update back atomically", but that is only true for raised exceptions, not for the error-row return at line 2547. Result: on pg_cron < 1.5 with empty history, `ash.start('5 seconds')` reports an error, schedules nothing, yet `ash.config.sample_interval` is now 5 seconds and a later bare `ash.start()` resumes that value.
> `update ash.config set sample_interval = every where singleton;` (line 2420-2423) precedes `if v_seconds < 60 then job_type := 'error'; ... return next; return; end if;` (lines 2546-2554), and no exception is raised on that path.
> **Fix:** Move the validating `update ash.config set sample_interval = every` to just after the pg_cron version check (still before any `cron.unschedule`/`cron.schedule`), or convert that error-row return into a `raise exception` so the update rolls back. Add an assertion mirroring the existing "ash.start(oversized): config unchanged" check for this path, and drop the now-redundant duplicate `set sample_interval = every` in the two later updates (lines 2462 and 2877) since the trigger already handled the change.

---

### Summary

| CI/Pipeline | 1 | 0 | 0 |
| Security | 0 | 0 | 0 |
| Bugs | 1 | 0 | 0 |
| Tests | 0 | 0 | 0 |
| Guidelines | 0 | 0 | 0 |
| Docs | 0 | 0 | 0 |
| Metadata | 1 | 0 | 0 |

Note:

<details>

```text
provider=github
kind=pr
project=NikolayS/pg_ash
number=253
target=github:NikolayS/pg_ash#253
state=OPEN
draft=true
diff_lines=5772
diff_added=4449
diff_removed=277
diff_bytes=240299
comments_count=0
commits_count=28
ci_status=pending
ci_summary=total=15 success=4 failure=0 pending=10 other=1
blocking=true
samorev_review=1
review_round=1/10
escalated=false
posted_by=local
no_comment=true
live_posting=not-run
defender_dropped=1
```

**Contested / dropped**

- [bugs] `ash.take_sample()` reads `sample_interval` (and the other config fields) before acquiring the `pg_ash_sampler` advisory lock. The trigger's `pg_try_advisory_xact_lock` only excludes a sampler that already holds the lock, so a sampler that has read config, then loses the race to a cadence change that commits with empty history, will still take the lock and insert one appearance that AAS now weights at the new cadence. The PR description says changes "serialize with the sampler"; this window is small but real, and the Python concurrency test only models a holder that already owns the advisory lock, so it cannot catch it. — The stale read is only used for the 1–60s range check (also enforced by the trigger); the pg_stat_activity sample and insert happen after the lock, so the single row lands under the cadence in force at commit and the next pg_cron tick already runs at that new cadence, making the weighting consistent rather than wrong.

</details>

---
*samorev-assisted review (AI analysis by [Tanya301/samorev](https://github.com/Tanya301/samorev))*

</details>

NikolayS:
Scoped samorev review completed against the original own change, split into untruncated runtime/tests (28,508 characters) and integration/docs groups. All actionable notes have follow-up commits:

- `ba6deab`: move cadence mutation after pg_cron version rejection; real red/green config-rollback test.
- `46eb374`: expose unsupported legacy cadence through `sample_interval_supported`, count skipped calls, document SQLSTATEs and explicit reset recipe; legacy upgrade test passes.
- `c89e7d5`: document locks held until commit and roll back the interval-format test fixture.

The first full seven-cell CI run passed on `36a6d57`; follow-up heads have independent CI runs. Round-two samorev is running. #137 remains open for persisted weighted time and existing mixed-cadence history; the README includes the 59-second bucket-placement limitation.

<details><summary>Untruncated scoped runtime review before fixes</summary>

FINDING:
- severity: HIGH
- confidence: 8
- area: Bugs
- issue: `ash.start()` now persists the cadence change to `ash.config` before the pg_cron-version check, which returns an error row instead of raising. On pg_cron < 1.5 with sub-minute intervals, `start('5 seconds')` reports an error, creates no jobs, but leaves `sample_interval = 5 seconds` committed while any existing `ash_sampler` job keeps firing at the old schedule. That is precisely the configured-vs-actual cadence divergence #137 is meant to eliminate, and `start()` with no argument will now silently resume the wrong value via `_configured_sample_interval()`. The comment "Any later scheduling failure still rolls this update back atomically" is only true for exceptions, not for the error-row returns this function uses. Before this PR the update happened only at the end, after all checks.
- evidence: `update ash.config set sample_interval = every where singleton;` (early, ~line 2422) followed later by `if v_seconds < 60 then job_type := 'error'; ... return next; return; end if;` in the pg_cron version block (~line 2547).
- fix: Move the pg_cron version check ahead of the early config update (it only needs `v_seconds` and `pg_extension`), or convert that error path to `raise exception` so the update rolls back. Add a test: stub/skip on modern pg_cron is fine, but at minimum assert that any error-row return from `start()` leaves `ash.config` unchanged (extend the invalid-interval loop in `cadence_history.sql` to cover this path).
- file: devel/sql/ash-install.sql
- line: 2422

FINDING:
- severity: MEDIUM
- confidence: 6
- area: Bugs
- issue: After an upgrade with an unsupported legacy cadence, `take_sample()` silently returns 0 with only a `raise warning` in the server log. It does not bump `skipped_samples` or any counter, and `status()` still reports `sampling_enabled = true` with the legacy interval and no indication that sampling is disabled. An operator looking at `ash.status()` sees a healthy installation that is collecting nothing; the pg_cron job keeps running and logging warnings every tick indefinitely.
- evidence: `if v_sample_seconds is null or ... then raise warning 'ash.take_sample: unsupported legacy sample_interval; sampling skipped.' ...; return 0; end if;` with no counter update, versus the adjacent `not v_sampling_enabled` branch that increments `skipped_samples`.
- fix: Increment `skipped_samples` (or a dedicated counter) on this path, and expose a `sample_interval_supported` (or similar) metric in `ash.status()` so the disabled state is visible without log access. Assert the counter/metric in `cadence_legacy.sql` rather than only `take_sample() = 0`.
- file: devel/sql/ash-install.sql
- line: 1169

FINDING:
- severity: LOW
- confidence: 5
- area: Docs
- issue: `ash.start()` now has two different failure contracts: every interval-shape problem returns a `job_type = 'error'` row, but a refused cadence change with retained history raises SQLSTATE 55000 from the config trigger (and lock contention raises 55P03). The function comment says it "refuses cadence changes" without stating that this refusal is an exception, not an error row, so callers that check the result set will not see it and scripts using `ON_ERROR_STOP` will abort where they previously got a row.
- evidence: `perform ash.start(interval '5 seconds'); raise exception 'start changed cadence with raw history'; exception when sqlstate '55000'` in `cadence_history.sql` documents the exception behavior; the `comment on function ash.start(interval)` does not.
- fix: Either catch 55000/55P03 in `start()` and return an error row consistent with the other validation paths, or state explicitly in the function comment and column comment that retained-history refusal raises SQLSTATE 55000 and sampler contention raises 55P03.
- file: devel/sql/ash-install.sql
- line: 7981

FINDING:
- severity: LOW
- confidence: 5
- area: Docs
- issue: The trigger hint and the upgrade warning tell operators to "explicitly archive and remove all raw, minute-rollup, and hour-rollup history first", but neither the hint nor any function comment says how. `ash.sample` is a partitioned ring with per-slot children, so the only working recipe (`truncate ash.sample, ash.rollup_1m, ash.rollup_1h` as owner, which also needs the `last_rollup_*_ts` watermarks reset as the test does) is undocumented. The new tests themselves reset `last_rollup_1m_ts`/`last_rollup_1h_ts` to null, which suggests a cadence change on an emptied installation without that reset leaves stale watermarks.
- evidence: `hint = 'Keep the current cadence. To start a new cadence, explicitly archive and remove all raw, minute-rollup, and hour-rollup history first; ...'` and the test fixture `update ash.config set sample_interval = interval '1 second', last_rollup_1m_ts = null, last_rollup_1h_ts = null;`.
- fix: Document the exact statements in the hint or in the `sample_interval` column comment (truncate the three tiers, reset rollup watermarks), or clarify whether stale watermarks are harmless after a truncate and add a test that changes cadence after a truncate without resetting them.
- file: devel/sql/ash-install.sql
- line: 4513

</details>
<details><summary>Scoped integration/docs review before fixes</summary>

Review complete. CI on the PR is green across PG 14 through 19beta2, and I traced every changed test site against the new trigger in the installer. Findings below are minor.

FINDING:
- severity: LOW
- confidence: 5
- area: Docs
- issue: README and release notes describe the guard's fail-fast direction only. They omit that a successful cadence change keeps SHARE locks on ash.sample, ash.rollup_1m, and ash.rollup_1h until the enclosing transaction ends, so calling `ash.start(interval)` or updating the config inside a long transaction stalls the sampler and rollup writers instead of them skipping.
- evidence: README: "Changes require a `READ COMMITTED` transaction and fail promptly if a sampler or history writer is active." Trigger (devel/sql/ash-install.sql): `lock table ash.sample, ash.rollup_1m, ash.rollup_1h in share mode nowait;` held to transaction end.
- fix: Add one sentence to the README limitation bullet and the release-notes entry: run the cadence change in its own short autocommit statement, because it blocks history writers until commit.
- file: README.md
- line: 666

FINDING:
- severity: LOW
- confidence: 4
- area: Tests
- issue: The interval-format DO block now truncates all three history tiers permanently, outside any transaction, unlike the other new call sites in this PR which wrap the truncate in `begin ... rollback` or `begin ... commit`. It also leaves `sample_interval` at 60 seconds afterward. CI is green, so nothing downstream depends on it today, but any future step placed after this one that seeds history and assumes the prior cadence will fail with the new "retained history exists" error in a non-obvious place.
- evidence: `truncate ash.sample, ash.rollup_1m, ash.rollup_1h;` at the top of the `do $$` block, followed by `perform ash.start('60 seconds')` / `ash.start('1 minute')` with no restoring update or transaction wrapper.
- fix: Wrap the block in `begin; ... rollback;` like `devel/tests/features_lifecycle.sql` does for its start/stop scenarios, or restore `sample_interval = '1 second'` at the end while history is still empty.
- line: 3495

Everything else in the diff checks out:

- Every direct `sample_interval` write that remains in tests runs after all three tiers are empty in the same transaction, so the trigger accepts it. The 5-second AAS scaling test at test.yml line 6453 is safe because that step truncated raw and both rollups at its start and only reinserts into `rollup_1m`.
- The migration Round 3 transaction cannot flake on the sampler advisory try-lock. The live legacy `ash_sampler` cron job runs with `sampling_enabled = false`, and that path returns before acquiring the lock.
- The rewritten invalid-interval loop uses `into strict`, so a missing error row fails loudly. All eight values are whole seconds above 60, so the expected "maximum 60 seconds" message is the correct branch.
- The raw-retention B3 assertions were updated correctly: `ash.start()` with no argument now preserves 7 seconds, and the cron status string matches the "schedule updated ...; command re-synced" form the installer emits.
- The new CI steps wire their psql variables and PG* environment consistently with what `cadence_legacy.sql` and the concurrency script read. The concurrency script's three lock scenarios and the two isolation-level rejections match the trigger's error codes.
- Dropping the 1-second reset before the live `take_sample` assertion in features_lifecycle is fine. Those assertions check row counts, slot, and decoded wait/query id, none of which depend on cadence.

</details>

NikolayS:
Latest pushed head c45a11e includes the narrow unsupported-cadence health follow-up and reproducible physical-runtime test. No existing status metric names changed; sample_interval_supported is additive and unsupported sampler calls count as skipped samples.

The new devel/scripts/test_physical_runtime.py passed against integrated source eea1ffc on PostgreSQL18.3: actual streaming logged raw, UNLOGGED standby aggregate/status reads, promotion, clean restart, immediate shutdown recovery for both persistence modes. It creates and cleans private native clusters and never uses an existing database.

Samorev round2 was unavailable, not a pass. Diagnosis: samorev24924b4 invokes Claude Code2.1.261; a minimal health probe exits1 with stdout “You've hit your session limit · resets6:30pm (America/Los_Angeles)” and empty stderr. Exact-head review remains pending until the September4 18:30PDT reset. Prior completed, untruncated scoped findings have been addressed.

PR #255, opened 2026-09-04
fix: stage atomic v2 upgrades and test promoted public entrypoints

### LINKED ISSUE #202: 1.5-to-2.0 migration can stamp 2.0 before config normalization fails

NikolayS:
Fixes #202. Contributes to #248. Salvages the migration transaction correction from #200 and token hardening from #205 without their obsolete final-2.0 stamp.

Stages the complete current-line cumulative migration under devel/sql, including sample_unlogged in every normalization definition and copy projection. Installer replay and normalization share one transaction; a transaction-ID token and cached installer ownership decision prevent ambient settings from bypassing atomicity. Released SQL and all version stamps are unchanged.

Discovery allows a staged replacement only for the current prerelease's exact migration edge and matching installer line. It rejects finalized/older overrides and mismatched targets. Full, pinned and reapply development paths exercise this staged transaction. A separate temporary-promotion test copies the candidate into the real canonical/root-wrapper layout and verifies fresh install and upgrades from actual v1.0 through v1.5 plus alpha1 through alpha5 and beta1, including schema/config/OID preservation, wrapper reapply and logged/unlogged choices. The stamp promotes this tested body, not a newly invented migration.

Evidence on repository-built PG18 Docker:
- #202 full atomicity suite PASS: custom trigger, dependent view, same-client marker rollback, external marker on v1.5 and empty database; fresh-connection identity/schema/config checks.
- RED against held #200 installer: external marker allows partial publication 1.5|1.5|11 -> 2.0|2.0|11 after forced late failure. Current implementation rolls back to 1.5|1.5|11.
- Temporary promoted public entrypoints: fresh + all 12 actual historical tag origins PASS; schema/config/reapply/OID and unlogged-choice checks.
- 5 new discovery tests and all 17 existing stamp/chain tests PASS. #244 extractor selects the updated upgrade range. YAML remains below 460,000 bytes.

Tests are external files wired into hosted CI and the existing local upgrade-chain range. docs/RELEASE_PROCESS explains staging and removes contradictory empty-devel instructions.

Stack: base main deliberately enables workflows. Includes #235/#239/#240 and lifecycle #251; unique migration commit 8b74310. Lifecycle review fixes are eaf1888. This replaces #200/#205's migration work only; originals remain open until integration accounts for everything. Draft pending samorev/hosted checks/full integrated gate. RC-only project: no merge, stamp or tag, and no authorization for final v2.0.


NikolayS:
Independent integrated-source review at eea1ffc: no blocking finding. The staged normalizer now recognizes/recreates both rotation and cadence validation triggers, preserves all23 canonical config columns, and creates triggers after copying retained legacy config. The migration-owned transaction token keeps installer and normalization rollback together.

Real PostgreSQL18 config_normalization_atomicity.sh passed all custom-trigger, dependent-view, and external-marker rollback cases against integrated candidate SQL. Evidence: /tmp/pg-ash-255-independent.log. Frozen-SHA upgrade matrix and actual historical tags remain separate required evidence; historical migration files themselves are not retroactively made atomic.

NikolayS:
Samorev first-round review adjudication (pinned Tanya301/samorev24924b4; report `/tmp/pg-ash-migration-samorev.md`). This is not an exact-head passing review.

All four substantive findings fixed in16cc3cd: promotion-test psql now uses explicit hosted-compatible host/user defaults; actual historical origins come from immutable git tags and discovered graph edges rather than a fixed list; archived installer filenames are discovered; staging and release-process documentation distinguish immutable released files from a current-prerelease staged refresh, including correct relative include paths.

Real PG18 checks pass all12 actual v1.0–v1.5/v2 prerelease origins through temporary promoted public wrappers and reapply; the full #202 fault suite proves installer/migration rollback from immutable v1.5 state. The old transaction marker reproduces the partial-commit defect before the fix. No frozen historical SQL files were edited, and multi-file legacy upgrade legs are not claimed to be one atomic transaction.

Integration #256 additionally recreates and verifies the new cadence-validation trigger during23-column normalization. Independent API agent reviewed that complete source and reran the full real PG18 atomicity suite successfully. Full integrated42-cell acceptance and hosted exact-head results remain required. Draft/CI state findings remain gating until resolved. Exact-head samorev is pending because the Claude session quota resets18:30PDT September4; provider failure is not a semantic pass.

PR #257, opened 2026-09-04
fix: disclose partial rollup source omissions

NikolayS:
When a requested window starts before physical raw retention, the best available minute rollup can be stale while newer raw activity exists. A two-hour fixture returned only one old backend-second and no diagnostic, despite ten newer raw appearances. Existing expected/observed bucket counts do not distinguish that stale suffix from idle time.

Preserve the documented best-partial-source result and schema. Emit an actionable NOTICE with SQLSTATE `01000` and prefix `pg_ash partial source:` when the selected rollup is stale and newer completed raw grains in the requested window have no corresponding minute-rollup row. Name the watermark, explicitly disclose omitted observations, and explain catch-up/narrow-window options. Clients must surface/preserve the diagnostic; source labels are not completeness metadata.

The real psql-client regression fails on the parent for missing diagnostic and passes on PostgreSQL 18. It verifies unchanged exact partial values and warnings through aas/timeline/top/periods/chart, plus NULL watermark. Negative controls cover healthy rollup, no newer raw, already-represented grains, raw outside the requested window, the current partial minute, full raw coverage, and the unreadable-ring branch. Physical standby coverage will run in the integrated release gate.

Stacked on #253 (therefore #235/#239/#240). Own change is commit 699729b; base remains main so repository CI triggers. No released SQL/version stamp changed. Independent review and samorev are in progress.

Relates to #122 and #248. This is explicit partial-source disclosure, not source blending or machine-readable completeness metadata. Keep #122 open for richer composition. No merge, tag, or final v2 release is requested by this draft.


NikolayS:
Independent reviewer reproduced all positive NOTICE and negative controls on PostgreSQL18 against699729b; source selection, schema, and exact partial values remain unchanged. Evidence: /tmp/pg-ash-v2-audit/257-independent.log. The docs agent additionally preserves psql stderr diagnostics in exported LLM transcripts and documents that absence of NOTICE is not proof of completeness.

Requested samorev review is pending: two untruncated scoped attempts failed because Claude Code's session usage limit is exhausted until September4 18:30PDT. These are availability failures, not successful reviews. Richer source composition/coverage metadata remains open in #122.

